PR #21, opened 2020-07-09
Extract docs into READMEs

tmlayton:
Cherry-picking the doc generation commit from [`extract-docs`](https://github.com/Shopify/argo-checkout/compare/extract-docs) so we can get that merged. The script is in a very rough, hacked state and don’t want that refactoring to hold up getting docs out.

I few notes about the design on the docs:

- Decided to use html table markup so I could nest tables to show nested interfaces. See the Image README. It’s a bit dense to have the nested table but I think the function overweighs the form.
- This could be flattened to two columns, name and description, and show the type at the top of the description which looks pretty nice.
- Since this table is markup, the markdown in the description is not formatted :/ and I don’t want to commit to a markdown-to-markup formatter unless we are happy with this format.
- I'm pulling in the autocomplete types which gets pretty crazy on select and text field, but still looks okay.
- tsdoc tags are not yet parsed. The only one used is the `@default` tag for image description which defaults to an empty string.


tmlayton:
I like these suggestions and going to play with the formatting a bit more whenever we get the script ready to ship.

PR #29, opened 2020-08-13
[Cross sells] Update postpurchase types to latest

dotLou:
1. Add shipping lines
1. Split calculate and apply outputs into their own types
1. Updated output of `calculate` to remove partial option, and remove `updatedPurchase` on error
1. Updated output of `apply` to remove purchase response.
1. Remove `changesetProcessingStatus` export - it's inline now
1. Add new `add_shipping_line` operation in `ChangeType`
1. Add new `ShippingLineChange` etc.

Edit on Aug 18:
> I just pushed another change:
> 
> `updatedPurchase` -> `calculatedPurchase`
> `UpdatedPurchase` -> `CalculatedPurchase`

### DIFF
--- a/packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
+++ b/packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
@@ -29,12 +29,14 @@ export interface PostPurchaseRenderApi
   /** Shop where the checkout/order is from */
   shop: Shop;
   /** Returns the calculations that would result from the provided changeset being applied. Used to provide cost-clarity for buyers. */
-  calculateChangeset(changeset: Readonly<Changeset>): Promise<ChangesetResult>;
+  calculateChangeset(
+    changeset: Readonly<Changeset>,
+  ): Promise<CalculateChangesetResult>;
   /**  Requests a changeset to be applied to the initial purchase, and to charge the buyer with the difference in total price, if any. */
   applyChangeset(
     signature: string,
     changeset: Readonly<Changeset>,
-  ): Promise<ChangesetResult>;
+  ): Promise<ApplyChangesetResult>;
   /**
    * Indicates that the extension has finished running.
    * Currently, effectively redirects buyers to the thank you page.
@@ -115,7 +117,15 @@ interface AddVariantChange {
   discount?: ExplicitDiscount;
 }
 
-type Changes = AddVariantChange[];
+interface AddShippingLineChange {
+  type: 'add_shipping_line';
+  price: number;
+  title?: string;
+  presentmentTitle?: string;
+  phone?: string;
+}
+
+type Changes = (AddVariantChange | AddShippingLineChange)[];
 
 interface Changeset {
   changes: Changes;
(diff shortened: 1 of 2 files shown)


packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
```diff
@@ -179,12 +193,11 @@ type ChangesetResult =
   | {
       errors: ChangesetError[];
       status: 'unprocessed';
-      updatedPurchase: UpdatedPurchase | null;
```
robin-drexler, line 182: Is the key `updatedPurchase` completely missing from the response or `null`?

robin-drexler:
If apply and calculate now produce different results (no `updatedPurchase` on the latter), should we make those two distinct types?

packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
dotLou: It would be entirely missing - you shouldn't ever get `unprocessed` unless there was an error, in which case that should be the only thing in your payload - we can't give you a calculated order if there was a failure.

dotLou:
> If apply and calculate now produce different results (no `updatedPurchase` on the former), should we make those two distinct types?

That might actually make sense. The only common piece now is the error + status.

dotLou:
> That might actually make sense. The only common piece now is the error + status.

DONE. Did a bit of cleanup, re-updated the PR description.

packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
```diff
@@ -115,7 +117,15 @@ interface AddVariantChange {
   discount?: ExplicitDiscount;
 }
 
-type Changes = AddVariantChange[];
+interface AddedShippingLineChange {
```
denislins, line 120: This should probably be `AddShippingLineChange` instead

ttemrahcelik:
These updates haven't gone into the doc yet, right?

dotLou:
> These updates haven't gone into the doc yet, right?

AFAIK they're in at least our own copy of the docs.

dotLou:
I just pushed another change:

`updatedPurchase` -> `calculatedPurchase`
`UpdatedPurchase` -> `CalculatedPurchase`

ttemrahcelik:
> I just pushed another change:
> 
> `updatedPurchase` -> `calculatedPurchase`
> `UpdatedPurchase` -> `CalculatedPurchase`

Wanna add this to the PR description too? 🙏

packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
```diff
@@ -115,7 +117,15 @@ interface AddVariantChange {
   discount?: ExplicitDiscount;
 }
 
-type Changes = AddVariantChange[];
+interface AddShippingLineChange {
+  type: 'add_shipping_line';
+  price: number;
+  title?: string;
+  presentmentTitle?: string;
+  phone?: string;
```
ginmrt, line 125: `phone` is not in the Docs (we should update the Docs :P)

packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
```diff
@@ -159,13 +169,19 @@ interface UpdatedLineItem {
   quantity: number;
 }
 
-interface UpdatedPurchase {
+interface AddedShippingLine {
+  priceSet: MoneyBag;
```
ginmrt, line 173: Another docs issue where it lists `price` as part of this (guessing that's wrong)

ttemrahcelik: Yeah we should remove `price` from the doc, there is both `priceSet` and `price` there.

ttemrahcelik: Done. ✅

packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
ttemrahcelik: I added the field to the doc, but I'm not sure what the description should be there.

PR #31, opened 2020-08-20
serve config json

robin-drexler:


packages/argo-run/src/dev.ts
```diff
@@ -98,3 +113,13 @@ function getOpenUrl(args: string[]) {
   const openArg = namedArgument('open', args);
   return (openArg ?? '').startsWith('http') ? new URL(openArg!) : undefined;
 }
+
+function readConfig() {
+  const path = resolve(join(process.cwd(), 'extension.config.json'));
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    return require(path);
```
robin-drexler, line 121: just used that as a shortcut to read a json quick and dirty.

PR #32, opened 2020-08-24
read config and provide query string

robin-drexler:
fixes https://github.com/Shopify/cross-sell-app/issues/188

Reads `config.extension.yml` and prints the content as copy/pastable query string to the terminal. 
Partners can copy paste this to apply metafields during development. (not implemented in core yet)

This also exposes a `/config` endpoint which responds with the config as json.
The endpoint is going to be used by our extension to automatically append the config, just like we're doing for the script_url itself. (PR for that will follow)

🎩 

We'll use a test branch in the `argo-checkout-template` repo with a locally built and linked argo-run (package in this repo) version. 

Make sure, you have all the repos:

```
dev clone argo-checkout
dev clone argo-checkout-template
```

prepare this branch and build all the things™
```
dev cd argo-checkout
git fetch
git checkout config-query
yarn
yarn build
```

Prepare the top hat branch
```
dev cd argo-checkout-template
git fetch
git checkout tophat-config
yarn
```

🔗  it all together
```
yarn link --cwd $(dev cd -n argo-checkout)/packages/argo-checkout
yarn link --cwd $(dev cd -n argo-checkout-template) "@shopify/argo-run"
```

serve
```
dev cd argo-checkout-template
yarn server
```

You should see an output like: 

```
🔭 > Starting dev server on http://localhost:39351
🔭 > Your extension is available at http://localhost:39351/assets/extension.js
🔭 > You can append this query string: script_url=http%3A%2F%2Flocalhost%3A39351%2Fassets%2Fextension.js&config=%7B%22metafields%22%3A%5B%7B%22namespace%22%3A%22my-app%22%2C%22key%22%3A%22first-key%22%7D%2C%7B%22namespace%22%3A%22my-app%22%2C%22key%22%3A%22second-key%22%7D%5D%7D
```

Take a look at `config.extension.yml` to familiarize yourself with its content.

Start a checkout until the payments page.
Append (or replace if the browser extension already added script_url) the mentioned query string.

Open the console:

```
JSON.parse(new URLSearchParams(location.search).get('config'))
```
Should log metafield arrays.

Go to 

 http://localhost:39351/config and check that it responds with config json

If you delete or empty or put invaid yml inside `extension.config.yml`  the endpoint should return 404 and the terminal should not show the query string output. 

In case of an invalid file, it should, however, show that error.

### DIFF
--- a/packages/argo-run/src/dev.ts
+++ b/packages/argo-run/src/dev.ts
@@ -15,6 +15,7 @@ export async function dev(...args: string[]) {
   const publicPath = `${url}/assets/`;
   const filename = 'extension.js';
   const fileUrl = `${publicPath}${filename}`;
+  const configPath = `/config`;
 
   const compiler = webpack(
     createWebpackConfiguration({
@@ -61,6 +62,15 @@ export async function dev(...args: string[]) {
       },
     },
   });
+
+  if (extensionConfig) {
+    app.use(async (ctx, next) => {
+      if (ctx.path !== configPath) {
+        return next();
+      }
+      ctx.body = extensionConfig;
+    });
+  }
 
   app.use(middleware);
 
--- a/packages/argo-run/src/dev.ts
+++ b/packages/argo-run/src/dev.ts
@@ -88,12 +88,12 @@ export async function dev(...args: string[]) {
 
   log(`Your extension is available at ${fileUrl}`);
 
-  // XXX we should replace this with a check for extension type
+  // we could replace this with a check for extension type
   // to only show the query parameters for post purchase extensions
   if (extensionConfig) {
     const query = new URLSearchParams();
-    query.set('script_url', fileUrl);
 
+    query.set('script_url', fileUrl);
     query.set('config', JSON.stringify(extensionConfig));
 
     log(`You can append this query string: ${query.toString()}`);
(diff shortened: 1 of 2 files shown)


packages/argo-run/src/dev.ts
```diff
@@ -73,6 +88,17 @@ export async function dev(...args: string[]) {
 
   log(`Your extension is available at ${fileUrl}`);
 
+  // we could replace this with a check for extension type
+  // to only show the query parameters for post purchase extensions
+  if (extensionConfig) {
```
robin-drexler, line 93: In the future, we could also listen for file changes it update the config so consumers would not have to restart upon config change, but I think for now that's fine

packages/argo-run/src/dev.ts
```diff
@@ -98,3 +124,18 @@ function getOpenUrl(args: string[]) {
   const openArg = namedArgument('open', args);
   return (openArg ?? '').startsWith('http') ? new URL(openArg!) : undefined;
 }
+
+function readConfig() {
```
kmdavis, line 128: suggestion 1: support `extension.config.json` also

suggestion 2: support an `extension_config` object inside `package.json`

robin-drexler: yeah, I thought about this and was very close to just using https://github.com/davidtheclark/cosmiconfig, however this file also needs to be read from the ruby cli (for the `push` command), so in the interest of simplicity I kept it for `yml` for now.

PR #36, opened 2020-09-02
Adding Banner component

kmdavis:
This PR adds a Banner component (from checkout-web-ui)

Screen Shot 2020-09-02 at 11 18 27 AM

```jsx
<Banner title="This is an informative message" />
<Banner status="success" title="This is a success message" />
<Banner status="warning" title="This is a warning message" />
<Banner status="critical" title="This is an error message" />
<Banner title="This is a message with content">
  <Text>
    Lorem ipsum, dolor sit amet consectetur adipisicing elit. Dolores
    commodi, vitae explicabo, laudantium consequatur eaque dignissimos
    dolorem optio minus porro nesciunt magni reprehenderit accusamus quasi
    sit architecto sequi nihil.
  </Text>
</Banner>
```

### DIFF
--- a/packages/argo-checkout/src/components/Banner/Banner.ts
+++ b/packages/argo-checkout/src/components/Banner/Banner.ts
@@ -1,10 +1,11 @@
 import {createRemoteComponent} from '@remote-ui/core';
 
 /**
-   * Info:
-   * Use to convey general information or actions that aren’t critical or tied to a particular action. 
- 
-   * Success:
+ * Info:
+ * Use to convey general information or actions that aren’t critical or tied to
+ * a particular action.
+ *
+ * Success:
  * Use rarely, only if you need additional visual confirmation that a
  * non-standard action has been completed successfully, for example adding an
  * item to an order as an upsell.
--- a/packages/argo-checkout/src/components/Banner/Banner.ts
+++ b/packages/argo-checkout/src/components/Banner/Banner.ts
@@ -1,0 +1,43 @@
+import {createRemoteComponent} from '@remote-ui/core';
+
+/**
+ * Success:
+ * Use rarely, only if you need additional visual confirmation that a
+ * non-standard action has been completed successfully, for example adding an
+ * item to an order as an upsell.
+ *
+ * Warning:
+ * Use to display information that needs attention or that customers need to
+ * take action on. Seeing these banners can be stressful for customers so be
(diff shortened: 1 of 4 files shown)


packages/argo-checkout/src/components/Banner/Banner.ts
```diff
@@ -0,0 +1,43 @@
...
+ * take action on. Seeing these banners can be stressful for customers so be
+ * cautious about using them. Does not block progress to next step.
+ *
+ * Critical:
+ * Use to communicate problems that have to be resolved immediately for
+ * customers to complete a task. For example, using a different payment method
+ * if card details couldn't be processed. Seeing these banners can be stressful
+ * for customers so be cautious about using them.
+ */
+type Status = 'info' | 'success' | 'warning' | 'critical';
```
jffortier, line 20: We're missing the `info` description.

packages/argo-checkout/src/components/Banner/Banner.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface BannerProps {
+  /**
+   * Banners have an optional title. Use a title to grab the buyers attention
+   * with a short, concise message.
+   */
+  title?: string;
+  /**
+   * What status icon & color scheme to use
```
jffortier, line 29:
```suggestion
   * Sets the status of the banner.
```

I would prefer not mentioning any visual treatment as this may vary based on theming. We should make sure the status is aligned with its content, not the color or icon we want.

packages/argo-checkout/src/components/Banner/Banner.ts
```diff
@@ -0,0 +1,43 @@
+import {createRemoteComponent} from '@remote-ui/core';
+
+/**
+ * Success:
```
jffortier, line 4: We're missing the `info` description.

Something like that:

```suggestion
   * Info:
   * Use to convey general information or actions that aren’t critical or tied to a particular action. 
 
   * Success:
```

packages/argo-checkout/src/components/Banner/Banner.ts
kmdavis: yeah, I copied this from checkout-web-ui; I'm guessing it wasn't there because it's the default? I'll see if I can cobble something together.

PR #46, opened 2020-09-25
[Cross Sell] Introduce SetMetafield as a change type

ttemrahcelik:
Closes: https://github.com/Shopify/cross-sell-app/issues/279

### DIFF
--- a/packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
+++ b/packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
@@ -127,7 +127,11 @@ interface AddShippingLineChange {
   phone?: string;
 }
 
-type Changes = (AddVariantChange | AddShippingLineChange)[];
+interface SetMetafield extends Metafield {
+  type: 'set_metafield';
+}
+
+type Changes = (AddVariantChange | AddShippingLineChange | SetMetafield)[];
 
 interface Changeset {
   changes: Changes;


packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
```diff
@@ -127,7 +127,11 @@ interface AddShippingLineChange {
   phone?: string;
 }
 
-type Changes = (AddVariantChange | AddShippingLineChange)[];
+interface SetMetafield extends Metafield {
```
robin-drexler, line 130: My gut tells me 

```ts
interface SetMetafield {
  type: 'set_metafield';
  metafield: Metafield
}
```

would be better / cleaner, but I can't really say why, so either is fine

robin-drexler: ah and it'd also impact the backend, so I guess we're probably not that inclined to change it. :D

robin-drexler (approved):

packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
ttemrahcelik: Hmm, it's not how we defined `AddVariant` and `AddShippingLine` either. So yeah I'll skip this. 😅  But not because it doesn't make sense (I'm good either way), but to stay consistent with other inputs too.

denislins (approved):

PR #49, opened 2020-10-01
[Post Purchase] Add InputData type

ginmrt:
## Description

Compatibility update for https://github.com/Shopify/shopify/pull/262422

### DIFF
--- a/packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
+++ b/packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
@@ -4,13 +4,10 @@ import type {ValueOrPromise} from '../shared';
 /** Input given to the ShouldRender extension point (Checkout::PostPurchase::ShouldRender) */
 export interface PostPurchaseShouldRenderApi
   extends StandardApi<'Checkout::PostPurchase::ShouldRender'> {
-  /** Initial purchase */
-  initialPurchase: Purchase;
+  /** Input data given to the extension point */
+  inputData: InputData;
   /** General purpose storage for extensions */
   storage: Storage;
-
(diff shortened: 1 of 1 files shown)


packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
```diff
@@ -49,6 +44,21 @@ interface Storage {
...
+  /** Identifier for the extension point */
+  extensionPoint: string;
+  /** Initial purchase */
+  initialPurchase: Purchase;
+  /** Checkout customer locale */
+  locale: string;
+  /** Shop where the checkout/order is from */
+  shop: Shop;
+  /** JWT representing the input_data payload */
+  token: string;
```
robin-drexler, line 57: For my understanding: `token` holds the entire signed `inputData` sans the token itself? 🤔

ginmrt: That's correct, I could word the comments better, if that was the concern?

robin-drexler: No, the comments are fine, I just found it a bit surprising to have the token be part of the object that the token itself represents without itself being included.

I personally would have expected the token to be a sibling of inputData, but that's not a blocker of course.
Just wanted to be sure I understood correctly.

ginmrt: It's how it had been implemented already with `initial_purchase`, with this change the token moves but is still contained inside the signed object. I can see the oddity of it for sure.

robin-drexler (approved):

PR #50, opened 2020-10-01
Update remote-ui dependencies

lemonmade:
This PR updates the remote-ui dependencies. I made one breaking change to `@remote-ui/react` that needs to be accounted for: `render()` no longer calls `root.mount()` automatically, but does allow a callback where we can call that method at the same point it was previously being called. This change was made to allow `render` to support more use cases, most notably "taking over" a remote root object that is already mounted.

cc/ @henrytao-me @vividviolet @alanthai the same change will be needed on the admin library.

### DIFF
--- a/packages/argo-checkout-react/package.json
+++ b/packages/argo-checkout-react/package.json
@@ -9,7 +9,7 @@
   "sideEffects": false,
   "license": "MIT",
   "dependencies": {
-    "@remote-ui/react": "^1.0.5",
+    "@remote-ui/react": "^2.0.0",
     "@shopify/argo-checkout": "^0.5.4",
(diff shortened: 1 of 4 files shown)


robin-drexler (approved):

henrytao-me:
@lemonmade Do you think it is important for launch? I think it isn't because the host implementation remains unchanged and it's backward compatible with old and new version of remote-ui. Please correct me if I am wrong.

lemonmade:
The implementation is not unchanged, if a developer were to upgrade their local `@remote-ui/react` dependency themselves, they would break their extension. I'd rather we take the practice to get better at dependency upgrades.

PR #51, opened 2020-10-08
ensure terser does not produce IE 11 incompatible bundles

robin-drexler:
It optimized objects to use shorthand property syntax.

`{setTimeout: setTimeout}` to `{setTimeout}`

Which isn't supported in IE(11).

Changed ecma version to 5 to avoid even more potential surprises. 😬


lemonmade (approved):
Good find 👍

PR #54, opened 2020-10-21
[WIP] add stepper component

robin-drexler:
🎩 by building and copying the argo-* packages to checkout-web and my local extension and run it all together.

checkout-web PR will follow after this PR has been merged and the package has been shipped.

image


robin-drexler:
Just learned that the component isn't quite ready yet. 
Let's keep that PR open and only update props (if needed) once the Stepper is available

PR #57, opened 2020-11-03
include node_modules when transpiling

robin-drexler:
Fixes issues where required modules weren't probably transpiled to work with our browser matrix.
Usually packages are supposed to produce proper ES5 builds, but not all of them do :( 

E.g. `@lightspeed/i18n-number-format` contains rest and spread.

Will come at the expense of slower builds.

🎩 dev and prod build locally with newly created argo-checkout-template project. 
Installed and used `@lightspeed/i18n-number-format` (as this package is known to cause issues) and checked that no rest or spread was to be found anymore.

🎩 prod

```
yarn build
cd build 
npx http-serve -c-1 -p 39351
```

use http://localhost:39351/main.js as script_url

We can always go back to another solution (like allowing consumers to allowlist packages) later.

### DIFF
--- a/packages/argo-run/src/webpack-config.ts
+++ b/packages/argo-run/src/webpack-config.ts
@@ -114,7 +114,7 @@ export function createWebpackConfiguration({
           },
         },
         {
-          test: /\.(ts|tsx)$/,
+          test: /\.(tsx?)$/,
           exclude: /node_modules/,
           use: {
             loader: 'babel-loader',
--- a/packages/argo-run/src/webpack-config.ts
+++ b/packages/argo-run/src/webpack-config.ts
@@ -84,7 +84,7 @@ export function createWebpackConfiguration({
     module: {
       rules: [
         {
-          test: /\.js$/,
+          test: /\.(jsx?|esnext)$/,
           exclude: /node_modules/,
           use: {
             loader: 'babel-loader',
@@ -99,7 +99,8 @@ export function createWebpackConfiguration({
           },
         },
         {
-          test: /\.esnext$/,
+          test: /\.(js|esnext)$/,
+          include: /node_modules/,
           use: {
             loader: 'babel-loader',
             options: {
--- a/packages/argo-run/src/webpack-config.ts
+++ b/packages/argo-run/src/webpack-config.ts
@@ -105,6 +105,7 @@ export function createWebpackConfiguration({
             loader: 'babel-loader',
             options: {
               configFile: false,
+              compact: true,
               ...babelConfig({
                 react: false,
                 typescript: false,


robin-drexler:
I'd probably cut a beta release and have that tested before releasing that to all of the public

packages/argo-run/src/webpack-config.ts
```diff
@@ -98,23 +97,8 @@ export function createWebpackConfiguration({
...
-              ...babelConfig({
-                react: false,
-                typescript: false,
-                targets,
-              }),
-            },
-          },
-        },
         {
           test: /\.(ts|tsx)$/,
```
jared-dykstra, line 101: for consistency: `test: /\.tsx?$/,`

packages/argo-run/src/webpack-config.ts
```diff
@@ -98,23 +97,8 @@ export function createWebpackConfiguration({
...
           },
         },
-        {
-          test: /\.esnext$/,
-          use: {
-            loader: 'babel-loader',
-            options: {
-              configFile: false,
-              ...babelConfig({
-                react: false,
```
jared-dykstra, line 108: There's a difference: `react: false` => `react: useReact`  Is that difference OK?

packages/argo-run/src/webpack-config.ts
```diff
@@ -84,8 +84,7 @@ export function createWebpackConfiguration({
     module: {
       rules: [
         {
-          test: /\.js$/,
-          exclude: /node_modules/,
```
jared-dykstra, line 88: 👍  This line is the key change in this PR

packages/argo-run/src/webpack-config.ts
```diff
@@ -84,8 +84,7 @@ export function createWebpackConfiguration({
...
+          test: /\.(jsx?|esnext)$/,
```
lemonmade, line 87: I'd prefer we keep the rule for handling `.js` in `node_modules` separate from the one for app code, most likely other parts of it will be different (at least the cacheing, possibly also the Babel options at some point).

lemonmade (approved):

packages/argo-run/src/webpack-config.ts
robin-drexler: good catch, it should be false for node_modules as we don't wanna compile jsx there, I think.

That'd also go with what @lemonmade suggested to have a different config for node_modules vs app code.

packages/argo-run/src/webpack-config.ts
robin-drexler: Changed it so that this is only for app code and https://github.com/Shopify/argo-checkout/pull/57/files#diff-2087c6e948a1e8019253ff0fdfbd74b4e3911a6a6e5eb34fb82ca6000b0707b0R103 for node_modules

robin-drexler:
re-requested reviews as I now changed it to be two distinct configs for node_modules vs app code

packages/argo-run/src/webpack-config.ts
```diff
@@ -113,7 +115,7 @@ export function createWebpackConfiguration({
           },
         },
         {
-          test: /\.(ts|tsx)$/,
+          test: /\.(tsx?)$/,
```
lemonmade, line 118: Parens no longer necessary

lemonmade (approved):

robin-drexler:
Added compact mode for node_module files to speed things up, but also to remove the warning:
```
The code generator has deoptimised the styling of ...
``` 
caused by large packages (esp in development) like react-reconciler.

jared-dykstra (approved):
🎩

robin-drexler:
This was tested successfully by a partner with a beta version I had released.

PR #62, opened 2021-01-06
Update standard api interface for post purchase

alxclark:
## What problem are you trying to solve ?

While following the [Post-purchase getting started guide](https://docs.google.com/document/d/1JqTUEBXWZ2gKC7bfWc0wy7i6zHtHmldd3ef4N3FHoeU/edit#), I noticed that the extension was not rendering the `extensionPoint` field properly from the extension input. This is caused by `useExtensionInput` not returning a value for all of its fields and not respecting its own types.

```
// Expected
The body of this page was rendered by Checkout::PostPurchase::Render

// Actual
The body of this page was rendered by undefined
```

**Example**

image

This led me to investigate on the signature of the `useExtensionInput` hook which returns the `extensionPoint` field, where I noticed that some fields where duplicated.

For instance, here's the types for the `PostPurchaseRenderApi` :

<summary>
<details>

```ts
export interface PostPurchaseRenderApi
  extends StandardApi<'Checkout::PostPurchase::Render'> {
  /** Input data given to the extension point */
  inputData: InputData;
  /** General purpose storage for extensions */
  storage: Storage;
  /** Returns the calculations that would result from the provided changeset being applied. Used to provide cost-clarity for buyers. */
  calculateChangeset(
    changeset: Readonly<Changeset> | string,
  ): Promise<CalculateChangesetResult>;
  /**  Requests a changeset to be applied to the initial purchase, and to charge the buyer with the difference in total price, if any. */
  applyChangeset(changeset: string): Promise<ApplyChangesetResult>;
  /**
   * Indicates that the extension has finished running.
   * Currently, effectively redirects buyers to the thank you page.
   */
  done(): Promise<void>;
}

interface InputData {
  /** Identifier for the extension point */
  extensionPoint: string;
  /** Initial purchase */
  initialPurchase: Purchase;
  /** Checkout customer locale */
  locale: string;
  /** Shop where the checkout/order is from */
  shop: Shop;
  /** JWT representing the input_data payload */
  token: string;
  /** Post Purchase API version */
  version: string;
}
```
</details>
</summary>

And here's the type of the `StandardApi` :

```ts
export interface StandardApi<
  ExtensionPoint extends import('../../extension-points').ExtensionPoint
> {
  version: Version;
  locale: string;
  extensionPoint: ExtensionPoint;
}
```

The issue with the following types is that it requires the following structure :

```ts
// Representation of the type with the types from standard api merged into this interface
interface PostPurchaseRenderApi {
  version: Version;
  locale: string;
  extensionPoint: ExtensionPoint;

  inputData: {
    version: string; // Those three fields are available right above
    locale: string;
    extensionPoint: string;
    // ... more fields
  }

  // Other fields (not related)
  calculateChangeset: () => void,
  done: () => void,
  applyChangeset: () => void,
  storage: Storage;
}
```

**But** here's what the post purchase page is using as input for the post purchase extension render :

(https://github.com/Shopify/shopify/blob/master/app/assets/javascripts/checkout/esnext/post_purchase/post_purchase_page.js#L188-L197)

```js
return window.CheckoutExtensions.render({
          id: EXTENSION_POINT_RENDER,
          script: scriptUrl,
          root,
          themeSettings,
          i18n,
          input: { 
            // This should comply with the standard api and have a version, locale and extension point field
            // But it doesn't 
            inputData,
            done: this.handleDone,
            calculateChangeset: this.handleCalculateChangeset.bind(this),
            applyChangeset: this.handleApplyChangeset.bind(this),
            storage: {
              initialData: storageData,
              update: this.getStorage().set,
            },
          },
          onStopwatchMark,
        });
```

**Ultimately**, this is causing the following :

```tsx
const input = useExtensionInput<'Checkout::PostPurchase::Render'>();
console.log(input.extensionPoint); // undefined
console.log(input.inputData.extensionPoint); // "Checkout::PostPurchase::Render"
```

**Example:**

_**Note 1 (left):** Notice that the same fields are undefined on the root and defined when logging from the `inputData`_
_**Note 2 (right):**_ Here's the exact same extension running on https://argogogo.dev
image

## How are you solving it ?

There is really two solutions :
1. Update the post purchase JS file in core and pass the parameters in both place. This doesn't break anything existing, but there's redundancy now.
2. Update the Standard api for post purchase and accept `inputData` and `storage` for each extension point. This breaks the shopify CLI typescript example which refers directly to `useExtensionInput().extensionPoint` instead of `useExtensionInput().inputData.extensionPoint`. But it's now _"impossible"_ for a consumer of the `useExtensionInput` hook to refer to the value directly.

There is a big diff because I'm moving some types around, but the only changes are updating the standard api to this shape , which was already respected by both post-purchase extension points:

```ts
export interface StandardApi<ExtensionPoint extends RegisteredExtensionPoint> {
  /** Input data given to the extension point */
  inputData: InputData<ExtensionPoint>;
  /** General purpose storage for extensions */
  storage: Storage;
```

## Notes

* If we are happy with those changes I'll also update the readme for the standard api which would be out of sync. I'll also update the CLI example to refer to `inputData`.
* If you want to reproduce the original issue, you can simply run `shopify create extension` and follow along https://docs.google.com/document/d/1JqTUEBXWZ2gKC7bfWc0wy7i6zHtHmldd3ef4N3FHoeU/edit#

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,7 +6,9 @@
     "typescriptreact"
   ],
   "editor.formatOnSave": true,
-  "eslint.enable": true,
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true
+  },
   "[typescript]": {
     "editor.defaultFormatter": "dbaeumer.vscode-eslint"
   },
--- a/packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
+++ b/packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
@@ -1,6 +1,5 @@
 import type {StandardApi} from '../standard';
-import type {ValueOrPromise} from '../shared';
-import {Metafield, MoneyBag} from '../shared';
+import type {ValueOrPromise, Metafield, MoneyBag} from '../shared';
 
 /** Input given to the ShouldRender extension point (Checkout::PostPurchase::ShouldRender) */
 export interface PostPurchaseShouldRenderApi
--- a/packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
+++ b/packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
@@ -1,14 +1,10 @@
 import type {StandardApi} from '../standard';
 import type {ValueOrPromise} from '../shared';
+import {Metafield, MoneyBag} from '../shared';
 
 /** Input given to the ShouldRender extension point (Checkout::PostPurchase::ShouldRender) */
 export interface PostPurchaseShouldRenderApi
-  extends StandardApi<'Checkout::PostPurchase::ShouldRender'> {
-  /** Input data given to the extension point */
-  inputData: InputData;
-  /** General purpose storage for extensions */
-  storage: Storage;
-}
+  extends StandardApi<'Checkout::PostPurchase::ShouldRender'> {}
 
 /** Output expected from the ShouldRender extension point (Checkout::PostPurchase::ShouldRender) */
 export type PostPurchaseShouldRenderResult = ValueOrPromise<{
@@ -19,10 +15,6 @@ export type PostPurchaseShouldRenderResult = ValueOrPromise<{
 /** Input given to the render extension point (Checkout::PostPurchase::Render) */
 export interface PostPurchaseRenderApi
   extends StandardApi<'Checkout::PostPurchase::Render'> {
-  /** Input data given to the extension point */
-  inputData: InputData;
-  /** General purpose storage for extensions */
-  storage: Storage;
   /** Returns the calculations that would result from the provided changeset being applied. Used to provide cost-clarity for buyers. */
   calculateChangeset(
     changeset: Readonly<Changeset> | string,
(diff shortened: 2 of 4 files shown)


.vscode/settings.json
```diff
@@ -6,7 +6,9 @@
     "typescriptreact"
   ],
   "editor.formatOnSave": true,
-  "eslint.enable": true,
```
alxclark, line 9: This setting is deprecated and the current configuration was non sufficient to enable format on save. The `codeActionsOnSave` option follows the same config we expose on the CLI.

packages/argo-checkout/src/extension-points/api/shared.ts
```diff
@@ -1 +1,70 @@
 export type ValueOrPromise<T> = T extends PromiseLike<any> ? T : T | Promise<T>;
+
```
alxclark, line 2: I've moved types used for the standard api to this shared file in order to keep the api file clean with only `InputData`, `Storage` and `StandardApi` and avoid circular references.

packages/argo-checkout/src/extension-points/api/post-purchase/post-purchase.ts
```diff
@@ -1,14 +1,9 @@
 import type {StandardApi} from '../standard';
```
lemonmade, line 1: Do we still need this package internally within the repo? Can we instead use the types from the `@shopify/argo-checkout` directly?

alxclark: I don't really have a strong opinion on this, I'll probably defer to you that decision since I'm still ramping up. I'm happy to remove it if we think we should.

lemonmade: @denislins do you know if there was any reason that we still needed an in-repo version of the Argo API types?

ginmrt:
Very well documented PR there! When we first discussed this, I didn't get it right away due to me not really ever directly using/interacting with those values that were duplicated directly under `StandardApi`, hence why I kept pointing to the ones that were part of `inputData`.

Makes sense to clean this up. I'm adding @robin-drexler as reviewer as he will have more context than myself in this area.

My main concern is about not breaking extensions currently in the wild, but it seems like the values under `StandardApi` were  ultimately unused (though don't just take my word for it), versus the ones in `inputData` which are provided to partner extensions.

robin-drexler:
Great write-up @alxclark 👏 

If I remember correctly, we introduced the `inputData` object to have the entire input signed. https://github.com/Shopify/shopify/pull/262422

I wonder if `version` needs to be part of this or should rather be pushed one level outside `inputData`. Which would also align more with all other extension points.

In order not to break things, I'd also suggest to pass it into both places and deprecate it inside `inputData` and remove it eventually.

Unless we think this should indeed be part of the signed input(?)

alxclark:
Thanks for the review @robin-drexler, I'll update the post purchase render to pass it in both place so at least we know the field will be filled for current extensions. I'll hold off on the current changes until we have a better idea and maybe deprecate.

PR #64, opened 2021-02-10
Make hot reloading more reliable

lemonmade:
This changes the hot reloading client code a bit to make it more reliable. Previously, the first compilation was always "ignored" because we used that to signal the socket was open. The main change here is using the `open` event to determine that instead, and to cleanup on refresh a bit better since I was noticing extra websockets hanging around.

cc/ @henrytao-me @vividviolet your version of this also suffers from the same issue I think: https://github.com/Shopify/argo-admin/blob/main/packages/argo-admin-cli/src/server/hot-client.ts

### DIFF
--- a/packages/argo-webpack-hot-client/src/worker.ts
+++ b/packages/argo-webpack-hot-client/src/worker.ts
@@ -33,7 +33,7 @@ type Message =
 
 run();
 
-function run({retry = 0} = {}) {
+function run() {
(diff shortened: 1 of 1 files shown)


henrytao-me:
I created a follow up ticket. Thanks Chris. FYI, argo admin version already handles WebSocket clean up but it can be improved.

PR #65, opened 2021-02-17
Revert back to default to "regular" React

lemonmade:
We were having issues with some extensions on the new 0.9 version of argo-checkout, which uses mini-react by default. This PR reverts using mini-react by default until we know the cause of the issues we are seeing. For partners, this change is completely transparent, but under the hood their extensions will go back up in size a bit now that we are using the "real" React again.


henrytao-me (approved):
Changes look good 👍

denislins (approved):

PR #68, opened 2021-03-24
Use webpack-dev-server for Argo dev mode

lemonmade:
This PR switches us from using `koa-webpack` to using `webpack-dev-server` to serve local extensions. This is basically a no-op for consumers, but it brings us in line with [argo-admin](https://github.com/Shopify/argo-admin/tree/main/packages/argo-admin-cli), and gives us a way of specifying the websocket path on the port as the asset server, which we need to make it easier to use ngrok for local checkout development.

### DIFF
--- a/packages/argo-run/package.json
+++ b/packages/argo-run/package.json
@@ -26,9 +26,8 @@
   "devDependencies": {
     "@types/glob": "^7.1.1",
     "@types/js-yaml": "^3.12.5",
-    "@types/koa": "^2.11.3",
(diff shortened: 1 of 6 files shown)


henrytao-me (approved):
🎩-ed with argo-checkout-template. You might want to use `historyApiFallback` to point any invalid paths to a default path. Ex:

```
    historyApiFallback: {
      rewrites: [
        {
          from: /./,
          to: '/assets/extension.js',
        },
      ],
    },
```

Otherwise, it shows all directories in the home page 

image

PR #69, opened 2021-04-01
Update local development for checkout extensions

lemonmade:
This PR updates the local development experience for extensions that are detected as being checkout extensions (this detection will rely on the new template being introduced in https://github.com/Shopify/argo-checkout-template/pull/40):

- A new endpoint, `/query` exists. This endpoint fetches the metadata of your extension "workspace". Right now, it will just be a listing of one extension, because that's all our CLI supports. That metadata includes the asset URL and socket URL
- We print instructions for running these new checkout extensions, which goes through the whole "start a tunnel, make a checkout, paste this query param" flow that the developer needs to do next.

cc/ @henrytao-me @vividviolet

### DIFF
--- a/packages/argo-run/src/dev.ts
+++ b/packages/argo-run/src/dev.ts
@@ -91,10 +91,80 @@ export async function dev(...args: string[]) {
     publicPath: PUBLIC_PATH,
 
     before(app) {
-      app.get('/', (_, res) => {
+      app.get('/', (req, res) => {
+        const protocol = req.headers['x-forwarded-proto'] ?? req.protocol;
+        const host = req.headers.host;
+        const origin = `${protocol}://${host}`;
+
         res.set('Content-Type', 'text/html');
         res.set('Cache-Control', 'no-cache');
-        res.send('<html>TODO</html>');
+
+        function renderIndexPage(content: string) {
+          res.send(`
+            <html>
+              <head>
+                <meta data-react-html="true" name="viewport" content="width=device-width, initial-scale=1.0, height=device-height, viewport-fit=cover">
+                <style>
+                  * {
+                    box-sizing: border-box;
+                  }
+
+                  body {
+                    min-height: 100vh;
+                    min-height: -webkit-fill-available;
+                    display: grid;
+                    margin: 0;
+                    align-content: center;
+                    justify-content: center;
+
+                    font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen',
+                      'Ubuntu', 'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue',
+                      sans-serif;
+                  }
+
+                  html {
+                    height: -webkit-fill-available;
+                    background: black;
+                    color: white;
+                  }
+
+                  .content {
+                    max-width: 40rem;
+                    padding: 2rem;
+                    font-size: 1.5em;
+                  }
+
+                  code {
+                    color: mediumseagreen;
+                    word-break: keep-all;
+                  }
+                </style>
+              </head>
+
+              <body>
+                <div class="content">
+                  <p>
+                    This page is served by your local Argo development server. Instead of
+                    visiting this page directly, you will need to connect your local development
+                    environment to a real checkout environment.
+                  </p>
+                  <p>${content}</p>
+                </div>
+              </body>
+            </html>
+          `);
+        }
+
+        if (protocol === 'https') {
+          renderIndexPage(
+            `Create a checkout and append <code>?dev=${origin}/query</code> to the URL to start developing your extension.`,
+          );
+        } else {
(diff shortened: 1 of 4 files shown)


packages/argo-run/src/dev.ts
```diff
@@ -89,18 +91,115 @@ export async function dev(...args: string[]) {
...
+            replaced with your own ngrok URL.`);
+        }
+
         res.end();
       });
 
+      // This endpoint is used by C1 to request information about the local workspace,
+      // including what extensions are available locally, and how to connect to the
+      // build server websocket for those extensions.
+      app.get('/query', (req, res) => {
```
vividviolet, line 174: Should we make  this `/data` to be consistent with Admin?

packages/argo-run/src/dev.ts
```diff
@@ -89,18 +91,115 @@ export async function dev(...args: string[]) {
...
+
+        res.set('Access-Control-Allow-Origin', '*');
+        res.set('Cache-Control', 'no-cache');
+
+        res.json({
+          queryUrl: `${origin}${req.path}`,
+          extensions: [
+            {
+              scriptUrl: `${origin}${PUBLIC_PATH}${filename}`,
+              socketUrl: `${origin.replace(/^http/, 'ws')}${WEBSOCKET_PATH}`,
```
vividviolet, line 187: Should the manifest fields be named the same as Admin? https://github.com/Shopify/argo-admin/blob/1c977dfa7a261f083e01a56562a4dc012cc99544/packages/argo-admin-cli/src/server/index.ts#L83

packages/argo-run/src/dev.ts
```diff
@@ -141,11 +240,22 @@ export async function dev(...args: string[]) {
...
+        getLegacyPostPurchaseData(localhostScriptUrl, extension),
+      )}`,
+    );
+  } else {
+    log(`Next, run \`shopify tunnel start --port=${port}\` in a new terminal.`);
+    log(
+      `You’ll then need to create a checkout on your development shop, and append this query string to the first page of checkout,`,
+    );
+    log(
+      `replacing TUNNEL_URL with your own ngrok URL: \`?dev=https://TUNNEL_URL/query\``,
```
vividviolet, line 256: Admin is using `?url=https://TUNNEL_URL/data` - should we make this consistent?

packages/argo-run/src/dev.ts
lemonmade: We already used `/data` for the post-purchase extension, and it has a different shape than what I'd like to provide here. We can probably move back to that route once we clean up the post-purchase dev flow, though personally I think `/data` is a rather unusual route name and `/query` leaves more room for making it a more powerful communication channel between the host app and CLI

packages/argo-run/src/dev.ts
lemonmade: I don't think it matters, personally. The shape you have in the admin is clearly too limited for the next step (multiple extensions in one project, so I wanted to jump ahead a bit and get a more flexible manifest shape. The fields could be made the same, but considering the host code will be custom for each right now, I am not too worried about that.

henrytao-me:
I cannot wait to see how the workspace project can unify both Argo and other project types. 🤔

kumar303 (approved):
It worked for me! 🎉  I was able to see the extension running in checkout from an ngrok URL.

I did not test if post purchase still works, FYI 😅

packages/argo-run/src/dev.ts
```diff
@@ -89,18 +91,115 @@ export async function dev(...args: string[]) {
...
     before(app) {
-      app.get('/', (_, res) => {
+      app.get('/', (req, res) => {
+        const protocol = req.headers['x-forwarded-proto'] ?? req.protocol;
+        const host = req.headers.host;
+        const origin = `${protocol}://${host}`;
+
         res.set('Content-Type', 'text/html');
-        res.send('<html>TODO</html>');
+        res.set('Cache-Control', 'no-cache');
```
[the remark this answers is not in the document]
lemonmade: Changed to `no-store` 👍

packages/argo-run/src/dev.ts
```diff
@@ -89,18 +91,115 @@ export async function dev(...args: string[]) {
...
+
+        res.json({
+          queryUrl: `${origin}${req.path}`,
+          extensions: [
+            {
+              scriptUrl: `${origin}${PUBLIC_PATH}${filename}`,
+              socketUrl: `${origin.replace(/^http/, 'ws')}${WEBSOCKET_PATH}`,
+              extensionPoint:
+                extension.type === 'checkout'
+                  ? extension.config.extensionPoints[0]
```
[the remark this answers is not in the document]
kumar303: It's going to be surprising that all but the first configured extension point is ignored here. I filed a bug so we don't forget to address it: https://github.com/Shopify/checkout-web/issues/4573

PR #71, opened 2021-04-07
Packages for developer preview

lemonmade:
This PR adjusts the Argo Checkout packages for our developer preview as discussed with the checkout extensibility team. It copies exactly what was in `@shopify/argo-checkout(-react)` into `@shopify/argo-post-purchase(-react)`, with all the necessary import renaming. It then copies what we had in our private version of `@shopify/argo-checkout(-react)` into this repo, which will become the new `0.10.0` version of `@shopify/argo-checkout`. I also added a short guide explaining the situation to post-purchase developers.

cc/ @jffortier @danrosenthal to make sure we are OK with the component diffs in `packages/argo-checkout`
cc/ @ecbrodie for post-purchase
cc/ @Fionoble


packages/argo-checkout-react/README.md
```diff
@@ -0,0 +1,92 @@
...
+    >
+      Log extension point to console
+    </Button>
+  );
+}
+```
+
+## Other React-specific APIs
+
+### `useExtensionInput()`
```
tleunen: We need to rename it with `useExtensionApi` in this section :)

documentation/upgrade/argo-checkout-to-argo-post-purchase.md
```diff
@@ -0,0 +1,13 @@
+# Upgrading from `@shopify/argo-checkout(-react)` to `@shopify/argo-post-purchase(-react)`
+
+In `@shopify/argo-checkout` and `@shopify/argo-checkout-react` version 0.10.x, we have removed the post-purchase extensions temporarily as we work towards broader availability of checkout extensions. The post-purchase APIs you import from `@shopify/argo-checkout` have been moved to `@shopify/argo-post-purchase` (or `@shopify/argo-post-purchase-react` for the React bindings). These new packages start at the same semantic version as the last release of the `@shopify/argo-checkout` package that contained these APIs. To upgrade, first make sure your extension works on the latest version of `@shopify/argo-checkout(-react)`. Once you have confirmed that, swap out your dependency on `@shopify/argo-checkout(-react)` with the matching post-purchase package:
```
tleunen, line 3: > These new packages start at the same semantic version as the last release of the `@shopify/argo-checkout` package that contained these APIs.

This is not true anymore because you reset the versions in a more recent commit, right?
But yup, I think we should indeed keep the same versions as it could help people understanding it's really the same code. The only change is the name.

> To upgrade, first make sure your extension works on the latest version of `@shopify/argo-checkout(-react)`.

Should we be more explicit and specify "The latest 0.9.x version"?

lemonmade: Good callout about "the latest 0.9.x version", will change that. The post-purchase packages will be the same versions as argo-checkout once I actually publish, I just had to set it to a lower version because our versioning flow asks you to pick a new, higher, version for each new or changed package.

jffortier:
I reviewed the diffs to argo-checkout components and argo-checkout-react and its good. Only thing we'll need to push _before_ the new version is the recent changes to layout the responsive props. The [PR is approved](https://github.com/Shopify/checkout-web/pull/4480), so it's a question of hours before its on main. We can take care of the update here, but if we can coordinate the publishing, that would be great.

packages/argo-checkout-react/README.md
lemonmade: Great catch!

kumar303 (commented):
Can you get this extension to work? 
[extension-scratch-dev.zip](https://github.com/Shopify/argo-checkout/files/6281797/extension-scratch-dev.zip)

When I link all modules to your `argo-checkout` branch, it fails to load on the checkout page (using `?dev=<ngrok_url>`) with this error in `@remote-ui/rpc`:

```
Uncaught (in promise) TypeError: Object(...) is not a function
```

When I link all packages to those in `checkout-web`, it renders the extension as expected.

<hr />

I was running the extension from that directory like:

```
yarn argo-run dev
```

... and I had `ngrok` running in a separate terminal to that port.

To link the packages, I was doing this from the extension directory:

```
pushd ~/src/github.com/Shopify/argo-checkout
cd packages/argo-run
yarn link
...etc
popd
yarn link @shopify/argo-checkout-react @shopify/argo-checkout @shopify/argo-run
```

<details>
<summary>Unhelpful screenshot of stack trace</summary>

Screen Shot 2021-04-08 at 4 58 39 PM


</details>

PR #72, opened 2021-04-14
Add `paths` arg to components renderer

iainmcampbell:
tldr the script now takes a `paths` object to allow for usage in multiple repos. I've only implemented this in `components()` for now (need to do some additional thinking about how we want to output extension point docs for Admin, as it looks a bit different than Checkout).

You should be able to merge this in without breaking anything. There should be no changes to the output: running `yarn run:ts scripts/generate-docs.ts` in this branch should create the same output as the exact same as the existing docs script (actually there’s one `\n` added to the `argo-checkout/index.md` file 😂).

### DIFF
--- a/scripts/typedoc/typedoc.ts
+++ b/scripts/typedoc/typedoc.ts
@@ -12,9 +12,25 @@ import type {
 } from './types';
 import {createDependencyGraph, Module} from './utilities/dependency-graph';
 
-extensionPoints();
-components();
-gettingStarted();
+interface Packages {
+  [key: string]: string;
+}
+
+interface Paths {
+  root: string;
+  /**
+   * Output path (in shopify-dev repo)
+   */
+  devDocsRoot: string;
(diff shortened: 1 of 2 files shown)


scripts/typedoc/typedoc.ts
```diff
@@ -208,6 +245,25 @@ async function components() {
...
+  url: string;
+  title: string;
+  description?: string;
+  hidden: boolean;
+}
+
+function renderYamlFrontMatter(frontMatter: FrontMatter) {
+  let matter = '---\n';
+
+  (Object.keys(frontMatter) as (keyof FrontMatter)[]).forEach((key) => {
```
iainmcampbell, line 259: I probably don't want to know why Typescript won't automatically type the result of `Object.keys(object)` as `keyof` the type of the object, which it knows

tmlayton (approved):

PR #73, opened 2021-04-15
Parse and render tsdoc tags in PropertySignature nodes

iainmcampbell:
Screen Shot 2021-04-15 at 5 37 29 PM

note: it turns out the `@default` we've been using is supposed to be `@defaultValue`, so you'll need to find+replace in the component source before this will work.

This should work with multiple tags, and tags without a value like `@beta`.

Should be able to merge this into `extract-docs` without breaking anything.


tmlayton (approved):
❤️

PR #74, opened 2021-04-19
Merge argo admin

vividviolet:
### TODOs before merging
- [ ] Delete unused files
- [ ] Remove commit lint?
- [ ] Update shared files inside .github like CODEOWNERS, ISSUE_TEMPLATE, pull_request_template.md
- [ ] Update shared files (lerna.json)


iainmcampbell (commented):
should we have a root-level `CHANGELOG`? it's just an amalgamation of changes from `argo-admin*` right?

I feel like a bunch of the other root-level files will probably have unwanted effects on checkout too. they've been working without a lot of root-level config, maybe we can too?

we should probably also consider either `scripts/argo-admin` or renaming a bunch of scripts (ie `scripts/new-admin-extension-api`)

vividviolet:
@iainmcampbell Yeah, the root-level CHANGELOG just combines the other change logs. If we're getting rid of commit hooks I think we should remove this one as well as we'd have to manually generate it.

> we should probably also consider either scripts/argo-admin or renaming a bunch of scripts (ie scripts/new-admin-extension-api). some scripts may end up being shared, but at least for now all of them are admin/checkout specific, right? even the docs generation script has admin/checkout specific config

I'll move them into `scripts/argo-admin` for now. We can do a follow up PR to clean it up later

vividviolet:
Closing this in favour the other PR with ALL commits from argo-admin: 
https://github.com/Shopify/argo-checkout/pull/75

PR #76, opened 2021-04-22
Fix React builds for post-purchase

lemonmade:
We are currently detecting whether we should apply React transformations on the basis of whether the developer uses the `@shopify/argo-checkout-react` package. When I added the new `argo-post-purchase-react` package, I forgot to also check for it. This PR fixes that :)

### DIFF
--- a/packages/argo-run/src/utilities.ts
+++ b/packages/argo-run/src/utilities.ts
@@ -34,17 +34,22 @@ export function log(message: string, {error = false} = {}) {
   console.log(`🔭 ${separator} ${message}`);
 }
 
+const REACT_ARGO_PACKAGES = [
+  '@shopify/argo-checkout-react',
+  '@shopify/argo-post-purchase-react',
+];
+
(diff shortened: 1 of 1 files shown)


kumar303 (approved):
Huh, TIL. Makes sense.

joeyfreund (approved):
Thank you!

I'm not sure how to 🎩 , but the code changes LGTM.

WDYT, is there anything we can put in CI to catch such an error in the future, or is it just a one-off?

lemonmade:
@joeyfreund we probably should have tests, but at this point it's not clear which of them would be throwaway since we'll want to have some shared approach between admin and checkout here.

joeyfreund:
> we probably should have tests, but at this point it's not clear which of them would be throwaway since we'll want to have some shared approach between admin and checkout here.

👍  Sounds good.

/cc @vividviolet , just wanted to make sure you're aware of it, since you're working on combining the libraries into a mono-repo.

PR #77, opened 2021-04-23
Merge argo admin all 2

vividviolet:
Bring argo-admin and argo-admin-react into the monorepo


.gitignore
```diff
@@ -10,3 +10,28 @@ packages/*/*.js
...
+build
+
+# Debug log from npm
+**/npm-debug.log
+
+# Debug log from lerna
+lerna-debug.log
+
+# Debug logs from yarn
+**/yarn-error.log
```
Fionoble, line 26: Line 20, 23, 26 and 32 would probably get addressed by Line 12.

.gitignore
```diff
@@ -10,3 +10,28 @@ packages/*/*.js
...
+
+# IntelliJ IDE ignore
+.idea
+
+tmp
+**/test.log
+
+tsconfig.tsbuildinfo
+tsconfig.prod.tsbuildinfo
```
Fionoble, line 35: Does this impact the `@sewing-kit` package at all?

Fionoble (approved):
A couple of non-blocking comments.

elanalynn (approved):
This is a lot 😅  but nothing odd is sticking out. Nice work!

.github/CODEOWNERS
```diff
@@ -0,0 +1,9 @@
+# Owners of docs
+*.md @iainmcampbell
+
+# Owners of the generators
+scripts/* @vividviolet @henrytao-me @alanthai @Fionoble @elanalynn @iainmcampbell
+
+# Owners of the packages.
+packages/argo-admin/* @vividviolet @henrytao-me @alanthai @Fionoble @elanalynn @iainmcampbell
+packages/argo-admin-react/* @vividviolet @henrytao-me @alanthai @Fionoble @elanalynn @iainmcampbell
```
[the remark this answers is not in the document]
vividviolet: I'll delete it for now

.gitignore
vividviolet: I don't think so - `.tsbuildinfo` will be recreated if it doesn't exist. It shouldn't be checked in since we want to do a fresh build when making a release

.github/workflows/ci.yml
```diff
@@ -7,118 +7,118 @@ jobs:
...
-      name: Restore TypeScript cache
-      with:
-        path: .sewing-kit/cache/typescript/
-        key: ${{ runner.os }}-typescript-v2-${{ github.sha }}
-        restore-keys: |
-          ${{ runner.os }}-typescript-v2-
-
-    - name: Type check
-      run: yarn type-check
+      - uses: actions/checkout@v1
```
lemonmade, line 10: Did more than just formatting change in this file?

.gitignore
lemonmade: Why are these going to the root of the project?

.gitignore
```diff
@@ -10,3 +10,12 @@ packages/*/*.js
 packages/*/bin
 !packages/*/.eslintrc.js
 *.log
+
+#TS output
+packages/**/dist
```
lemonmade, line 15: This is only for the admin Argo, right? Will we be aligning the build pipelines?

railgun.yml
```diff
@@ -0,0 +1,17 @@
+# https://development.shopify.io/tools/dev/railgun/Railgun-Config
```
lemonmade, line 1: Do we still need this file?

scripts/build-consumer.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/sh
```
lemonmade, line 1: Will this script work for checkout, too? If not, can we make it clear it only works with the admin?

lemonmade (approved):

scripts/build-consumer.sh
vividviolet: Yes, it does work with Checkout  - you can test it out :)

.gitignore
vividviolet: Yes, I'll be doing that in the next PR so that we build our packages the same way

PR #83, opened 2021-04-27
Admin use sk subfolders

vividviolet:
### Background
 - Run Admin tests using @sewingkit
-  Run Build using @sewingkit
-  Generate esnext build for argo-admin Generate esnext files for argo-admin and argo-admin-react

### Solution
Update the configs for argo-admin and argo-admin-react so we can use the same build tools.

### 🎩

**Argo Admin**
1. In a new directory run `shopify create extension` and follow the prompts in order to create a Product Subscription extension
1. Select the `typescript` template
1. Navigate to the new extension directory and run `shopify build`. Jot down the compiled bundle size - it should be **~40KB**
1. In the `argo-checkout` repo, remove all built files and start fresh by running `yarn clean; yarn clear-cache`. **This step is important**
1.  Run build-consumer for `argo-admin` with the relative path to your extension directory: `yarn build-consumer "../../extensions/testnewbuild" argo-admin`
1. Run `shopify serve` and verify that the Simulator is able to render the extension as expected. You might need to manually remove the `SHOP` from your local `.env` to get the Simulator to run.
1. Run `shopify build` and verify that the extension can be built
1. Inspect the `main.js` compiled code and verify that the bundle size is now reduced to **~14KB**. We should have a size reduction now that we're consuming the `esnext` files for argo-admin.

**Argo Admin React**

1. Select the `typescript-react` template
1. Navigate to the new extension directory and run `shopify build`. Jot down the compiled bundle size - it should be **~139KB**
1. In the `argo-checkout` repo, remove all built files and start fresh by running `yarn clean; yarn clear-cache`. **This step is important**
1.  Run build-consumer for  both `argo-admin` and `argo-admin-react` with the relative path to your extension directory: `yarn build-consumer "../../extensions/testnewbuildreact" argo-admin`
1. Run `shopify serve` and verify that the Simulator is able to render the extension as expected. You might need to manually remove the `SHOP` from your local `.env` to get the Simulator to run.
1. Run `shopify build` and verify that the extension can be built
1. Inspect the `main.js` compiled code and verify that the bundle size is now reduced to **~105KB**. We should have a size reduction now that we're consuming the `esnext` files for argo-admin.

### Checklist


packages/argo-admin-react/src/test/api.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  it('extend calls extend from argo-admin', () => {
+    const callback = () => {};
+    jest.spyOn(argoAdmin, 'extend');
+
+    extend('Playground', callback);
+
+    expect(argoAdmin.extend).toHaveBeenCalledWith('Playground', callback);
+  });
+});
```
Fionoble, line 64: 🙏

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "argo-checkout",
   "version": "0.0.0",
   "private": true,
-  "repository": "https://github.com/lemonmade/remote-ui.git",
+  "repository": "https://github.com/Shopify/argo-checkout.git",
```
Fionoble, line 5: 😄

packages/argo-admin/src/extension-api/index.ts
```diff
@@ -1,19 +1,6 @@
...
-export {isSessionTokenApi} from './sessionToken';
-
-export type {ToastApi} from './toast';
-export {isToastApi} from './toast';
+export * from './layout';
+export * from './locale';
+export * from './container';
+export * from './data';
+export * from './sessionToken';
+export * from './toast';
```
Fionoble, line 6: I think this causes problems for the tsdoc generator. cc: @iainmcampbell

config/sewing-kit/index.ts
```diff
@@ -11,7 +11,7 @@ import type {} from '@sewing-kit/plugin-jest';
 
 const PLUGIN = 'ArgoCheckout.DefaultPackage';
 
-export function argoCheckoutPackage({react: useReact = false} = {}) {
+export function argoPackage({react: useReact = false} = {}) {
```
Fionoble, line 14: Why the name change?

Fionoble (commented):
Haven't 🎩 'd yet

packages/argo-admin/src/extension-api/index.ts
iainmcampbell: yeah, it will break tsdoc. would be good if we can keep this file as it was previously. did `tsc` do this reformatting automatically? I had problems with `tsc` reformatting my `components/index.ts` file like this too, very annoying

config/sewing-kit/index.ts
vividviolet: Because we're using the same util for Admin now

packages/argo-admin/src/extension-api/index.ts
vividviolet: No, this was unrelated to auto-formatting. The change was made because we are missing some required exports argo-admin-react was relying on when nested subfolder import is no longer supported.

@iainmcampbell can you walk me through what tsdoc wants? Is this format with multiple type exports ok?

```
export type {LayoutApi, Layout} from './layout';
export {isLayoutApi} from './layout';
````

iainmcampbell: > Is this format with multiple type exports ok?

yeah, the only thing tsdoc doesn't like is `export *`. if your exports are all named (`{LayoutApi, Layout}`) you're good 👌

iainmcampbell (approved):
tested both typescript and react-typescript, working 👌

elanalynn (approved):
🎩

PR #84, opened 2021-04-29
comment out CI typescript cache

iainmcampbell:
let's see how it impacts CI runtime


iainmcampbell:
before:

type-check takes ~35s

after:

type-check takes 1m 31s

PR #85, opened 2021-04-29
Admin docs inlining part 2

iainmcampbell:
Part of https://github.com/Shopify/argo-private/issues/1507

export shared types: since the docs generator script starts parsing component types from `src/components/index.ts`, this puts these shared types where it can access them.

inline API docs: same as we did for components (https://github.com/Shopify/argo-checkout/pull/82)


packages/argo-admin-react/src/extension-api/data/examples/Data.example.tsx
```diff
@@ -0,0 +1,21 @@
...
+      - variant id: ${variantId}
+    `}
+    </Text>
+  );
+}
+
+extend(
+  'Admin::Product::SubscriptionPlan::Add',
+  render(() => <App />),
+);
```
elanalynn, line 21: It may worth adding the EOF new lines to all these new files so they don't get modified with unrelated code changes later.

elanalynn (approved):
This is great (and he build works 🎩)!

Fionoble (approved):
Nice 👍

packages/argo-admin-react/src/extension-api/data/examples/data.example.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import {extend, render, Text, useData} from '@shopify/argo-admin-react';
+
+function App() {
+  const {productId, variantId} = useData();
```
vividviolet, line 5: We should be passing the extension point here for type safety
`const {productId, variantId} = useData<Admin::Product::SubscriptionPlan::Add>();`

packages/argo-admin/src/extension-api/container/index.ts
```diff
@@ -1,9 +1,12 @@
-import {ExtensionPoint, ExtensionApi} from '../extension-points';
+import {ExtensionPoint, ExtensionApi} from '../../extension-points';
```
vividviolet, line 1: No docs for this?

vividviolet (commented):
Can we add docs for `useExtensionApi`? It's currently inside `utils`

packages/argo-admin-react/src/extension-api/data/examples/data.example.tsx
iainmcampbell: so far, I've been writing the examples using typescript but not actually doing any type annotations. this way, vscode catches obvious issues (like the fact that we had extensionPoint imported for every example) but the examples show up in 'JS'. probably should just have tsx and jsx versions for each though…

packages/argo-admin/src/extension-api/container/index.ts
iainmcampbell: I'm still thinking about how to document extension points. They're not near as simple as component, or even APIs, for which listing the props + a brief intro pretty much covers.

"listing the props" for an extension point mean something more like a list of allowable components and APIs, plus the actual props for the data API. however, the way we've implemented extension points is quite complicated to parse with tsdoc. I'm treating it as a separate task, and thinking about how automated it should really be.

iainmcampbell: tldr this is my next task after the other APIs

packages/argo-admin-react/src/extension-api/data/examples/data.example.tsx
vividviolet: For the `useData` and `useContainer` hooks you need to explicitly pass in the extension point otherwise the data object comes back as `any`. I don't think we want to encourage this? We want people to be able to get type safety for the data.

packages/argo-admin/src/extension-api/container/index.ts
iainmcampbell: same problem as `data` - the way we've typed the containers using generics on the extension point is quite hard to parse with tsdoc. there's like 3 levels of tracing generics to get from this file to

```
data: {
  productId: string;
  variantId?: string;
};
```

packages/argo-admin-react/src/extension-api/data/examples/data.example.tsx
iainmcampbell: yeah, we should just have JS and TS examples. i'll add them.

Can you get type safety using the typescript vanilla API?

packages/argo-admin/src/extension-api/container/index.ts
iainmcampbell: we also didn't have an example for useContainer in the existing docs: https://github.com/Shopify/argo-admin-template/blob/master/docs/Utilities/Container.md

packages/argo-admin-react/src/extension-api/data/examples/data.example.tsx
vividviolet: @iainmcampbell you get it for free in vanilla without doing anything. Only the hooks require manually specifying the extension point.

image

packages/argo-admin/src/extension-api/container/index.ts
vividviolet: Maybe we should remove the docs for `useData` to be consistent then?

packages/argo-admin-react/src/extension-api/data/examples/data.example.tsx
iainmcampbell: ok so the way we have it in the template (define a named function, then pass that to `extend()` does require you to manually specify the extension point:

Screen Shot 2021-05-05 at 1 57 11 PM

Screen Shot 2021-05-05 at 1 57 23 PM

but if you pass an anonymous function to `extend()` you do get the types for free.
Screen Shot 2021-05-05 at 1 57 58 PM

packages/argo-admin/src/extension-api/container/index.ts
iainmcampbell: i'm gonna write an example, better to have than not.

iainmcampbell:
@vividviolet added examples for data and container

vividviolet (approved):
Looks great! 👏

PR #91, opened 2021-05-06
Support mulitple extension points for checkout

henrytao-me:
### Background

Current `/query` endpoint doesn't returns all extension points. This PR fixes it. 

Resolves https://github.com/Shopify/checkout-web/issues/4573

### Solution

### 🎩

- In https://github.com/Shopify/argo-checkout-template, run `yarn && yarn generate --type=CHECKOUT_ARGO_EXTENSION --template=react-typescript`
- In this PR, run `yarn && yarn build && cp -R packages/* ../argo-checkout-template/node_modules/@shopify/`
- Back to argo-checkout-template repo:
  + Update `extension.config.yml` 
  ```
  ---
  extension_points:
    - Checkout::Feature::Render
    - Checkout::KitchenSink
  ```
  + Run `yarn server`
- Open http://localhost:39351/query, verify that extensions field returns an array of two.

### Checklist

### DIFF
--- a/packages/argo-run/src/dev.ts
+++ b/packages/argo-run/src/dev.ts
@@ -179,30 +179,18 @@ export async function dev(...args: string[]) {
         res.set('Access-Control-Allow-Origin', '*');
         res.set('Cache-Control', 'no-cache');
 
-        const defaultExtension = {
-          scriptUrl: `${origin}${PUBLIC_PATH}${filename}`,
-          socketUrl: `${origin.replace(/^http/, 'ws')}${WEBSOCKET_PATH}`,
-        };
-        let extensions;
-        if (extension.type === 'checkout') {
-          extensions = extension.config.extensionPoints.map(
-            (extensionPoint) => ({
-              ...defaultExtension,
-              extensionPoint,
-            }),
-          );
-        } else {
-          extensions = [
-            {
-              ...defaultExtension,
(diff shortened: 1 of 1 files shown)


packages/argo-run/src/dev.ts
```diff
@@ -185,10 +185,10 @@ export async function dev(...args: string[]) {
             {
               scriptUrl: `${origin}${PUBLIC_PATH}${filename}`,
               socketUrl: `${origin.replace(/^http/, 'ws')}${WEBSOCKET_PATH}`,
-              extensionPoint:
+              extensionPoints:
                 extension.type === 'checkout'
-                  ? extension.config.extensionPoints[0]
-                  : 'Checkout::PostPurchase::Render',
+                  ? extension.config.extensionPoints
+                  : ['Checkout::PostPurchase::Render'],
```
henrytao-me, line 191: Considering this `/query` endpoint is not being used for PostPurchase, I think it's safe to change to `extensionPoints`.

kumar303 (approved):
Looks good to me, thanks. Just be sure to coordinate a patch in `checkout-web` to use the new key:

https://github.com/Shopify/checkout-web/blob/2059c1df3cccfe77692a16a9b0330e97e2548946/app/utilities/development-mode/context.ts#L6

/cc @alxclark as he may be in the process of adding post purchase support ?

PR #102, opened 2021-05-18
Update to latest argo-checkout & argo-checkout-react definitions

tleunen:
This PR updates `argo-checkout` and `argo-checkout-react` to their latest definitions.


lemonmade (approved):
Thanks Tommy! I left a few comments on documentation, probably fine to fix them at the source instead and they can be part of the next changes we publish.

PR #116, opened 2021-05-25
Remove unused GeographicalCoordinates type

lemonmade:
Removes an unused type in our Argo API related to geographical coordinates, which we took off the address type.


lemonmade:
Meant to do this on our internal repo, whoops :S

PR #118, opened 2021-05-26
Typo fix for button description

owenanderson:
### Background

Button description has one to many or's.

> Buttons are used for actions, such as “Add”, “Continue”, or “Pay now”, or “Save”.

### Solution
Buttons are used for actions, such as “Add”, “Continue”, “Pay now”, or “Save”.


tmlayton (approved):
:shipit:

PR #119, opened 2021-05-27
[post-purchase] Adds redirectUrl to the ApplyChangesetResult type

ecbrodie:
### Background

This PR adds a `redirectUrl` attribute to the `ApplyChangesetResult` type for post-purchase.

This new attribute is representing what is already being done at present with the `ApplyChangeset` action. In order to communicate to the frontend where to redirect the user for 3DS authentication in Checkout Post Purchase, we send a `redirectUrl` attribute alongside `errors` and `status`.

### Solution

**Solution section deleted due to this PR being closed.**

### 🎩

There does not seem to be any tophatting that can be applied here, besides a clean CI pipeline.

### Checklist


ecbrodie:
Alex and I chatted further about this PR. We realized that we probably do not have a need to reveal `redirectUrl` to our partners. 3DS redirects are a detail that we manage internally in our checkout codebase. Therefore, I am closing this.

PR #120, opened 2021-05-28
Fix webSocket for liveReload in Safari

henrytao-me:
### Background

This PR fixes webSocket for liveReload in Safari. Currently, argo-run hardcode localhost for webSocket. It works in Chrome but Safari doesn't like it due to insecure connection. 

Resolves https://github.com/Shopify/checkout-web/issues/5179

### Solution

Shopify cli already pass `publicUrl` argument. It's used to construct the scriptUrl. We can use it to construct webSocket url too. 

### 🎩

- Download latest shopify-app-cli
- Create new checkout extension with shopify cli
- Run `shopify serve`
- Open ngrok url, ex: https://fd1df9e54b46.ngrok.io/query. Verify that socket url looks like `"socketUrl":"wss://fd1df9e54b46.ngrok.io/build"`
- Run `yarn start`
- Open localhost url, ex http://localhost:8910/query. Verify that socket url looks like `"socketUrl":"ws://localhost:8910/build"`
- Add a product and checkout the cart. You can use this shop for testing https://shop1.shopify.remove-line-item.henry-tao.us.spin.dev/
- Verify that the extension is reloaded as normal in both Chrome and Safari.

### Checklist

### DIFF
--- a/packages/argo-run/src/dev.ts
+++ b/packages/argo-run/src/dev.ts
@@ -29,9 +29,8 @@ export async function dev(...args: string[]) {
   const publicUrl = namedArgument('publicUrl', args);
   const tunnelStarted = Boolean(publicUrl);
   const filename = 'extension.js';
-  const script = new URL(`${PUBLIC_PATH}${filename}`, publicUrl || url);
-  const scriptUrl = script.toString();
-  const isHttps = script.protocol === 'https:';
+  const scriptUrl = new URL(`${PUBLIC_PATH}${filename}`, publicUrl || url);
+  const isHttps = scriptUrl.protocol === 'https:';
 
   const compiler = webpack(
     createWebpackConfiguration({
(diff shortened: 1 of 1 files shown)


packages/argo-run/src/dev.ts
```diff
@@ -29,10 +29,9 @@ export async function dev(...args: string[]) {
   const publicUrl = namedArgument('publicUrl', args);
   const tunnelStarted = Boolean(publicUrl);
   const filename = 'extension.js';
-  const scriptUrl = new URL(
-    `${PUBLIC_PATH}${filename}`,
-    publicUrl || url,
-  ).toString();
+  const script = new URL(`${PUBLIC_PATH}${filename}`, publicUrl || url);
+  const scriptUrl = script.toString();
```
lemonmade, line 33: Can we make the `URL` object be called `scriptUrl`, and change what was `scriptUrl` to be `scriptHref` or just do `scriptUrl.toString()`/`scriptUrl.href` when we need it?

lemonmade (approved):
Looks great. I think we probably also need to update the `socketUrl` field here: https://github.com/Shopify/argo/pull/120/files#diff-5f341c6a044f1652c65001b022ed94206d6fc435c7536d874555c3759a17fbfeR195? Can you open a separate issue for the other thing we mentioned around this, how we should probably just stop having the sandbox connect to the socket itself, and instead have checkout connect? As long as we don't have real hot reloading, it will be more performant and straightforward to have the checkout manage these connections.

PR #125, opened 2021-06-02
argo-admin-react: fix component props exports

vividviolet:
### Background

Related to esnext issue https://github.com/Shopify/argo-private/issues/1524#issuecomment-852318742

### Solution

TS is very strict when it comes to exporting from another package. argo-admin-react re-exports component props from argo-admin so we need to explicitly set these as a type export

### 🎩

1. Run `yarn build` and verify that both argo-admin-react and argo-admin compiles

### Checklist


henrytao-me (approved):
🎩-ed with https://github.com/Shopify/argo-admin/pull/94. LGTM

PR #127, opened 2021-06-03
Use 'export type' in argo-admin

iainmcampbell:
### Background

In the `main` branch, run `yarn clean; yarn clear-cache; yarn build`. In `packages/argo-admin/build/esm`, you’ll see import statements referencing empty files. This prevents argo-admin from being used with skypack: <https://cdn.skypack.dev/@shopify/argo-admin>.

### Solution

Always use `export type` when exporting types, so the bundler knows to strip them out.

### 🎩

Run `yarn clean; yarn clear-cache; yarn build` on this branch and you should see no references to types in `build/esm`. ie `build/esm/index.mjs` shouldn't reference `extension-points` at all, as only types are exported from there.

### Checklist

### DIFF
--- a/packages/argo-admin-react/src/index.ts
+++ b/packages/argo-admin-react/src/index.ts
@@ -2,5 +2,7 @@ export * from '@shopify/argo-admin/extension-points';
 export * from '@shopify/argo-admin/extension-api';
 
 export * from './api';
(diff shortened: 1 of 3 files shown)


henrytao-me (approved):
LGTM

PR #135, opened 2021-06-11
add BaseButtonProps

elanalynn:
### Background

Closes: https://github.com/Shopify/argo-private/issues/1487
Relates to: https://github.com/Shopify/argo-rfcs/issues/2

### Solution

This adds the `BaseButtonProps` decided on in this RFC discussion. These props are not yet consumed by either ui-extensions library. 

### 🎩

- This API is not yet consumed and has no impact on the ui-extensions libraries.

### DIFF
--- a/packages/ui-extensions/src/types/Button.ts
+++ b/packages/ui-extensions/src/types/Button.ts
@@ -9,7 +9,7 @@ export interface BaseButtonProps {
    * `primary`: button used for main actions or green-path. Ex: Continue to next step, Discount field
    * `secondary`: button used for secondary actions not blocking user progress. Ex: Download Shop app
    * `plain`: Renders a button that visually looks like a Link
-   * @default 'primary'*/
+   */
   kind?: 'primary' | 'secondary' | 'plain';
   /**
    *  Specifies the color of the Button. The button will keep the style of the chosen `kind`,
@@ -28,7 +28,7 @@ export interface BaseButtonProps {
   /**
    * Whether the button should fill all available inline space.
    * */
-  fill?: boolean;
+  inlineSize?: 'fill';
   /**
    * Replaces content with a loading indicator
    */
--- a/packages/ui-extensions/src/types/Button.ts
+++ b/packages/ui-extensions/src/types/Button.ts
@@ -1,0 +1,58 @@
+export interface BaseButtonProps {
+  /**
+   * Destination to link to, renders a Link
+   */
+  to?: string;
+  /**
+   * The type of button that will be rendered.
+   *
+   * `primary`: button used for main actions or green-path. Ex: Continue to next step, Discount field
+   * `secondary`: button used for secondary actions not blocking user progress. Ex: Download Shop app
+   * `plain`: Renders a button that visually looks like a Link
+   * @default 'primary'*/
+  kind?: 'primary' | 'secondary' | 'plain';
+  /**
+   *  Specifies the color of the Button. The button will keep the style of the chodsen `kind`,
+   *  but replace its color according to the appearance.
+   *
+   * `monochrome`: button will inherit the color of its parent
+   * `critical`: button will take inherit the color of the Critical color group (in Checkout)
+   *  		 and map to 'destructive' (in Admin/Polaris). Typically used for destructive actions.
+   *
+   *  Excludes 'success' | 'warning' since buttons in a Banner will be `monochrome`
+   *  and use the default colors set by the Banner.
+   */
+  appearance?: 'monochrome' | 'critical';
+  /** Changes the size of the button
+   * @default 'base'
+   */
+  size?: 'base' | 'large';
+  /**
+   * Whether the button should fill all available inline space.
+   * */
+  fill?: boolean;
+  /**
+   * Replaces content with a loading indicator
+   */
+  loading?: boolean;
+  /**
+   * A label replacing the loading indicator when user prefers reduced motion
+   */
+  loadingLabel?: string;
+  /**
+   * A label that will be announced to buyers using assistive technologies
+   */
+  accessibilityLabel?: string;
+  /**
+   * Disables the button, disallowing any interaction
+   */
+  disabled?: boolean;
+  /**
+   * Allows the button to submit a form
+   */
+  submit?: boolean;
+  /**
+   * Callback when pressed
+   */
+  onPress?(): void;
+}
--- a/packages/ui-extensions/src/types/Button.ts
+++ b/packages/ui-extensions/src/types/Button.ts
@@ -12,15 +12,13 @@ export interface BaseButtonProps {
    * @default 'primary'*/
   kind?: 'primary' | 'secondary' | 'plain';
   /**
-   *  Specifies the color of the Button. The button will keep the style of the chodsen `kind`,
+   *  Specifies the color of the Button. The button will keep the style of the chosen `kind`,
    *  but replace its color according to the appearance.
    *
    * `monochrome`: button will inherit the color of its parent
    * `critical`: button will take inherit the color of the Critical color group (in Checkout)
    *  		 and map to 'destructive' (in Admin/Polaris). Typically used for destructive actions.
    *
-   *  Excludes 'success' | 'warning' since buttons in a Banner will be `monochrome`
-   *  and use the default colors set by the Banner.
    */
   appearance?: 'monochrome' | 'critical';
   /** Changes the size of the button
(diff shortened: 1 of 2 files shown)


packages/ui-extensions/src/types/Button.ts
```diff
@@ -0,0 +1,58 @@
...
+   */
+  to?: string;
+  /**
+   * The type of button that will be rendered.
+   *
+   * `primary`: button used for main actions or green-path. Ex: Continue to next step, Discount field
+   * `secondary`: button used for secondary actions not blocking user progress. Ex: Download Shop app
+   * `plain`: Renders a button that visually looks like a Link
+   * @default 'primary'*/
+  kind?: 'primary' | 'secondary' | 'plain';
```
lemonmade, line 13: IMO this prop is too design-system specific, there's no guarantee checkout will be able to frame the different button types this way

packages/ui-extensions/src/types/Button.ts
```diff
@@ -0,0 +1,58 @@
+export interface BaseButtonProps {
+  /**
+   * Destination to link to, renders a Link
+   */
+  to?: string;
```
lemonmade, line 5: I feel like there should be some conversation about this one. There are many different options — `url` (as Polaris uses), `to` (as I think Checkout uses), `href`, `link`, ... I think it's also important to consider how this prop could potentially change in the future, including in ways that may diverge between the hosts. In particular, I think it's likely we will want a way of providing a `to` that doesn't hardcode assumptions about URL structure (that is, `to` would not just be an arbitrary path/ URL). This is really important in checkout, where buttons from extensions will need to respect the fact that checkouts can sometimes be on a single page, or spread across multiple pages, depending on the merchant theme.

packages/ui-extensions/src/types/Button.ts
```diff
@@ -0,0 +1,58 @@
...
+   *  but replace its color according to the appearance.
+   *
+   * `monochrome`: button will inherit the color of its parent
+   * `critical`: button will take inherit the color of the Critical color group (in Checkout)
+   *  		 and map to 'destructive' (in Admin/Polaris). Typically used for destructive actions.
+   *
+   *  Excludes 'success' | 'warning' since buttons in a Banner will be `monochrome`
+   *  and use the default colors set by the Banner.
+   */
+  appearance?: 'monochrome' | 'critical';
```
lemonmade, line 25: IMO this is also an admin design system prop, I also find it hard to understand the difference between this and `kind` (neither prop seems to imply it adds semantic meaning, so they all just seem like visual variations)

packages/ui-extensions/src/types/Button.ts
```diff
@@ -0,0 +1,58 @@
...
+  /** Changes the size of the button
+   * @default 'base'
+   */
+  size?: 'base' | 'large';
+  /**
+   * Whether the button should fill all available inline space.
+   * */
+  fill?: boolean;
```
lemonmade, line 33: We do have this in the button props, but I am not sure it fits with some of the more recent APIs we've done. In a lot of those APIs, we have left ourselves more room by defining `blockSize`/ `inlineSize` props with union types, which I expect we'd use for `Button` too when we do a polish pass on our components (e.g., `<Button inlineSize="fill" />`). I think this impacts `size` as well, which I think will be confusing if we have `blockSize`/ `inlineSize` as well. cc/ @jffortier @jun-shop

packages/ui-extensions/src/types/Button.ts
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Replaces content with a loading indicator
+   */
+  loading?: boolean;
+  /**
+   * A label replacing the loading indicator when user prefers reduced motion
+   */
+  loadingLabel?: string;
```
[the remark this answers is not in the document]
lemonmade: This API is also a little awkward when we have a separate `loading` prop, I think we might look at consolidating the label as an optional property on an object you can pass to `loading` in Checkout... cc/ @jffortier

packages/ui-extensions/src/types/Button.ts
```diff
@@ -0,0 +1,58 @@
...
+   */
+  accessibilityLabel?: string;
+  /**
+   * Disables the button, disallowing any interaction
+   */
+  disabled?: boolean;
+  /**
+   * Allows the button to submit a form
+   */
+  submit?: boolean;
```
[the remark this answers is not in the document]
lemonmade: Yeah, this is also one I'd like to revisit in checkout. We've discussed `role`/ `accessibilityRole` props that I feel are a more flexible way of declaring this kind of additional semantic information cc/ @jffortier @danrosenthal @simontaisne

lemonmade (changes requested):

packages/ui-extensions/src/types/Button.ts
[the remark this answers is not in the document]
jffortier: Yes, I prefer `role` as well and is consistent with [View](https://github.com/Shopify/checkout-web/blob/master/packages/checkout-web-ui/src/components/View/View.tsx#L11). This is something we'll need in Pressable too (@jun-shop you have `kind?: button | submit | toggle` at the moment)

packages/ui-extensions/src/types/Button.ts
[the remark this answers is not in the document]
jffortier: TBH, this option is a nice touch, but I've been close to removing it multiple times. I leaned towards the safe side when I introduce this difference for user that prefers reduced motion. The spinner is such a small motion in the overall UI and I couldn't find any validation that it would cause any harms.

I'd be ok to remove it OR the object route is elegant too. Is this what you had in mind Chris?
`loading?: boolean | { status: boolean; accessibilityLabel?: string; }`

packages/ui-extensions/src/types/Button.ts
jffortier: `kind` in checkout defines the semantic of the button. It does come with some default visual styles, but they are editable by the merchant.

`appearance` is the visual prop. It is combined with the kind prop to offer more flexibility.

This is what is possible with the defaults styles of Checkout.
image

packages/ui-extensions/src/types/Button.ts
elanalynn: @lemonmade You liked this approach in April 😂 !  https://github.com/Shopify/argo-rfcs/issues/2#issuecomment-824058184. That said, I know things I things evolve. @jffortier addresses these concerns below. If these appearance is still editable by the merchant, does this sit better with you? On the Admin side, this maps pretty well with Polaris.  While they don't have `secondary`, but we can map this to the non-primary kind.

elanalynn:
@lemonmade @jffortier @vividviolet I addressed the comments and removed the props that we all agreed about (`submit`, and `loadingLabel`). Are we okay with moving forward with `to`, `kind`, `appearance`, and `fill`?

packages/ui-extensions/src/types/Button.ts
lemonmade: Hmm well the evidence is pretty overwhelming :P I think it will be pretty difficult to have a shared set of button types that are common across all our apps, but it seems like no one else is super worried about it so 👍

lemonmade: I'd at least remove the `@default` here though — I don't think this is the default we will want within checkout in the end, and in general defaults seem like a pretty host-specific concern.

packages/ui-extensions/src/types/Button.ts
lemonmade: I think I'm just pretty hesitant to commit to `boolean` props unless it is clear there will never be more than two options. If we choose `fill` for this, and later want a prop that is able to restrict the max size of the button or something similar to that, we are signed up for having a mix of attributes that do similar, but different, things. In contrast, using `inlineSize="fill"` here gives us tons of flexibility for the future to add more semantic sizes, or to allow for arbitrary sizes.

Don't feel super strongly about this one, but other than `disabled`, I haven't seen many props that are truly boolean in the long run.

lemonmade (approved):
These are well documented 👍 I have a few nits and concerns still but if they aren't shared, I am happy to have this 🚢 and we can see how it works out.

packages/ui-extensions/src/types/Button.ts
jffortier: I'm good with `inlineSize="fill"`. We can make that change in Checkout.

PR #140, opened 2021-06-15
Render markdown lists in docs builder

iainmcampbell:
### Background

Currently we strip newlines from jsdoc comments in a way that makes it impossible to render markdown lists, ie

```
/** Represents an error occurred while calculating or applying a changeset.
 *
 * **Possible values for `code`:** 
 *
 * - `payment_required` - The original payment method couldn't be charged.
 * - `insufficient_inventory` - An item requested to be added is out of stock.
 * - `changeset_already_applied` - Indicates if the changeset has been applied. Changesets can only be applied once.
 * - `unsupported_payment_method` - Indicates that the payment method was unsupported. The payment method used to pay for the initial purchase can't be used to pay for changesets.
 * - `invalid_request` - The request is invalid. It shouldn't be re-sent without modifications.
 * - `server_error` - An unexpected error has happened.
 */
```

### Solution

Stop using the microsoft string builder.

Unfortunately, `contentWithTags` doesn't pick up all tags yet, but I don't think that affects checkout - https://github.com/Shopify/argo/issues/139

### 🎩

- ...

### Checklist

### DIFF
--- a/scripts/typedoc/shopify-dev-renderer/shared.ts
+++ b/scripts/typedoc/shopify-dev-renderer/shared.ts
@@ -305,13 +305,23 @@ export function strip(content: string) {
   );
 }
 
+/**
+ * extract first sentence (line or up to '.') for yaml description field
+ */
 export function firstSentence(content: string) {
   const lines = content.split(/(\n|\. )/g);
   let firstSentence = lines.length ? lines[0] : content;
+
+  // try to split on period if first line doesn't have one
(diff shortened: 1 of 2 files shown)


tmlayton (approved):

andrewjtech123 (approved):
Thanks @iainmcampbell ! I lack the expertise to give this a proper review but this is the functionality that we need.

PR #143, opened 2021-06-17
Add BaseHeadingProps to shared types

elanalynn:
Closes: Shopify/argo-private#1498
Relates to: Shopify/argo-rfcs#9

### Solution
This adds the `BaseHeadingProps` to the shared `ui-extensions` package. These props are not yet consumed by either ui-extensions library.

This is a pretty straightforward addition to the shared APIs, considering that it is a new component for Admin and matches Checkout's current implementation. 

I do have one question for the Admin team that I overlooked during the RFC discussion.

What should the default heading level be?

In Checkout, the heading will use its “automatic” heading level, which is determined by the level of nesting within ancestor `HeadingGroup`. At this point, we have not discussed creating a `HeadingGroup` in Admin. Is it reasonable to make the default level `1`? (I know this is an implementation detail, but I want to make sure that this API works for the way we work in Admin).

### 🎩

- This API is not yet consumed and has no impact on the ui-extensions libraries.

### DIFF
--- a/packages/ui-extensions/src/types/Heading.ts
+++ b/packages/ui-extensions/src/types/Heading.ts
@@ -1,0 +1,20 @@
+type Level = 1 | 2 | 3;
+
+export interface BaseHeadingProps {
+  /**
+   * Unique identifier. Typically used to make the heading a target
+   * that another component can refer to in order to provide an alternative
+   * accessibility label.
+   */
+  id?: string;
+  /**
+   * The visual level of the heading.
+   */
+  level?: Level;
+  /**
+   * A custom accessibility role for this heading. Currently, only `presentation`
+   * is supported. `presentation` will strip the semantic meaning of the heading,
+   * but leave the visual styling intact.
+   */
+  role?: 'presentation';
+}
--- a/packages/ui-extensions/src/types/Heading.ts
+++ b/packages/ui-extensions/src/types/Heading.ts
@@ -1,4 +1,4 @@
-type Level = 1 | 2 | 3;
+export type Level = 1 | 2 | 3;
 
 export interface BaseHeadingProps {
   /**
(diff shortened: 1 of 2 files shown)


packages/ui-extensions/src/types/Heading.ts
```diff
@@ -0,0 +1,20 @@
+type Level = 1 | 2 | 3;
+
+export interface BaseHeadingProps {
+  /**
+   * Unique identifier. Typically used to make the heading a target
+   * that another component can refer to in order to provide an alternative
+   * accessibility label.
+   */
+  id?: string;
```
james-woo, line 9: This field will potentially be used by a partner?

packages/ui-extensions/src/types/Heading.ts
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * The visual level of the heading.
+   */
+  level?: Level;
+  /**
+   * A custom accessibility role for this heading. Currently, only `presentation`
+   * is supported. `presentation` will strip the semantic meaning of the heading,
+   * but leave the visual styling intact.
+   */
+  role?: 'presentation';
```
james-woo, line 19: Should this be an enum?

elanalynn: I think this could be an enum, but we use this string pattern everywhere ([Button](https://github.com/Shopify/argo/blob/8a4948ad1f2bdd4a71c9e5da331de57756f19541/packages/ui-extensions/src/types/Button.ts), [Banner](https://github.com/Shopify/argo/blob/8a4948ad1f2bdd4a71c9e5da331de57756f19541/packages/ui-extensions/src/types/Banner.ts), etc.). We are restricting the type to this exact string so it's pretty safe.

packages/ui-extensions/src/types/Heading.ts
elanalynn: Yes, a partner may want to use `aria-labelledby` and refer to this id.

packages/ui-extensions/src/types/Heading.ts
```diff
@@ -0,0 +1,20 @@
+type Level = 1 | 2 | 3;
```
lemonmade, line 1: We should export this type. In checkout we will want to redeclare the `level` prop, even though it is already on the shared type, so that we can add the additional documentation that is relevant for checkout's automatic heading level (https://github.com/Shopify/argo/blob/main/packages/argo-checkout/src/components/Heading/Heading.ts#L12-L18)

lemonmade (approved):
One small note but otherwise LGTM 👍

PR #147, opened 2021-06-21
Add documentation for error handling

henrytao-me:
### Background

This PR adds document for error handling. 

See https://github.com/Shopify/checkout-web/issues/5252#issuecomment-864227636

### DIFF
--- a/documentation/error-handling.md
+++ b/documentation/error-handling.md
@@ -83,7 +83,7 @@ self.onerror = (error) => {
 };
 ```
 
-By default, Bugsnag has already sent error for `unhandledrejection`. You only need to listen to `onerror`.
+By default, Bugsnag will listen for `unhandledrejection`. You only need to manually listen to `onerror`.
 
 ### Sentry
 
--- a/documentation/error-handling.md
+++ b/documentation/error-handling.md
@@ -1,56 +1,30 @@
 # Error handling
 
-Because UI extension scripts run in non-browser environments (Web Worker in web, JavascriptCore on Mobile), it is recommended to manually set up error listeners and error reporting. However, it is possible to use external libraries for error reporting.
+UI extensions support standard web errors. You can add `unhandledrejection` listener for unhandled promise rejections or `error` listener for other exceptions like Javascript runtime error or a resource fails to load.
 
-## Error handling in UI extensions
+```javascript
+// For unhandled promise rejections
+self.addEventListener('unhandledrejection', (error) => {
+  console.warn('event unhandledrejection', error);
+});
 
-UI extensions can throw two types of error: lifecycle errors and non-lifecycle errors.
+// For other exceptions
+self.addEventListener('error', (error) => {
+  console.warn('event error', error);
+});
+```
 
-### Lifecycle error
+## Reporting errors
 
-Most errors will fall into this type. These errors can be thrown when:
-
-- a UI extension script is loaded or imported
-- the UI extension script is run
-- during any callback interaction with UI extension components or APIs
-
-To catch these errors, add an event listener for `unhandledrejection`.
-
-  ```javascript
-  self.onunhandledrejection = (error) => {
-    console.warn('onunhandledrejection', error);
-  };
-  // or
-  self.addEventListener('unhandledrejection', (error) => {
-    console.warn('event unhandledrejection', error);
-  });
-  ```
-
-### Non-lifecycle error
-
-These errors can be thrown when the script encounters an exception while running a `setTimeout` callback. To catch these errors, add an event listener for `error`.
-
-  ```javascript
-  self.onerror = (error) => {
-    console.warn('onerror', error);
-  };
-  // or
-  self.addEventListener('error', (error) => {
-    console.warn('event error', error);
-  });
-  ```
-
-## Handle errors via 3rd-party libraries
+It's possible to use 3rd party libaries for reporting errors like Bugsnag or Sentry. However, it might require some extra configs or setup depending on which libraries you are using because UI extensions are run inside WebWorker.
 
 ### Bugsnag
 
-See https://docs.bugsnag.com/platforms/javascript/.
-
-You can run Bugsnag using the CDN version or npm dependencies. However, Bugsnag requires additional polyfills at runtime, as it will try to access the browser environment. The script below has been tested at the time of writing these docs.
+Follow [Bugsnag documentation](https://docs.bugsnag.com/platforms/javascript/) to install it to your extension. If you use [CDN version](https://docs.bugsnag.com/platforms/javascript/cdn-guide/), you need to add polyfill code below as Bugsnag tries to access some variables that are not available in WebWorker.
 
 ```javascript
 /**
- * Window polyfill
+ * CDN polyfill
  */
 (() => {
   const document = {
@@ -68,46 +42,23 @@ You can run Bugsnag using the CDN version or npm dependencies. However, Bugsnag 
   window.document = document;
   window.history = history;
 })();
-
-/**
- * Bugsnag
- */
-// Add bugsnag CDN javascript here
-Bugsnag.start({apiKey: 'YOUR-BUGSNAG-API-KEY'});
-
-/**
- * Add error handlers
- */
-self.onerror = (error) => {
-  Bugsnag.notify(error);
-};
 ```
 
-By default, Bugsnag will listen for `unhandledrejection`. You only need to manually listen to `onerror`.
+There is no need to add custom event listeners as Bugsnag does it for you out of the box.
 
 ### Sentry
 
-See https://docs.sentry.io/platforms/javascript/.
-
-Using Sentry for UI extensions is straightforward. You only need to import the main Sentry bundle or `@sentry/browser` No polyfills are required.
+Follow [Sentry documentation](https://docs.sentry.io/platforms/javascript/) to install it to your extension. There is no polyfill required for using Sentry. However, Sentry does not add any event listeners out of the box. So, you need to add them manually.
 
 ```javascript
-// Add Sentry CDN javascript here
-self.Sentry = Sentry;
-Sentry.init({
-  dsn: 'YOUR-SENTRY-DSN-VALUE',
-});
-
 /**
  * Add error handlers
  */
-self.onerror = (error) => {
+self.addEventListener('unhandledrejection', (error) => {
   Sentry.captureException(new Error(error.reason.stack));
-};
+});
 
-self.onunhandledrejection = (error) => {
+self.addEventListener('error', (error) => {
   Sentry.captureException(new Error(error.reason.stack));
-};
+});
 ```
-
-By default, Sentry does not listen for any errors. You will need to manually add listeners for both `error` and `unhandledRejection`.
--- a/documentation/error-handling.md
+++ b/documentation/error-handling.md
@@ -1,6 +1,6 @@
 # Error handling
 
-Because ui extensions script runs in non-browser environment (ex: WebWorker in Web, JavascriptCore on Mobile), it is recommended to listen and send errors to server manually. However, it is still possible to use external libraries for reporting errors. This document shows a couple of examples about how you can handle it.
+Because UI extension scripts run in non-browser environments (Web Worker in web, JavascriptCore on Mobile), it is recommended to manually set up error listeners and error reporting. However, it is possible to use external libraries for error reporting.
 
 ## The generic way of handling errors
 
--- a/documentation/error-handling.md
+++ b/documentation/error-handling.md
@@ -2,11 +2,19 @@
 
 Because UI extension scripts run in non-browser environments (Web Worker in web, JavascriptCore on Mobile), it is recommended to manually set up error listeners and error reporting. However, it is possible to use external libraries for error reporting.
 
-## The generic way of handling errors
+## Error handling in UI extensions
 
-There are two types of errors when using ui extensions: lifecycle and non-lifecycle error.
+UI extensions can throw two types of error: lifecycle errors and non-lifecycle errors.
 
-- **Lifecycle error:** most errors will fall into this type. It could happen when ui extensions script is loaded / imported, the script is run / rendered, or any interaction via callback of ui extensions components or apis. To catch this error, you can add event listener for `unhandledrejection`.
+### Lifecycle error
+
+Most errors will fall into this type. These errors can be thrown when:
+
+- a UI extension script is loaded or imported
+- the UI extension script is run
+- during any callback interaction with UI extension components or APIs
+
+To catch these errors, add an event listener for `unhandledrejection`.
 
   ```javascript
   self.onunhandledrejection = (error) => {
@@ -18,7 +26,9 @@ There are two types of errors when using ui extensions: lifecycle and non-lifecy
   });
   ```
 
-- **Non-lifecycle error:** this error could happen when the script throw exception while running setTimeout callback. To catch this error, you can add event listener for `error`.
+### Non-lifecycle error
+
+These errors can be thrown when the script encounters an exception while running a `setTimeout` callback. To catch these errors, add an event listener for `error`.
 
   ```javascript
   self.onerror = (error) => {
@@ -36,7 +46,7 @@ There are two types of errors when using ui extensions: lifecycle and non-lifecy
 
 See https://docs.bugsnag.com/platforms/javascript/.
 
-You can use either CDN version or npm dependencies to run Bugsnag. However, at runtime, Bugsnag requires additional polyfill as it tries to access browser environment. Below script has been tested at the time of writing this docs
+You can run Bugsnag using the CDN version or npm dependencies. However, Bugsnag requires additional polyfills at runtime, as it will try to access the browser environment. The script below has been tested at the time of writing these docs.
 
 ```javascript
 /**
@@ -79,7 +89,7 @@ By default, Bugsnag has already sent error for `unhandledrejection`. You only ne
 
 See https://docs.sentry.io/platforms/javascript/.
 
-Using Sentry for ui extensions is very straightforward. You only need Sentry main bundle or `@sentry/browser`.
+Using Sentry for UI extensions is straightforward. You only need to import the main Sentry bundle or `@sentry/browser` No polyfills are required.
 
 ```javascript
 // Add Sentry CDN javascript here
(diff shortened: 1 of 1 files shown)


documentation/error-handling.md
```diff
@@ -0,0 +1,103 @@
+# Error handling
+
+Because ui extensions script runs in non-browser environment (ex: WebWorker in Web, JavascriptCore on Mobile), it is recommended to listen and send errors to server manually. However, it is still possible to use external libraries for reporting errors. This document shows a couple of examples about how you can handle it.
```
iainmcampbell, line 3:
```suggestion
Because UI extension scripts run in non-browser environments (Web Worker in web, JavascriptCore on Mobile), it is recommended to manually set up error listeners and error reporting. However, it is possible to use external libraries for error reporting.
```

documentation/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+
+## The generic way of handling errors
```
iainmcampbell, line 5:
```suggestion
## Error handling in UI extensions
```

documentation/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+
+There are two types of errors when using ui extensions: lifecycle and non-lifecycle error.
```
iainmcampbell, line 7:
```suggestion
UI extensions can throw two types of error: lifecycle errors and non-lifecycle errors.
```

documentation/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+
+- **Lifecycle error:** most errors will fall into this type. It could happen when ui extensions script is loaded / imported, the script is run / rendered, or any interaction via callback of ui extensions components or apis. To catch this error, you can add event listener for `unhandledrejection`.
```
iainmcampbell, line 9:
```suggestion
### Lifecycle error

Most errors will fall into this type. These errors can be thrown when:

- a UI extension script is loaded or imported
- the UI extension script is run
- during any callback interaction with UI extension components or APIs

To catch these errors, add an event listener for `unhandledrejection`.
```

documentation/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+  self.onunhandledrejection = (error) => {
+    console.warn('onunhandledrejection', error);
+  };
+  // or
+  self.addEventListener('unhandledrejection', (error) => {
+    console.warn('event unhandledrejection', error);
+  });
+  ```
+
+- **Non-lifecycle error:** this error could happen when the script throw exception while running setTimeout callback. To catch this error, you can add event listener for `error`.
```
iainmcampbell, line 21:
```suggestion
### Non-lifecycle error

These errors can be thrown when the script encounters an exception while running a `setTimeout` callback. To catch these errors, add an event listener for `error`.
```

documentation/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+  });
+  ```
+
+## Handle errors via 3rd-party libraries
+
+### Bugsnag
+
+See https://docs.bugsnag.com/platforms/javascript/.
+
+You can use either CDN version or npm dependencies to run Bugsnag. However, at runtime, Bugsnag requires additional polyfill as it tries to access browser environment. Below script has been tested at the time of writing this docs
```
iainmcampbell, line 39:
```suggestion
You can run Bugsnag using the CDN version or npm dependencies. However, Bugsnag requires additional polyfills at runtime, as it will try to access the browser environment. The script below has been tested at the time of writing these docs.
```

documentation/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Add error handlers
+ */
+self.onerror = (error) => {
+  Bugsnag.notify(error);
+};
+```
+
+By default, Bugsnag has already sent error for `unhandledrejection`. You only need to listen to `onerror`.
```
iainmcampbell, line 76:
```suggestion
By default, Bugsnag will listen for `unhandledrejection`. You only need to manually listen to `onerror`.
```

documentation/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Sentry
+
+See https://docs.sentry.io/platforms/javascript/.
+
+Using Sentry for ui extensions is very straightforward. You only need Sentry main bundle or `@sentry/browser`.
```
iainmcampbell, line 82:
```suggestion
Using Sentry for UI extensions is straightforward. You only need to import the main Sentry bundle or `@sentry/browser` No polyfills are required.
```

documentation/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+self.onerror = (error) => {
+  Sentry.captureException(new Error(error.reason.stack));
+};
+
+self.onunhandledrejection = (error) => {
+  Sentry.captureException(new Error(error.reason.stack));
+};
+```
+
+By default, Sentry does not send any error. You need to add listeners for both `error` and `unhandledRejection`.
```
iainmcampbell, line 103:
```suggestion
By default, Sentry does not listen for any errors. You will need to manually add listeners for both `error` and `unhandledRejection`.
```

iainmcampbell (approved):
great doc 👌 english is a dumb language

documentation/error-handling.md
henrytao-me: This is what I observe. Not sure if there are more cases :D

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
+# Error handling
+
+Because UI extension scripts run in non-browser environments (Web Worker in web, JavascriptCore on Mobile), it is recommended to manually set up error listeners and error reporting. However, it is possible to use external libraries for error reporting.
```
lemonmade, line 3: I'm not sure what this first sentence is telling me I should be doing.

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
...
+  self.onunhandledrejection = (error) => {
+    console.warn('onunhandledrejection', error);
+  };
+  // or
+  self.addEventListener('unhandledrejection', (error) => {
+    console.warn('event unhandledrejection', error);
+  });
+  ```
+
+### Non-lifecycle error
```
lemonmade, line 29: I don't think we should break these down this way. They do technically happen at these times, but that is just a factor of our implementation. For a developer, what they need to know is just that they can add the same global event listeners for untaught exceptions (the error event) and for unhandled promise rejections, just like they can in other JS environments.

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
...
+
+These errors can be thrown when the script encounters an exception while running a `setTimeout` callback. To catch these errors, add an event listener for `error`.
+
+  ```javascript
+  self.onerror = (error) => {
+    console.warn('onerror', error);
+  };
+  // or
+  self.addEventListener('error', (error) => {
```
lemonmade, line 38: Can we pick one of the formats and recommend only that one? I think we should just do the more modern addEventListener() examples.

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
...
+    console.warn('event error', error);
+  });
+  ```
+
+## Handle errors via 3rd-party libraries
```
lemonmade, line 43: This is more about reporting than handling

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Bugsnag
+
+See https://docs.bugsnag.com/platforms/javascript/.
```
lemonmade, line 47: Do I have to go here? Are the instructions in this doc not enough? I'd rather we give them all the common answers they need here, and link them to the docs as an afterthought if they have special cases.

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
...
+
+You can run Bugsnag using the CDN version or npm dependencies. However, Bugsnag requires additional polyfills at runtime, as it will try to access the browser environment. The script below has been tested at the time of writing these docs.
```
lemonmade, line 49: How would they use the CDN version? If they can use both, which do we prefer? Can we show them the commands and code they need for the preferred option?

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
...
+
+```javascript
+/**
+ * Window polyfill
```
lemonmade, line 53: Is there an issue on bugsnag for a version that could work in a worker? If not, could we open one? It's very unfortunate that developers need to do this, since they are changing the global environment of a sandbox they are not supposed to control. Maybe we should just add this in our sandbox instead?

documentation/error-handling.md
lemonmade: Also, what's our recommendation for someone who wants to know they have errors? I think we should probably just recommend the one with the smallest bundle size, though if that's bugsnag, the required polyfill is pretty terrible...

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Add error handlers
+ */
+self.onerror = (error) => {
+  Bugsnag.notify(error);
+};
+```
+
+By default, Bugsnag will listen for `unhandledrejection`. You only need to manually listen to `onerror`.
```
lemonmade, line 86: Why is this, it seems very unexpected to me. The unhandled rejections are newer, I would have thought for sure bugsnag listened for errors automatically.

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Sentry
+
+See https://docs.sentry.io/platforms/javascript/.
+
+Using Sentry for UI extensions is straightforward. You only need to import the main Sentry bundle or `@sentry/browser` No polyfills are required.
+
+```javascript
+// Add Sentry CDN javascript here
+self.Sentry = Sentry;
```
lemonmade, line 96: Does it need to be a global?

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
...
+};
+```
+
+By default, Bugsnag will listen for `unhandledrejection`. You only need to manually listen to `onerror`.
+
+### Sentry
+
+See https://docs.sentry.io/platforms/javascript/.
+
+Using Sentry for UI extensions is straightforward. You only need to import the main Sentry bundle or `@sentry/browser` No polyfills are required.
```
lemonmade, line 92: Is "import the main Sentry bundle" talking about loading from a CDN again? I don't know what it is describing compared to the NPM package.

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
...
+self.onerror = (error) => {
+  Sentry.captureException(new Error(error.reason.stack));
+};
+
+self.onunhandledrejection = (error) => {
+  Sentry.captureException(new Error(error.reason.stack));
+};
+```
+
+By default, Sentry does not listen for any errors. You will need to manually add listeners for both `error` and `unhandledRejection`.
```
lemonmade, line 113: This is equally odd to me as for bugsnag. Do they also not listen by default in a browser context?

documentation/error-handling.md
```diff
@@ -0,0 +1,113 @@
+# Error handling
```
lemonmade, line 1: We should maybe add a quick note/ links for connecting react error boundaries to these services for developers writing react extensions

lemonmade (changes requested):

documentation/error-handling.md
henrytao-me: Are you thinking about using ErrorBoundary for React? I don't think it can catch everything, like importScripts... I don't think there is anything specific to React that we could highlight here.

henrytao-me (commented):
Ready for another review

documentation/error-handling.md
```diff
@@ -0,0 +1,64 @@
...
+
+UI extensions support standard web errors. You can add `unhandledrejection` listener for unhandled promise rejections or `error` listener for other exceptions like Javascript runtime error or a resource fails to load.
```
kumar303, line 3:
```suggestion
UI extensions support standard web errors. You can add an `unhandledrejection` listener for unhandled promise rejections or an `error` listener for other exceptions like JavaScript runtime errors or resources that fail to load.
```

documentation/error-handling.md
```diff
@@ -0,0 +1,64 @@
...
+
+// For other exceptions
+self.addEventListener('error', (error) => {
+  console.warn('event error', error);
+});
+```
+
+## Reporting errors
+
+It's possible to use 3rd party libaries for reporting errors like Bugsnag or Sentry. However, it might require some extra configs or setup depending on which libraries you are using because UI extensions are run inside WebWorker.
```
kumar303, line 19:
```suggestion
It's possible to use 3rd party libaries like Bugsnag or Sentry for reporting errors. However, it might require extra configuration or setup, depending on which libraries you use, because UI extensions run inside [WebWorkers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API).
```

documentation/error-handling.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Bugsnag
+
+Follow [Bugsnag documentation](https://docs.bugsnag.com/platforms/javascript/) to install it to your extension. If you use [CDN version](https://docs.bugsnag.com/platforms/javascript/cdn-guide/), you need to add polyfill code below as Bugsnag tries to access some variables that are not available in WebWorker.
```
kumar303, line 23:
```suggestion
Follow [Bugsnag's documentation](https://docs.bugsnag.com/platforms/javascript/) to install it into your extension. If you use a [CDN version](https://docs.bugsnag.com/platforms/javascript/cdn-guide/), you need to add polyfill code below as Bugsnag tries to access some variables that are not available in WebWorkers.
```

documentation/error-handling.md
```diff
@@ -0,0 +1,64 @@
...
+  window.document = document;
+  window.history = history;
+})();
+```
+
+There is no need to add custom event listeners as Bugsnag does it for you out of the box.
+
+### Sentry
+
+Follow [Sentry documentation](https://docs.sentry.io/platforms/javascript/) to install it to your extension. There is no polyfill required for using Sentry. However, Sentry does not add any event listeners out of the box. So, you need to add them manually.
```
kumar303, line 51:
```suggestion
Follow [Sentry's documentation](https://docs.sentry.io/platforms/javascript/) to install it into your extension. A polyfill is not required for using Sentry. However, Sentry does not add any event listeners out of the box. So, you need to add them manually.
```

documentation/error-handling.md
```diff
@@ -0,0 +1,64 @@
...
+    },
+    addEventListener: () => {},
+  };
+  const history = {};
+  const window = self;
+  self.window = window;
+  window.document = document;
+  window.history = history;
+})();
+```
```
kumar303, line 45: I know Chris mentioned it earlier but, yeah, this is a bug. If you get a chance, please file an issue for Bugsnag and link to it here (if possible) so that someone can see if it's fixed. 

This polyfill code seems fragile since a future version of Bugsnag could access different DOM properties.

kumar303 (approved):
It's exciting to see this come together!

I suggested some grammar tweaks just to make it flow better.

joeyfreund (approved):
LGTM

henrytao-me:
cc @lemonmade any extra thoughts on this PR?

documentation/error-handling.md
lemonmade: I was just thinking of a one-sentence thing, something like "If you are writing your UI extension in React, you can follow Bugsnag’s [React integration guide](https://docs.bugsnag.com/platforms/javascript/legacy/react/) to get additional context on errors thrown while rendering your components."

documentation/error-handling.md
```diff
@@ -0,0 +1,64 @@
+# Error handling
+
+UI extensions support standard web errors. You can add an `unhandledrejection` listener for unhandled promise rejections or an `error` listener for other exceptions like JavaScript runtime errors or resources that fail to load.
```
lemonmade, line 3: I forget, did we actually implement these for the native JS sandbox as well?

documentation/error-handling.md
```diff
@@ -0,0 +1,64 @@
...
+
+// For other exceptions
+self.addEventListener('error', (error) => {
+  console.warn('event error', error);
+});
+```
+
+## Reporting errors
+
+It's possible to use 3rd party libaries like Bugsnag or Sentry for reporting errors. However, it might require extra configuration or setup, depending on which libraries you use, because UI extensions run inside [WebWorkers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API).
```
lemonmade, line 19:
```suggestion
It's possible to use 3rd party libaries like Bugsnag or Sentry for reporting errors. However, it might require extra configuration or setup, depending on which libraries you use, because UI extensions run inside [Web Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API).
```

documentation/error-handling.md
lemonmade: Did we file an issue for this?

lemonmade: Also, do we think it's obvious enough where this code would go in an extension file? We don't show any context of what file/ where in a file we are adding this code, but not sure if we need to...

documentation/error-handling.md
```diff
@@ -0,0 +1,64 @@
...
+  window.document = document;
+  window.history = history;
+})();
+```
+
+There is no need to add custom event listeners as Bugsnag does it for you out of the box.
+
+### Sentry
+
+Follow [Sentry's documentation](https://docs.sentry.io/platforms/javascript/) to install it into your extension. A polyfill is not required for using Sentry. However, Sentry does not add any event listeners out of the box. So, you need to add them manually.
```
lemonmade, line 51: Sentry also has a React-specific library we can reference: https://docs.sentry.io/platforms/javascript/guides/react/

documentation/error-handling.md
lemonmade: Can we add a quick recommendation here? Something like, "If you don’t already have an error reporting tool, but you want to be alerted when errors occur in your extension, we recommend X. It has a free version, and has the smallest bundle size of the error tracking services documented in this guide."

lemonmade (approved):

documentation/error-handling.md
henrytao-me: > did we actually implement these for the native JS sandbox as well?

By `native JS sandbox`, you mean mobile, don't you? If so, no. Current Android gets this for free because it switch to use WebView like Web. I haven't verified it in current iOS as it is using JavascriptCore. 

For future Android and iOS powered by React Native in Mobile Shopify and POS, we get it for free as they will be both using WebView (hidden) and WebWorker like Web. So, I wou

lemonmade: I was talking about mobile, yeah. Thanks for the explanation!

PR #149, opened 2021-06-22
Jwt link

andrewjtech123:
### Background

Adds a link to the JWT spec on shopify.dev and updates soem README as a result of running the script

### Solution

### 🎩

- ...

### Checklist


andrewjtech123:
not sure what happened here 🤔

PR #151, opened 2021-06-22
Update admin dependency

vividviolet:
### Background

`@shopify/ui-extensions` should be a dependency since we are re-exporting types from this package


henrytao-me (approved):
LGTM

PR #152, opened 2021-06-22
Add changesetApplicationsRemaining to post-purchase types

denislins:
### Background

Following the implementation in core (see https://github.com/Shopify/shopify/pull/298195), we're now updating the post-purchase types to include the `changesetApplicationsRemaining` attribute.

You can see more information on the PR liked above, but in summary, it returns how many changesets a partner can still apply to an initial purchase. This is important for payment gateways that only support one changeset per initial purchase, like PayPal express.

### Solution

Just adding the type, along with a brief description.

### Checklist


vividviolet (commented):
@denislins We're in the middle of a rename before publishing the new packages. Do you need this change to be published right away under the old `argo-post-purchase` package? 
If you don't could you base this PR off of https://github.com/Shopify/argo/tree/rename?

vividviolet:
@denislins We just merged the rename branch so I changed your base branch back to `main`

janees-e (approved):
👍

PR #155, opened 2021-06-25
updates for pp

andrewjtech123:
### Background

Makes changes to what content outputs in the docs and some front matter changes.

`ChoiceList` is not a component for post-purchase so I removed the link.

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/scripts/generate-docs-checkout.ts
+++ b/scripts/generate-docs-checkout.ts
@@ -27,8 +27,8 @@ const postPurchase = {
 };
 
 const componentsPageContent = (url: string) => ({
-  title: 'Components for checkout extensions',
-  frontMatterDescription: 'A list of components for checkout extensions.',
+  title: 'Components for Post-purchase extensions',
+  frontMatterDescription: 'A list of components for Post-purchase extensions.',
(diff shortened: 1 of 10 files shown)


scripts/generate-docs-checkout.ts
```diff
@@ -27,8 +27,8 @@ const postPurchase = {
 };
 
 const componentsPageContent = (url: string) => ({
-  title: 'Components for checkout extensions',
-  frontMatterDescription: 'A list of components for checkout extensions.',
+  title: 'Components for Post-purchase extensions',
```
vividviolet, line 30: Is this title not supposed to be generic Checkout UI extensions?

It looks like the post purchase

vividviolet (commented):
It looks like these scripts are for both Post purchase and Checkout but I don't really have enough context. Can you take a look @joeyfreund ?

scripts/generate-docs-checkout.ts
andrewjtech123: For unite we want to scope it only to post-purchase. Afterward, we'll change it back

PR #156, opened 2021-06-25
Consume latest version of remote-ui to support error handling

henrytao-me:
### Background

This PR consumes latest version of remote-ui to support error handling. 

Ref https://github.com/Shopify/checkout-web/issues/5252

### Solution

Consume latest version of remote-ui

### 🎩

- Run and verify it works
```
git clean -xdf . && yarn && yarn type-check && yarn build
```
- Also see this PR for the integration https://github.com/Shopify/web/pull/45842

### Checklist


config/typescript/tsconfig.base.json
```diff
@@ -26,6 +26,7 @@
       "esnext",
       "esnext.asynciterable"
     ],
-    "typeRoots": ["../../node_modules/@types", "./typings"]
+    "typeRoots": ["../../node_modules/@types", "./typings"],
+    "skipLibCheck": true
```
henrytao-me, line 30: This is necessary because sewing-kit will check the node_modules without it 

Screen Shot 2021-06-28 at 11 00 46 AM

olavoasantos (approved):

config/typescript/tsconfig.base.json
vividviolet: Every time this comes up we end up deciding that it's preferable to keep the lib check unless absolutely necessary. Why is this happening again and is there another way to fix the issue? Last time this happened it was due to pulling in multiple versions of React.

henrytao-me: There is an alternative to add `@types/react` to `resolutions` in `package.json` because `@types/react-reconciler` and `@types/react` are different version.

vividviolet: Yes but removing this setting skips checking all node dependencies not just specify one's. Can we dedupe to use the same version of `@types/react`?

henrytao-me: We could deduplicate it. I tried with Olavo

henrytao-me:
Ready for another look

vividviolet:
@henrytao-me I'm reading the TS docs and I think the root of the problem is that remote-ui/react should include `@types/react-reconciler` as a dependency (similar to how `@types/react` is included) since it's a library that bundles its declaration files.

From the example in the docs: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package

> Here, our package depends on the browserify and typescript packages. browserify does not bundle its declaration files with its npm packages, so we needed to depend on @types/browserify for its declarations. typescript, on the other hand, packages its declaration files, so there was no need for any additional dependencies.

> Our package exposes declarations from each of those, so any user of our browserify-typescript-extension package needs to have these dependencies as well. For that reason, we used "dependencies" and not "devDependencies", because otherwise our consumers would have needed to manually install those packages. If we had just written a command line application and not expected our package to be used as a library, we might have used devDependencies.

I think we should update `remote-ui/react` to set `@types/react-reconciler` as a dependency instead of manually adding it for our react packages

packages/post-purchase-ui-extensions-react/package.json
```diff
@@ -21,9 +21,11 @@
...
   },
   "dependencies": {
-    "@remote-ui/react": "^4.0.0",
-    "@shopify/post-purchase-ui-extensions": "^0.10.0",
-    "@types/react": "^17.0.0"
+    "@remote-ui/react": "^4.1.3",
+    "@shopify/post-purchase-ui-extensions": "^0.10.0"
+  },
+  "devDependencies": {
+    "@types/react": "^17.0.11"
```
henrytao-me, line 28: I think this just needs to be a devDependencies. What do you think @lemonmade ?

lemonmade (approved):
👍 Once the resolution is gone

PR #162, opened 2021-06-29
fix: lock down babel to 7.12.13 for @shopify/checkout-ui-extensions-run

vividviolet:
### Background

Fix https://github.com/Shopify/ui-extensions-private/issues/1612

### Solution

Lock down babel dependencies to the version in our yarn.lock

### 🎩

- We will test with an alpha release

### Checklist


henrytao-me (approved):
LGTM

PR #166, opened 2021-07-07
[Docs script] Handle Admin APIs, use new Codeblock tag

iainmcampbell:
### Background

closes https://github.com/Shopify/ui-extensions-private/issues/1614, https://github.com/Shopify/ui-extensions-private/issues/1558

Screen Shot 2021-07-07 at 5 40 33 PM

### Solution

Most of this is a workaround for admin’s nested object types, which complicate the documentation. ie:

```ts
export interface LayoutApi {
  layout: {
    initialValue: Layout;
    setOnChange: (onChange: (layout: Layout) => void) => void;
  };
```

I’ve pulled out the nested value (`layout: {`) into its own interface (`LayoutApiContent`). Then, we skip rendering the first prop table `LayoutApi`, because it just nests the props we actually want to render.

This is mildly hacky, but it doesn't change the public interface and should be future proof, as we will continue to use the same nested pattern for new APIs.

### 🎩

Run the doc generation script: `yarn docs:admin`.

Run `dev up; dev s` in shopify-dev. Check out <https://shopify-dev.myshopify.io/apps/app-extensions/ui-extensions/api>.

You may need to add the feature `post_unite` to Flipper, and fully enable it: <https://shopify-dev.myshopify.io/internal/flipper/features/>.

### Checklist


iainmcampbell:
@tmlayton: the only change to Checkout content is replacing the deprecated `highlight` & `sections` liquid tags with the new `codeblock` liquid tag in the component examples. I can make a PR for this content change.

scripts/generate-docs-admin.ts
```diff
@@ -20,4 +20,8 @@ Each component has general guidelines for usage as well as additional informatio
 - 🖥 denotes desktop specific information
   `,
 });
-extensionPoints(paths);
\ No newline at end of file
+extensionPoints(paths);
+
+adminExtensionApi(paths, {
+  componentsToSkip: ['ContainerApi', 'DataApi']
+});
```
vividviolet, line 27: Missing an empty line at the end of the file

packages/admin-ui-extensions-react/src/extension-api/index.ts
```diff
@@ -1,7 +1,7 @@
 export {useExtensionApi} from './utils';
-export * from './layout';
-export * from './locale';
-export * from './container';
-export * from './sessionToken';
-export * from './toast';
-export * from './data';
+export * from './LayoutApi';
```
vividviolet, line 2: Any particular reasons for renaming the files? We don't usually capitalize unless it's a component

scripts/typedoc/shopify-dev-renderer/api-admin.ts
```diff
@@ -0,0 +1,162 @@
...
+      break;
+    }
+  }
+
+  if(indexToRemove !== -1) {
+    lines.splice(indexToRemove, 1);
+  }
+
+  return lines.join('\n');
+}
```
vividviolet, line 162: Missing an empty line

scripts/typedoc/utilities/dependency-graph.ts
```diff
@@ -771,3 +760,18 @@ function undocumented(node: Node) {
...
+    const localValues = [...value.locals.values()];
+    allInterfaces.push(
+      ...(localValues.filter(
+        ({name, kind}) => filterMethod({name, kind})
+      ) as InterfaceType[]),
+    );
+  })
+
+  return allInterfaces;
+}
```
vividviolet, line 777: Same here, missing an empty line

vividviolet (commented):
The approach makes sense to me - I'll tophat tomorrow

packages/admin-ui-extensions/src/extension-api/LocaleApi/index.ts
```diff
@@ -1,14 +1,16 @@
...
  */
 export interface LocaleApi {
-  locale: {
-    /** IETF-formatted locale at time of page load */
-    initialValue: string;
-
-    /** Callback when user changes locale during runtime */
-    setOnChange(onChange: (locale: string) => void): void;
-  };
+  locale: LocaleApiContent;
```
henrytao-me, line 13: Same comment as LayoutApi

packages/admin-ui-extensions-react/src/extension-api/index.ts
iainmcampbell: the docs parser (`dependency-graph.ts`) assumes files are named after their main/default export

packages/admin-ui-extensions/src/extension-api/LayoutApi/index.ts
```diff
@@ -4,15 +4,17 @@ export interface Layout {
...
 /**
  * You can use the Layout utility to adjust your content based on the user's screen size.
  * Layout returns `'compact'` on mobile devices and at narrow screen widths, and `'regular'` otherwise.
  */
 export interface LayoutApi {
-  layout: {
-    initialValue: Layout;
-    setOnChange: (onChange: (layout: Layout) => void) => void;
-  };
+  layout: LayoutApiContent;
```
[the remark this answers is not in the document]
iainmcampbell: totally! there's a ticket for that already: https://github.com/Shopify/ui-extensions-private/issues/1613

this would involve changing the host implementation and introducing a breaking API change. this PR is just trying to get the docs script to work for APIs.

scripts/generate-docs-admin.ts
iainmcampbell: we had eslint disabled for `scripts/`, i've enabled it which has fixed a whole bunch of formatting issues including this. couldn't enable it for `scripts/argo-admin`, but you're removing those right?

henrytao-me (commented):
I ran `yarn docs:admin` and got this 

```
Can’t resolve export type `Promise` in /Users/henrytao/src/github.com/Shopify/ui-extensions/packages/admin-ui-extensions/src/components/index.ts. Maybe it’s not exported from the component index or imported from a remote package.
```

Also, can you please add more details to tophat instruction? I don't know what today next in order to run this page https://shopify-dev.myshopify.io/apps/app-extensions/ui-extensions/api

iainmcampbell:
@henrytao-me that warning is expected. I added more detailed instructions; forgot about flipper which is probably why that route isn't rendering for you

tmlayton:
> also added success logs so hopefully it's a bit clearer that the warnings the script throws are just noise.

Haha it’s not noise, it’s a feature backlog!

tmlayton:
> @tmlayton: the only change to Checkout content is replacing the deprecated `highlight` & `sections` liquid tags with the new `codeblock` liquid tag in the component examples. I can make a PR for this content change.

Sounds good re: the new liquid tag and thanks in advance for the content change PR

tmlayton (approved):

PR #170, opened 2021-07-13
feature: form component

alanthai:
Ticket: https://github.com/Shopify/ui-extensions-private/issues/1623

### Background

### Solution

Adding a new Form component

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/admin-ui-extensions-react/src/components/Form/index.ts
+++ b/packages/admin-ui-extensions-react/src/components/Form/index.ts
@@ -1,0 +1,2 @@
+export {Form} from './Form';
+export type {FormProps} from '@shopify/admin-ui-extensions';
--- a/packages/admin-ui-extensions/src/components/Button/Button.ts
+++ b/packages/admin-ui-extensions/src/components/Button/Button.ts
@@ -22,6 +22,9 @@ export type ButtonProps = RequiredTitleOrIcon & {
    */
   primary?: boolean;
 
+  /** Allows the button to submit a form */
+  submit?: boolean;
+
   /** Callback when button is pressed */
   onPress?(): void;
 
--- a/packages/admin-ui-extensions/src/components/Form/Form.ts
+++ b/packages/admin-ui-extensions/src/components/Form/Form.ts
@@ -1,0 +1,36 @@
+import {createRemoteComponent} from '@remote-ui/core';
+
+export interface FormProps {
+  /**
+   * Whether the form is able to be submitted. When set to `true`, this will
+   * disable the implicit submit behavior of the form.
+   */
+  disabled?: boolean;
+  /**
+   * Whether this form should be submitted when the merchant presses the "enter"
+   * key on their keyboard. Defaults to `true`. When set to a string value,
+   * the implicit submit behavior will be enabled, and the string will be used
+   * as a label for an explicit submit button that is visible only to assistive
+   * technologies.
+   *
+   * You can also use your own submit button with <Button submit>
+   */
+  implicitSubmit?: boolean | string;
+  /**
+   * A callback that is run when the form is submitted.
+   */
+  onSubmit(): void;
+}
+
+/**
+ * The `Form` component should be used to wrap one or more form controls.
+ * This component provides an "implicit submit" behavior, where merchants can
+ * submit the form from any input by pressing "enter" on their keyboards. This
+ * behavior is widely expected, and should be respected as often as possible.
+ *
+ * Unlike an HTML `form` element, this component does not support configuring
+ * the descendant fields to be submitted via HTTP automatically. Instead, you
+ * must provide an `onSubmit` callback that will perform the necessary HTTP
+ * requests in JavaScript.
+ */
+export const Form = createRemoteComponent<'Form', FormProps>('Form');
--- a/packages/admin-ui-extensions/src/components/Form/examples/Form.example.ts
+++ b/packages/admin-ui-extensions/src/components/Form/examples/Form.example.ts
@@ -1,0 +1,17 @@
+import {extend, ExtensionPoint, TextField, Form} from '@shopify/admin-ui-extensions';
+
+extend('Playground', (root) => {
+  const form = root.createComponent(Form, {
+    onSubmit: () => console.log('submitted!'),
+  }, [
+    root.createComponent(TextField, {
+      label: 'Super text field',
+      type: 'text',
+      onInput: (value) => console.log(value, ' was typed'),
+    }),
+  ]);
+
+  root.appendChild(form);
+
+  root.mount();
+});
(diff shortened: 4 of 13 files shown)


packages/admin-ui-extensions/src/components/Form/examples/Form.example.ts
```diff
@@ -0,0 +1,17 @@
+import {extend, ExtensionPoint, TextField, Form} from '@shopify/admin-ui-extensions';
+
+extend('Playground', (root) => {
+  const form = root.createComponent(Form, {
+    onSubmit: () => console.log('submitted!'),
+  }, [
+    root.createComponent(TextField, {
+      label: 'Super text field',
+      type: 'text',
+      onInput: (value) => console.log(value, ' was typing'),
```
olavoasantos, line 10: Little typo here: `(value) => console.log(value, ' was typed')`

packages/admin-ui-extensions/src/components/Form/examples/Form.example.ts
```diff
@@ -0,0 +1,17 @@
+import {extend, ExtensionPoint, TextField, Form} from '@shopify/admin-ui-extensions';
+
+extend('Playground', (root) => {
+  const form = root.createComponent(Form, {
+    onSubmit: () => console.log('submitted!'),
+  }, [
+    root.createComponent(TextField, {
```
iainmcampbell, line 7: we need to assign this component to a variable and append it to the form component, right?

packages/admin-ui-extensions/src/components/Form/Form.ts
```diff
@@ -0,0 +1,34 @@
...
+   */
+  disabled?: boolean;
+  /**
+   * Whether this form should be submitted when the buyer presses the "enter"
+   * key on their keyboard. Defaults to `true`. When set to a string value,
+   * the implicit submit behavior will be enabled, and the string will be used
+   * as a label for an explicit submit button that is visible only to assistive
+   * technologies.
+   */
+  implicitSubmit?: boolean | string;
```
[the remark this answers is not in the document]
alanthai: @iainmcampbell  I think `disabled` and `implicitSubmit` are non-overlapping in concern. `disable` corresponds to disabling of the form. That could mean if you have a custom `Button.submit` and press it, but it won't call `onSubmit`. You could also have `implicitSubmit=true` and `disabled=true` for a time when the form can't submit.

I should mention the `Button.submit` though.

@olavoasantos I think you could split it if you want, but I prefer checkout's implementation because it's concise. You'd probably need something like `implicitSubmitLabel` instead of something generic like `label` since it's not a form label, just a button label. But you'd never use it if you have `implicitSubmit=false`, so to keep things in a sane state, I like just keeping it as is.

Open to discuss both if you disagree!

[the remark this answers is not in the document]
iainmcampbell: ah! I understand now. since `disabled` is indeed different, I do agree that the concise `implicitSubmit` is nice.

packages/admin-ui-extensions/src/components/Form/Form.ts
```diff
@@ -0,0 +1,36 @@
+import {createRemoteComponent} from '@remote-ui/core';
+
+export interface FormProps {
+  /**
+   * Whether the form is able to be submitted. When set to `true`, this will
+   * disable the implicit submit behavior of the form.
```
iainmcampbell, line 6:
```suggestion
   * Whether the form is able to be submitted. When set to `true`,
   * also disable the implicit submit behaviour.
```

packages/admin-ui-extensions/src/components/Form/examples/Form.example.ts
iainmcampbell: oh! gotcha, I didn't realize remote-ui could do that. I still think it would be good to assign the component to a variable for readability, but that's just nitpicking now

packages/admin-ui-extensions/src/components/Form/Form.ts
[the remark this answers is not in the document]
olavoasantos: Yea, I understand. I do think it would be more intuitive in terms of API to split but I don't think it's a big deal if you prefer to leave as is.

packages/admin-ui-extensions/src/components/Form/Form.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * The `Form` component should be used to wrap one or more form controls.
+ * This component provides an "implicit submit" behavior, where merchants can
+ * submit the form from any input by pressing "enter" on their keyboards. This
+ * behavior is widely expected, and should be respected as often as possible.
+ *
+ * Unlike an HTML `form` element, this component does not support configuring
+ * the descendant fields to be submitted via HTTP automatically. Instead, you
+ * must provide an `onSubmit` callback that will perform the necessary HTTP
+ * requests in JavaScript.
```
elanalynn, line 34: Great description!

elanalynn (approved):

packages/admin-ui-extensions-react/src/components/Form/index.ts
```diff
@@ -0,0 +1,2 @@
+export {Form} from './Form';
+export type {FormProps} from '@shopify/admin-ui-extensions';
```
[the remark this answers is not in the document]
alanthai: Ah OK. Should we create a ticket to do it for all react components, maybe?

packages/admin-ui-extensions/src/components/Button/Button.ts
```diff
@@ -22,6 +22,9 @@ export type ButtonProps = RequiredTitleOrIcon & {
    */
   primary?: boolean;
 
+  /** Allows the button to submit a form */
+  submit?: boolean;
```
vividviolet, line 26: Sorry, I totally missed this PR. How will this prop actually work? Is it supposed to find the parent Form's `onSubmit` and trigger it? I'm not sure how feasible this would be on mobile. cc @james-woo @gutivg

james-woo: It is not feasible for the button to find the parent form.

vividviolet: @alanthai can we remove this prop then? I don't think it's necessary especially since the form has its own `onSubmit`. If developers need to make a button also submit the form they can just call the same `onSubmit` handler for `onPress` instead of us automating this.

alanthai: @vividviolet @james-woo We don't automate this. This is an HTML feature and it's handled by the browser by default. What's the convention on form submit buttons on mobile?

Issue #132, opened 2021-06-10
[Docs script] Parse types imported from outside a project


iainmcampbell:
ie https://github.com/Shopify/argo/pull/131/files#diff-d9f7c7d2fb6a1b6c7fe5317ad99f1a16380074c311023c3edf920ba60667b575

PR #172, opened 2021-07-13
[Docs script] Resolve imported types

iainmcampbell:
### Background

Closes https://github.com/Shopify/ui-extensions/issues/132

:warning: based on https://github.com/Shopify/ui-extensions/pull/166

We’re starting to store types in a shared package (`ui-extensions`) and extend them to produce our public interfaces. The doc generation script isn’t capable of following those imports, therefore the generated docs are missing information.

[`Banner`](https://github.com/Shopify/ui-extensions/blob/main/packages/admin-ui-extensions/src/components/Banner/Banner.ts#L12) is the only component consuming a remotely imported type at the time of this PR.

[`PostPurchaseShouldRenderApi`](https://github.com/Shopify/ui-extensions/blob/38e3adacc084152591126acf89e1094db94be568/packages/post-purchase-ui-extensions/src/extension-points/api/post-purchase/post-purchase.ts#L5-L11) and [`PostPurchaseRenderApi`](https://github.com/Shopify/ui-extensions/blob/38e3adacc084152591126acf89e1094db94be568/packages/post-purchase-ui-extensions/src/extension-points/api/post-purchase/post-purchase.ts#L20-L37) extend an imported type from the same package, and are also missing properties in the generated docs.

### Solution

The dependency graph now processes imports from sibling libraries in this monorepo, flags remote `extends` (instead of ignoring them) and does a final pass over the entire graph to resolve remote extends.

<table>
<tr><th colspan="2"><code>BannerProps extends BaseBannerProps</code></th></tr>
<tr>
<th>before</th>
<th>after</th>
</tr>
<tr>
<td>Screen Shot 2021-07-13 at 5 41 02 PM</td>
<td>Screen Shot 2021-07-13 at 5 41 56 PM</td>
</tr>
</table>

### 🎩

On branch `admin-api-doc-extraction`, run `yarn docs:admin`. Commit the changes in a temporary branch in shopify-dev repo.

Check out this branch, then run `yarn docs:admin` again. Note the changes.

Repeat the same process, but run `yarn docs:checkout`.

### Checklist

### DIFF
--- a/packages/ui-extensions/src/types/index.ts
+++ b/packages/ui-extensions/src/types/index.ts
@@ -1,5 +1,5 @@
-export * from './Banner';
-export * from './BlockStack';
-export * from './Button';
-export * from './Heading';
-export * from './InlineStack';
+export type {BaseBannerProps} from './Banner';
+export type {BaseBlockStackProps} from './BlockStack';
+export type {BaseButtonProps} from './Button';
+export type {BaseHeadingProps} from './Heading';
+export type {BaseInlineStackProps} from './InlineStack';
--- a/packages/ui-extensions/src/index.ts
+++ b/packages/ui-extensions/src/index.ts
@@ -1,1 +1,1 @@
-export * from './types';
+export type {BaseBannerProps, BaseButtonProps, BaseHeadingProps} from './types';
--- a/scripts/typedoc/types.ts
+++ b/scripts/typedoc/types.ts
@@ -39,6 +39,7 @@ export interface ImportedReference {
 export interface InterfaceType extends Documentable {
   kind: 'InterfaceType';
   name: string;
+  remoteExtends?: string;
   properties: PropertySignature[];
 }
 
--- a/scripts/typedoc/utilities/dependency-graph.ts
+++ b/scripts/typedoc/utilities/dependency-graph.ts
@@ -128,16 +128,11 @@ export async function createDependencyGraph(entry: string) {
     }
   });
 
-  return modules;
+  return resolveRemoteExtends(modules);
 }
 
 async function extractModule(file: string): Promise<Module> {
   const source = readFileSync(file, 'utf8');
-
-  // const parser = new TSDocParser();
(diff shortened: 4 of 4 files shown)


packages/ui-extensions/src/types/index.ts
```diff
@@ -1,3 +1,3 @@
-export * from './Banner';
-export * from './Button';
-export * from './Heading';
+export type {BaseBannerProps} from './Banner';
+export type {BaseButtonProps} from './Button';
+export type {BaseHeadingProps} from './Heading';
```
iainmcampbell, line 3: gotta do the same thing here that we had to do in `admin-ui-extensions`: all exported interfaces need to use both `export type` and be named (no `*`). the algorithm we're using to parse imports isn't sophisticated enough to handle `*`.

iainmcampbell: cc @elanalynn

iainmcampbell: i wonder if I can set up a linting rule to disallow `export *`...

alanthai: Sure can:

*eslintrc.js*
```
module.exports = {
  rules: {
    'no-restricted-syntax': [
      {
        selector: 'ExportAllDeclaration',
        message: 'Error message goes here',
      },
    ],
  },
},
```

If you want it to only target certain filepaths you can add it into [overrides](https://eslint.org/docs/user-guide/configuring/configuration-files#how-do-overrides-work) instead

alanthai (approved):
Looks good! 🎉 

Screen Shot 2021-07-14 at 10 06 56 AM

henrytao-me (approved):
🎩-ed. LGTM

packages/ui-extensions/src/types/index.ts
iainmcampbell: faf

tmlayton (approved):

PR #179, opened 2021-07-26
Extract title variable

Igor10k:
### Background

The docs generation for Post-purchase extensions docs and `beta` Checkout extensions docs are using the same template, so there's no way to show different heading depending on the docs subject.

### Solution

Use a `title` variable instead of the hardcoded title and pass it as an argument to the docs generator function.


andrewjtech123 (approved):
LGTM ! Thank you

PR #180, opened 2021-07-26
chore: fix admin version bump

alanthai:
The previous ignore changes is not picking up any `admin` folders.

Also removing the global version bump

### Background

### Solution

### 🎩

- ...

### Checklist


package.json
```diff
@@ -11,9 +11,8 @@
     "lint": "sk lint",
     "type-check": "sk type-check",
     "build": "sk build",
-    "version-bump": "lerna version --no-push --include-merged-tags",
     "version-bump:checkout": "lerna version --no-push --ignore-changes 'packages/{admin-ui-extensions*,ui-extensions}/**' --include-merged-tags",
-    "version-bump:admin": "lerna version --no-push --ignore-changes 'packages/!({admin*,ui-extensions})/**' --include-merged-tags",
+    "version-bump:admin": "lerna version --no-push --ignore-changes 'packages/!(admin*)/**' --include-merged-tags",
```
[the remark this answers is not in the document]
alanthai: I was thinking after discussion that maybe we want to be more deliberate when version bumping but I think it's not fully fleshed out. I'm going to re-add it (but fixed) until we get a better grasp of the solution.

package.json
```diff
@@ -11,9 +11,8 @@
     "lint": "sk lint",
     "type-check": "sk type-check",
     "build": "sk build",
-    "version-bump": "lerna version --no-push --include-merged-tags",
     "version-bump:checkout": "lerna version --no-push --ignore-changes 'packages/{admin-ui-extensions*,ui-extensions}/**' --include-merged-tags",
-    "version-bump:admin": "lerna version --no-push --ignore-changes 'packages/!({admin*,ui-extensions})/**' --include-merged-tags",
+    "version-bump:admin": "lerna version --no-push --ignore-changes 'packages/!(admin*|ui-extensions)/**' --include-merged-tags",
```
henrytao-me, line 15: What does `|` do?

alanthai: admin* OR ui-extensions. I was just experimenting, and `!(a|b)` seems to work, but the old `!({a,b})` does not.

alanthai: the other ones are working with `{}`. I just noticed when I tried version bumping that the combo of `!` and `({})` weren't working for me.

henrytao-me (approved):
🎩-ed. LGTM

PR #181, opened 2021-07-27
ResourcePicker API

elanalynn:
### Background

Closes Shopify/ui-extensions-private#1448

This PR introduces a UI Extensions `ResourcePicker` API.

### Solution

Introduce new APIs for `ResourcePicker`:

- `ResourcePickerProps` 
-  `Item` (`ResourcePickerProps` requires an array of Items)
- `FilterControl` (different from `FilterControl` in `ResourceList` because `onQueryChange` and `onQueryClear` are optional)
- `AvatarProps` and `ThumbnailProps` to make media type in `Item` more explicit

Screen Shot 2021-08-27 at 2 08 47 PM

### 🎩

See 🎩  instructions [here](https://github.com/Shopify/web/pull/47118)

### Checklist

### DIFF
--- a/packages/admin-ui-extensions/src/components/ResourcePicker/ResourcePicker.ts
+++ b/packages/admin-ui-extensions/src/components/ResourcePicker/ResourcePicker.ts
@@ -1,0 +1,145 @@
+import {createRemoteComponent} from '@remote-ui/core';
+import {DestructableAction} from '../types';
+
+type Alignment = 'leading' | 'trailing' | 'center';
+
+interface AvatarProps {
+  kind: 'Avatar';
+  accessibilityLabel: string;
+  name: string;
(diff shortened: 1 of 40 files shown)


packages/admin-ui-extensions/src/components/ResourcePicker/ResourcePicker.ts
```diff
@@ -0,0 +1,145 @@
...
+  source: string;
+}
+
+interface ThumbnailProps {
+  kind: 'Thumbnail';
+  source: string;
+  alt: string;
+}
+
+interface FilterControl {
```
alanthai, line 20: This feels slightly opinionated but I guess we could always make this a React.Element or similar later if they need more flexibility.

PR #187, opened 2021-07-28
feat: add children prop

alanthai:
### Background

Added children prop to all components that take children in `admin-ui-extensions-react`.

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/admin-ui-extensions-react/src/components/Banner/index.ts
+++ b/packages/admin-ui-extensions-react/src/components/Banner/index.ts
@@ -1,2 +1,5 @@
+import type {PropsWithChildren} from 'react';
+import type {BannerProps as BaseBannerProps} from '@shopify/admin-ui-extensions';
+
+export type BannerProps = PropsWithChildren<BaseBannerProps>;
 export {Banner} from './Banner';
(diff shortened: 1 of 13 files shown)


packages/admin-ui-extensions-react/src/components/Button/index.ts
```diff
@@ -1,2 +1,5 @@
+import type {ReactPropsFromRemoteComponentType} from '@remote-ui/react';
+import type {ButtonProps as BaseButtonProps} from '@shopify/admin-ui-extensions';
+
+export type ButtonProps = ReactPropsFromRemoteComponentType<BaseButtonProps>;
```
henrytao-me, line 4: Admin UI Extension Button doesn't have children. So, `BaseButtonProps` is enough.

henrytao-me (changes requested):
Missing Banner

packages/admin-ui-extensions-react/src/components/Button/index.ts
alanthai: Oops thanks

PR #189, opened 2021-07-28
chore: update build-consumer and restore-consumer for spin to work with short form workspace name

vividviolet:
### Background

Improve build-consumer and restore-consumer scripts for Spin so you don't have to type out the full workspace names

### Solution

### 🎩

- Create a spin instance for Web
- Run `yarn build-consumer-spin web admin-ui-extensions`
- Open the Web spin instance and inspect the node modules to ensure the package content was copied over

### Checklist

### DIFF
--- a/scripts/build-consumer-spin.sh
+++ b/scripts/build-consumer-spin.sh
@@ -7,7 +7,7 @@ Usage: $(basename "$0") <workspace short name> [<package_name>]
 
 Build UI Extensions packages into a project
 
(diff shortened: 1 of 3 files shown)


henrytao-me (commented):
🎩-ed. I think there is some issues with `,`

Screen Shot 2021-07-29 at 3 03 04 PM

henrytao-me (commented):
Found the issue here https://github.com/Shopify/ui-extensions/blob/main/scripts/consumer-helper.sh#L3. 

They should not have `,`

henrytao-me (approved):
:shipit:

vividviolet:
Good catch! Fixed!

PR #190, opened 2021-07-28
Remove `hidden` property from post-purchase components' docs

igor10k:
### Background

The sidebar in the post-purchase extensions components docs doesn't work because the components docs page have a `hidden: true` property.

### Solution

Edit the docs generating script not add `hidden: true`.


andrewjtech123:
We don't need to have `post_unite: show` anymore either.

andrewjtech123 (approved):
LGTM

PR #206, opened 2021-08-23
Render examples using new format

iainmcampbell:
https://github.com/Shopify/shopify-dev/pull/10060 changes the format of the code block tag, so we need to change the output of the generator script.

I also added a comment to the YAML frontmatter stating that these files are autogenerated. I didn't include any other metadata otherwise the comment would change with every export, creating unnecessary diff noise. I think we can get the metadata we need from git logs.

### test

Run `yarn:docs-admin` and `yarn:docs-checkout`.

Run `git status` and `git diff` in shopify-dev to see what's changed. Only the format of the code block `{% code %}` tags should change, as well as the added frontmatter comment.


scripts/generate-docs-admin.ts
```diff
@@ -23,7 +23,8 @@ Each component has general guidelines for usage as well as additional informatio
   `,
   },
   {
-    compileExamples: true,
+    compileExamples: false,
+    visibility: 'visible',
```
iainmcampbell, line 27: we're not ready to use `compileExamples` in shopify-dev yet; got a bit ahead of myself here

paulomarg (approved):
Haven't 🎩 ed again, but LGTM!

JaKXz (approved):
We should merge this since shopify/shopify-dev#10060 has been merged

PR #236, opened 2021-10-19
Export ApplyChangesetOptions from PostPurchase

mawburn:
### Background

Needed to complete https://github.com/Shopify/checkout-web/issues/6626 and send the ApplyChangesetOptions to the api from the post purchase extension. 

Edit: 

~~I also noticed that ESLint didn't seem to be added to this repo for some reason, so I added that. For that I am adding lemonmade to this PR. This feels weird to me, especially since it's failing CI builds because of it.~~

Edit 2:

I realized after digging through the errors on this it was a problem with my original install. I thought it was pretty weird. 

### Solution

The type was already there & exported, so I just exported it the way everything else is.

### Checklist

### DIFF
--- a/packages/post-purchase-ui-extensions-react/package.json
+++ b/packages/post-purchase-ui-extensions-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/post-purchase-ui-extensions-react",
-  "version": "0.13.1",
+  "version": "0.13.0",
   "description": "React bindings for @shopify/post-purchase-ui-extensions",
   "publishConfig": {
     "access": "public",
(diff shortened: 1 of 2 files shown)


alxclark (approved):
Looks good 👍🏼  There's probably a lot more types that we should be exposing for post-purchase, but lets asses this later.

For the package changes, I would prefer if we kept the version to `0.13.0` for this PR and follow up with another PR to update the version and push the tags.

There's more info here on the release process: https://github.com/Shopify/ui-extensions-private/wiki/Release-steps-for-Admin-and-Checkout-UI-extensions#library-release-process

mawburn:
> I would prefer if we kept the version to 0.13.0 for this PR

If we did that, then we can't pull it in without wiping out the cache. This could be a big headache if someone didn't know they needed to do that.

PR #251, opened 2021-12-09
Add line in build-consumer to echo spin instance

elanalynn:
### Background

This is a tiny thing, but I feel more confident seeing where `build-consumer-spin` is copying the build.

### Solution

I added a line to echo the spin instance.

echo spin

### 🎩

Run `yarn build-consumer-spin web` and confirm that the output is your current spin instance.

### Checklist

### DIFF
--- a/scripts/consumer-helper.sh
+++ b/scripts/consumer-helper.sh
@@ -52,6 +52,7 @@ function copy_to_target {
     targetDir=$(resolve "$ROOT/../$projectDirectoryOrWorkspace/node_modules/@shopify/$package")
   else
     targetDir="$shopifyNodeModulesDir/$package"
+    echo "🌀 spin instance: `spin show | grep Shopify/$projectName | awk '{print $1}'` "
   fi
 
   if (run_command "[ -d $targetDir ]"); then


scripts/consumer-helper.sh
```diff
@@ -52,6 +52,7 @@ function copy_to_target {
     targetDir=$(resolve "$ROOT/../$projectDirectoryOrWorkspace/node_modules/@shopify/$package")
   else
     targetDir="$shopifyNodeModulesDir/$package"
+    echo "🌀 spin instance: `spin show | grep Shopify/$projectName | awk '{print $1}'` "
```
vividviolet, line 55: Instead of adding a new message here, can we add to line 27 instead? This is where it's saying whether we're running spin or not

vividviolet (commented):
Looks good - I have a minor suggestion of moving the message for simplicity

elanalynn:
@vividviolet I moved the spin echo up, and since we are using the url twice, I simplified and made it a variable. Also, since I'm now testing this on my M1, I had to update the node version.

vividviolet (approved):
Great job 👍

PR #273, opened 2022-03-01
Reduce noise when running ui-extension in Shopify's Spin

kumar303:
### Background

When running `ui-extensions` in Shopify's internal Spin development tool, the repo is marked as a "failed service"

### Solution

Adding an empty `Procfile` is the suggested workaround until this is resolved: https://github.com/Shopify/spin/issues/3928

### 🎩

- ...

### Checklist

### DIFF
--- a/.spin/Procfile
+++ b/.spin/Procfile
@@ -1,0 +1,3 @@
+# This is an empty Procfile so that ui-extensions does not appear as a failed service in Spin.
+# See https://github.com/Shopify/spin/issues/3928
+# ui-extensions is a library, not a service.


.spin/Procfile
```diff
@@ -0,0 +1,3 @@
+# This is an empty Procfile so that ui-extensions does not appear as a failed service in Spin.
+# See https://github.com/Shopify/spin/issues/3928
```
vividviolet, line 2: Should we remove this link since it cannot be accessed publicly?

vividviolet (approved):

.spin/Procfile
kumar303: I was mainly thinking that someone reading the code later could look at that issue and undo the workaround 🤷

PR #275, opened 2022-03-09
Added retail-ui-extensions package

js-goupil:
### Background

We're starting to move our UI components from our private repo to this one. This PR is to start setting up the project, however we don't want to publish anything just yet.

### Solution

I've copied the layout for admin-ui-extensions as the way we've set it up in our private repo has been heavily inspired by this.

### DIFF
--- a/packages/retail-ui-extensions/package.json
+++ b/packages/retail-ui-extensions/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "@shopify/retail-ui-extensions",
+  "version": "0.0.1",
+  "main": "index.js",
+  "module": "index.mjs",
+  "esnext": "index.esnext",
+  "types": "index.d.ts",
+  "exports": {
+    ".": {
+      "esnext": "./index.esnext",
+      "import": "./index.mjs",
+      "require": "./index.js"
+    },
+    "./": "./"
+  },
+  "license": "MIT",
+  "sideEffects": false,
+  "dependencies": {
+    "@remote-ui/core": "^2.1.3"
+  },
+  "publishConfig": {
+    "access": "public",
+    "@shopify:registry": "https://registry.npmjs.org/"
+  }
+}
--- a/packages/retail-ui-extensions/sewing-kit.config.ts
+++ b/packages/retail-ui-extensions/sewing-kit.config.ts
@@ -1,0 +1,7 @@
+import {createPackage} from '@sewing-kit/config';
+import {uiExtensionsPackage} from '../../config/sewing-kit';
+
+export default createPackage((pkg) => {
+  pkg.entry({root: './src/index'});
+  pkg.use(uiExtensionsPackage());
+});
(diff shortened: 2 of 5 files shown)


packages/retail-ui-extensions/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "./": "./"
+  },
+  "license": "MIT",
+  "sideEffects": false,
+  "dependencies": {
+    "@remote-ui/core": "^2.1.3",
+    "@shopify/ui-extensions": "^1.0.1"
+  },
+  "publishConfig": {
+    "access": "public",
```
js-goupil, line 23: Should we be making this private for now?

alanthai (approved):
Don't see anything wrong here. I'm not sure about sewing-kit set up, though. @vividviolet any thoughts?

packages/retail-ui-extensions/sewing-kit.config.ts
```diff
@@ -0,0 +1,9 @@
+import {createPackage} from '@sewing-kit/config';
+import {uiExtensionsPackage} from '../../config/sewing-kit';
+
+export default createPackage((pkg) => {
+  pkg.entry({root: './src/index'});
+  pkg.entry({name: 'extension-points', root: './src/extension-points'});
```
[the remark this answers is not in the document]
lemonmade: If you actually have these, the expectation would be that you also have additional entries in `exports` that map to these. IMO it is not necessary, this feels like another early optimization (we haven't made it in checkout despite the fact that we have a lot more API exposed).

packages/retail-ui-extensions/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../config/typescript/tsconfig.admin.json",
+  "compilerOptions": {
+    "rootDir": "src",
+    "outDir": "build/ts"
+  },
+  "references": [{"path": "../ui-extensions"}],
```
lemonmade, line 7: Same question as above — will you actually depend on this?

packages/retail-ui-extensions/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../config/typescript/tsconfig.admin.json",
```
lemonmade, line 2: @vividviolet @alanthai why does the admin have its own overrides for this? They seem unnecessary since the TS outputs are not actually for consumers, they are purely used as types

js-goupil: @lemonmade I ended up changing this to extend `"extends": "../../config/typescript/tsconfig.base.json"`

packages/retail-ui-extensions/sewing-kit.config.ts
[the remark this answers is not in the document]
js-goupil: Thanks for the feedback. I excluded these for now as there's no need to prematurely optimize.

packages/retail-ui-extensions/tsconfig.json
vividviolet: @lemonmade good catch! This is legacy and should be deleted. We used to use the typescript compiler to build our bundles that's why we had our own settings. @alanthai do you have time to delete this stuff since you're ATC this week?

PR #277, opened 2022-03-11
Moving the POS APIs into retail-ui-extensions

js-goupil:
### Background

This is the second PR in a series of PRs to move our vanilla JS retail-ui-extensions into this repo. This brings in the API interfaces and models, so that the partner's app can call various functions on the cart, receive cart updates, as well as locale updates and be able to fetch a session token.

These APIs were tested with an experimental branch of POS and suite our needs currently, however they are likely to change as we move into betas and such.

### Solution

This PR only brings in the vanilla Javascript APIs. Once we're done migrating all of our code from our private repo, we will work on the React wrapper around these.

### 🎩

- No way to tophat yet!

### Checklist


packages/retail-ui-extensions/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './extension-api';
```
lemonmade, line 1: How are the details here used? It's odd to me that we do not introduce the notion of "extension points" before introducing APIs extensions have, since in all other extension libraries the extension APIs are just a property of the extension points (https://github.com/Shopify/ui-extensions/blob/main/packages/checkout-ui-extensions/src/extension-points/extension-points.ts#L7-L50 for checkout)

packages/retail-ui-extensions/src/extension-api/CartApi/CartApiContent.ts
```diff
@@ -0,0 +1,102 @@
+import type {Address, Cart, Customer, CustomSale} from '../Models/Cart';
```
lemonmade, line 1: Just as a nitpicky note: I do not think we should use pascal case names unless we are talking about components. In the end it doesn't matter to consumers, but my original intention was for pascal case to signal a similar function as it does in our React (web) codebases, where it almost always means "component" (with everything else using kebab case, or sometimes camel case)

packages/retail-ui-extensions/src/extension-api/CartApi/CartApiContent.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export type DiscountType = 'Percentage' | 'FixedAmount' | 'Code';
+
+export interface CartApiContent {
+  /** Initial value of the Cart when the screen is rendered. */
+  initialValue: Cart;
+
+  /** Callback when the cart changes */
+  setOnChange(onChange: (cart: Cart) => void): void;
```
lemonmade, line 10: Could you please consider using `@remote-ui/async-subscription` (https://github.com/Shopify/remote-ui/tree/main/packages/async-subscription) for any changeable value you are sharing? I designed this based on what I knew we would need for checkout, where everything is mutable, and it handles edge cases you likely are not (handling subscriptions in a way that doesn't ever lose updates due to async timing issues, gracefully managing the retain and release of subscriptions, etc)

packages/retail-ui-extensions/src/extension-api/CartApi/CartApiContent.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  /** Adds custom properties to the cart
+   * @param properties the custom key to value object to attribute to the cart
+   */
+  addCartProperties(properties: Record<string, string>): void;
+
+  /** Removes the specified cart properties
+   * @param keys the collection of keys to be removed from the cart properties
+   */
+  removeCartProperties(keys: string[]): void;
```
lemonmade, line 57: These will all technically return promises at runtime, so I'd recommend making that clear in their return types.

packages/retail-ui-extensions/src/extension-api/CartApi/CartApiContent.ts
```diff
@@ -0,0 +1,102 @@
...
+  removeLineItemProperties(index: number, keys: string[]): void;
+
+  /** Add a discount on a line item to the cart
+   * @param index the index of the line item that should receive a discount
+   * @param type the type of discount applied (example: 'Percentage')
+   * @param title the title attributed with the discount
+   * @param amount the percentage or fixed monetary amount deducted with the discout
+   */
+  setLineItemDiscount(
+    index: number,
```
lemonmade, line 81: The final API is up to you folks, but I would encourage you to avoid the very positional argument-heavy structure you have for these APIs. They are very fragile as adding properties after the fact is very difficult. Most other APIs we design (for UI extensions and just in general) rely much more on object arguments, where adding properties is easy.

packages/retail-ui-extensions/src/extension-api/Models/Cart.ts
```diff
@@ -0,0 +1,55 @@
...
+  firstName?: string;
+  lastName?: string;
+  phone?: string;
+  province?: string;
+  country?: string;
+  zip?: string;
+  name?: string;
+  provinceCode?: string;
+  countryCode?: string;
+}
```
lemonmade, line 55: It would be great if we could try to get closer to the checkout shape for an address (https://github.com/Shopify/ui-extensions/blob/main/packages/checkout-ui-extensions/src/extension-points/api/standard/index.ts#L451-L463); a developer may never write both kinds of extensions but we are very clearly shipping the org chart with these subtly different shapes.

packages/retail-ui-extensions/src/extension-api/NavigationApi/index.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+/**
+ * Access the navigation API to present your app in a full screen modal.
+ */
+export interface NavigationApi {
+  navigation: NavigationApiContent;
+}
+
+export function isNavigationApi(api: any): api is NavigationApi {
```
lemonmade, line 16: Will not all extensions have access to these APIs? It's hard to tell what the value of these "does this API have this functionality" functions is without knowing the actual API available to POS extensions.

packages/retail-ui-extensions/src/extension-api/CartApi/CartApiContent.ts
js-goupil: Good feedback. So far most of these functions are based on what partners currently use for App Bridge, but we are definitely going to be trying to make that experience better at the same time.

PR #278, opened 2022-03-16
Added retail UI components

js-goupil:
### Background

This PR brings in all of our vanilla JS components from our private repo. We still need to add documentation for the props, but we can nail that in a follow up PR. 

### Solution

These components are written to work flawlessly with retail's design system. However, we have made a selection of props that we think are going to be impactful, but this may definitely change once we get feedback from our beta.

### 🎩

- ...

### Checklist


packages/retail-ui-extensions/src/components/Dialog/index.ts
```diff
@@ -0,0 +1 @@
+export {Dialog, DialogType} from './Dialog';
```
samiulmonir06, line 1: Also, do we need to export `PropTypes`? similar to `StackNavigator`?

samiulmonir06 (approved):
Looks good, Left questions regarding exporting `types` and `props`. Other than that, overall looks good.

PR #279, opened 2022-03-21
Added extension points for retail-ui-extensions

js-goupil:
### Background

This PR adds the extension points for the `retail-ui-extensions` library. TL;DR we have two extension points within POS:

1. Smart grid tile (the tile on the grid that can be dynamic depending on the cart state.
2. Smart grid modal (the full screen experience that can be presented by tapping on the tile).

### Solution

I followed the approach that `admin-ui-extensions` had for the product subscription extension points.

### Checklist


js-goupil:
@lemonmade I noticed that the checkout extension's extension points approach is quite different. Our original approach matched the admin-ui-extensions, let me know if there are benefits to doing it the checkout way.

henrytao-me (approved):
LGTM

packages/retail-ui-extensions/src/extension-points/identifiers/smart-grid.ts
```diff
@@ -0,0 +1,20 @@
+import {StandardApi} from '../types';
+
+export type ExtensionPoint =
+  | 'Retail::Pos::SmartGrid::Tile'
+  | 'Retail::Pos::SmartGrid::Modal';
+
+export type SmartGridStandardApi<T extends ExtensionPoint> = StandardApi<T>;
+
+export type SmartGridTileApi = SmartGridStandardApi<
+  'Retail::Pos::SmartGrid::Tile'
```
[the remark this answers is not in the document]
js-goupil: Pos does look ugly! I'll fix these names. I'm also going to try out the checkout way of doing things (especially if it's less code)

js-goupil:
@lemonmade I see what you mean about how admin pre optimized for different extension points. POS is likely to be even more different across extension points but I see that in your file that you linked to me, they're all referring to StandardAPI, but that API could in theory be anything right?

lemonmade:
Yeah, right now checkout only has one extension point with a different API, https://github.com/Shopify/ui-extensions/blob/main/packages/checkout-ui-extensions/src/extension-points/extension-points.ts#L30-L33. I still think checkout's structure works very nicely for supporting this, that API is all on its own in one file with no surrounding noise, and then included in the mapping.

In general, I would watch out for the API diverging between extension points. Many things can be expressed declaratively in components, instead of through an imperative API. At this point, if you need different data provided to different extension points, that's likely where most of the divergences would come from. If we move to [”input queries”](https://github.com/Shopify/ui-extensions-private/discussions/1678), where extensions list the data they need with GraphQL APIs, extension points likely won't even diverge based on the data they are provided (this is a model we are exploring for checkout extensions).

js-goupil:
@lemonmade That looks and sounds like it should meet our requirements. I've pushed a couple commits changing our approach to follow what checkout does. I really like the declaration of the extension points. The rest was just a copy/paste (ie the `types.ts`,  `render-extension.ts`. I think a lot of those exports such as `ApiForRenderExtension` are required to create the react hooks in the react wrapper package

packages/retail-ui-extensions/src/extension-points/extension-points.ts
```diff
@@ -0,0 +1,21 @@
...
+
+type Tile = typeof import('../components/Tile');
+type TileComponent = Tile[keyof Tile];
+
+export interface ExtensionPoints {
+  'Retail::SmartGrid::Tile': RenderExtension<
+    StandardApi<'Retail::SmartGrid::Tile'>,
+    TileComponent
+  >;
+  'Retail::SmartGrid::Modal': RenderExtension<
```
lemonmade, line 15: Out of curiosity, is the intention to have extensions that target one of these two extension points, or do you expect most/ all extensions will target both? If this is to create a "click on tile opens up modal"-type of workflow, I'd probably recommend making the modal be a component you render from the tile extension point, instead of a dedicated extension point. But I'm just guessing at what you're trying to achieve with this (a brief comment in code about the use cases for these extension points is probably a good idea).

lemonmade (approved):
Thank you for going back and restructuring this! I am obviously very biased, but I feel the code is a lot more readable, and it's a lot more clear what kind of types are shareable between the libraries when you see the duplication here and in checkout extensions.

@vividviolet @henrytao-me how do you feel about these different package structures?

packages/retail-ui-extensions/src/extension-points/extension-points.ts
js-goupil: Great question! The tile could open the full screen modal, or it could execute another action. I think the modal does need to be its own extension point as there is no guarantee it will only be presented modally (it could be pushed onto the stack in cart actions as well). In all cases, on the client side the "Modal" is a full on react native screen that needs to be navigated to, so the best way to handle this is by having it be its own extension point.

lemonmade: There doesn't need to be any direct connection between extension points and views in your application. The extension can live entirely outside the routing hierarchy, and you should be able to interpret the components rendered by the extension into the appropriate React component/ routing structure in the host application. Separating extensions across multiple extension points introduces complexity for the developer because there is no natural mechanism for sharing state between them (other than global state, which is generally smelly in JavaScript).

A single extension point lets extensions do a lot of powerful things, like share state, share React context, preload data for views that aren't needed just yet, etc. If you need extensions to be able to target each extension point individually, it makes sense to keep them separate, but if they are part of a single UI "flow", I strongly recommend keeping them as one extension point and using components to express the views getting pushed on the stack.

vividviolet:
> Thank you for going back and restructuring this! I am obviously very biased, but I feel the code is a lot more readable, and it's a lot more clear what kind of types are shareable between the libraries when you see the duplication here and in checkout extensions.
> 
> @vividviolet @henrytao-me how do you feel about these different package structures?

@lemonmade I'm totally open to simplifying Admin's types as long as we're able to infer the API interface based on extension point and make it easy to bootstrap new extension points. The current Admin flow works with the new-extension-point generator script and make it so that new extensions points are available for the `useData` and `useContainer` hooks. Ex. you can do `useData<"MyExtensionPoint">` and `useContainer<"MyExtensionPoint">`. 

We definitely assumed all Admin extension will be significantly different but I'm unsure what the future holds for Admin extensions. Certainly Product Subscription is an outlier as it's very specific and I don't think we'll see another similar extension.

lemonmade:
The pattern I am pushing for definitely support type inference (checkout uses the types from the ui-extension library in its app code, to make sure it is providing the correct API and component types for a given extension point). It is is a little harder to bootstrap, though, since there is a shared type that needs to be edited. Do we need to be so concerned with generators, though? We have 4 extension points in the admin after almost two years; at this point, we have made more UI extension libraries than we have made admin UI extension points.

PR #283, opened 2022-04-06
Add missing page to Shopify.dev

rennyG:
### Background

A broken anchor linking to missing content causes build failures on `shopify-dev` when generating the `beta/checkout-extension` docs. The problem is twofold:

- Failing build check on `shopify-dev` means the link has to be manually removed, to ship updates for either post-purchase or beta checkout extensions content

- Partners are missing information that they need

### Solution

Add `globals.md` to the array

### 🎩
globals

### Checklist

- [X ] I have :tophat:'d these changes
- [X ] I have updated relevant documentation

### DIFF
--- a/scripts/typedoc/shopify-dev-renderer/getting-started.ts
+++ b/scripts/typedoc/shopify-dev-renderer/getting-started.ts
@@ -43,7 +43,7 @@ export function gettingStarted(paths: Paths, options: Options = {}) {
 
   const docsInputPath = resolve(`${paths.inputRoot}/documentation`);
 
-  const files = ['extension-points.md'];
+  const files = ['extension-points.md', 'globals.md'];
 
   if (fs.existsSync(docsInputPath)) {
     files.forEach((file) => {


scripts/typedoc/shopify-dev-renderer/getting-started.ts
```diff
@@ -43,7 +43,7 @@ export function gettingStarted(paths: Paths, options: Options = {}) {
 
   const docsInputPath = resolve(`${paths.inputRoot}/documentation`);
 
-  const files = ['extension-points.md'];
+  const files = ['extension-points.md', 'globals.md'];
```
ecbrodie, line 46: I wonder if we could support globbing somehow? Like, support adding all Markdown files via a string pattern like `'*.md'`. But, out of scope of this PR.

ecbrodie (approved):
I trust that we'll verify correctness of this when we generate new docs.

scripts/typedoc/shopify-dev-renderer/getting-started.ts
rennyG: Yeah would be a cleaner one

rennyG:
@ecbrodie @lemonmade we're actually going to need this change in for `checkout-web`. It's important to include the page in the hidden docs, for the hackathon. We can discuss options. @jffortier for visibility

ecbrodie:
I already approved. Up to you if you think you need a second 👍

jamesvidler (approved):
I've tophatted this change and had to apply this patch manually to fix missing documentation when checkout docs are regenerated.

jamesvidler:
Closing this PR since this was addressed in another [PR](https://github.com/Shopify/ui-extensions/pull/291) which has shipped.

PR #285, opened 2022-04-08
Js add react retail

js-goupil:
### Background

### Solution

### 🎩

- ...

### Checklist


packages/retail-ui-extensions-react/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "exports": {
+    ".": {
+      "esnext": "./index.esnext",
+      "import": "./index.mjs",
+      "require": "./index.js"
+    },
+    "./": "./"
+  },
+  "dependencies": {
+    "@remote-ui/react": "^4.3.0",
```
lemonmade, line 24: I think you can do `^4.5.0` here

packages/retail-ui-extensions/src/extend.ts
```diff
@@ -0,0 +1,10 @@
+import type {ShopifyGlobal} from './globals';
+/**
```
lemonmade, line 2: Nit: missing newline above this line

packages/retail-ui-extensions/src/globals.ts
```diff
@@ -0,0 +1,15 @@
...
+  extend<ExtensionPoint extends keyof ExtensionPoints>(
+    extensionPoint: ExtensionPoint,
+    extend: ExtensionPoints[ExtensionPoint],
+  ): void;
+  reload(): void;
+}
+
+declare global {
+  interface WorkerGlobalScope {
+    readonly le_shopify: ShopifyGlobal;
```
lemonmade, line 13: I think you want `shopify` here, checkout had to do this for <reasons> that you probably don't need to worry about

packages/retail-ui-extensions/src/components/index.ts
```diff
@@ -8,3 +8,4 @@ export * from './SegmentedControl';
 export * from './Tag';
 export * from './Dialog';
 export * from './SearchBar';
+export * from './Types';
```
lemonmade, line 11: Nit, and no reason to change right now: we typically would not use a pascal-case file name unless it contains an actual component

lemonmade (approved):

henrytao-me (approved):
No extra comments 🙇

PR #286, opened 2022-04-11
Refactored and renamed retail extension files

js-goupil:
### Background

This PR refactors file contents and names to be more consistent, as well as follow the conventions that other ui extensions are following. Primarily, any file that contains interfaces or types, are renamed to use kebab case. I've also made it more consistent that index files primarily export types, instead of defining them.

### Solution

This PR is mainly just renaming and reorganization. There aren't any functionality changes.

### Checklist


henrytao-me (approved):

packages/retail-ui-extensions/src/extension-api/index.ts
```diff
@@ -1,26 +1,19 @@
-import type {CartApi} from './CartApi';
-import type {LocaleApi} from './LocaleApi';
-import type {NavigationApi} from './NavigationApi';
-import type {SessionTokenApi} from './SessionTokenApi';
+export type {LocaleApi} from './locale-api';
+export type {CartApiContent, DiscountType} from './cart-api/cart-api';
+export type {CartApi} from './cart-api';
+export type {NavigationApi, NavigationApiContent} from './navigation-api';
+export type {StandardApi} from './standard-api';
```
samiulmonir06, line 5: Curious to know, If we are exporting `LocaleApi`(and others) through `standard-api`, why we do need to export it separately again?

samiulmonir06 (approved):
Nice work, it looks cleaner than before. Left a comment.

PR #288, opened 2022-04-12
Update prettier to 2.6.2 and fix formatting

vividviolet:
### Background
Update to @remote-ui/core@^2.1.9, @remote-ui/react@^4.5.0 and @remote-ui/async-subscription@^2.1.9

### Solution

Update our prettier to latest so it understands TS template literals. Also bump our shared `@remote-ui` dependencies so all packages are consistent and we can clean up the yarn.lock.

### 🎩

I'm not sure we need to top hat as all of these bumps are minor/patch versions

### Checklist

### DIFF
--- a/packages/admin-ui-extensions-react/package.json
+++ b/packages/admin-ui-extensions-react/package.json
@@ -16,7 +16,7 @@
   "license": "MIT",
   "sideEffects": false,
   "dependencies": {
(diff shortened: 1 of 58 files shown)


jamesvidler (commented):
Can this PR be closed now?

PR #289, opened 2022-04-20
Fixed dependency things

js-goupil:
### Background

We're having compiler issues when running extensions on a third party app. Specifically with the `checkout-ui-extension-run` library. Hopefully this fixes some issues. Either way the tsconfig.json being ignored is causing complains in the retail-ui-extensions-react package.

### Solution

Delete the `npmignore` file, as well as fix the tsconfig.

### Checklist

### DIFF
--- a/packages/retail-ui-extensions-react/package.json
+++ b/packages/retail-ui-extensions-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/retail-ui-extensions-react",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "description": "React bindings for @shopify/retail-ui-extensions",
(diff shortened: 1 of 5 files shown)


packages/retail-ui-extensions-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/retail-ui-extensions-react",
-  "version": "0.1.0",
+  "version": "0.2.0",
```
samiulmonir06, line 3: why do we need to version dump? I thought we only need to do that when we publish our library.

samiulmonir06 (approved):
Looks good, left a comment

PR #291, opened 2022-04-25
Seperate checkout post purchase docs

jamesvidler:
### Background

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "deploy": "lerna publish from-package --yes",
     "docs:admin": "babel-node --extensions .ts,.tsx,.mjs,.js,.json scripts/generate-docs-admin.ts",
     "docs:checkout": "babel-node --extensions .ts,.tsx,.mjs,.js,.json scripts/generate-docs-checkout.ts",
-    "docs:post-purchase": "babel-node --extensions .ts,.tsx,.mjs,.js,.json scripts/generate-docs-checkout.ts",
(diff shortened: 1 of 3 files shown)


jamesvidler:
> I like this. I feel like, at this point, splitting these two out will make troubleshooting each a bit easier. That could be my Ruby use a class bias though  🙈

Ya it also allows us to get that `globals.md` fix in. We just won't update/generate the post-purchase docs until we are ready for that.

rennyG:
Tophat here, possibly an error on my part? https://github.com/Shopify/shopify-dev/pull/19392/files

ecbrodie (approved):
I'm trusting James and Ren on the tophat.

Change looks alright to me. Thoughts on introducing a `docs:all` script to generate docs for all three extension types at once?

jamesvidler:
@rennyG now that we shipped the fix that resolves the doc gen critical error, you can now test this properly by generating the docs seperately `yarn docs:checkout` vs `yarn docs:post-purchase`.

rennyG (approved):
Approving! Some issues on the docs end (see linked PR), but discussed with @jamesvidler and the issue is outside of the changes introduced in this PR. 

Running the `yarn docs:checkout` command regenerates references for checkout only, not post-purchase, so approving this PR, and the issue at source that's causing odd docs will be handled separately.

PR #299, opened 2022-05-02
Removed gitignore and reorganized package.json

js-goupil:
### Background

This PR removes a `gitignore` file that is not needed, and reorganizes the package.json (adds a description) so that it matches the other packages in this repo.

### Solution

We're having some build issues, so we're trying to replicate the other packages as closely as possible.

### Checklist


samiulmonir06 (approved):
Looks Good

PR #301, opened 2022-05-04
Added retail to tsconfig

js-goupil:
### Background

We're having some issues installing retail-ui-extensions and retail-ui-extensions-react on test projects. It looks like it's related to compiling issues. Maybe adding our packages to `tsconfig` might solve some problems (either way they should be added here).

UPDATE: Adding our packages to the tsconfig actually made CI fail due to type check errors. This means it's corrected some build issues. I've corrected the type check errors, and also made all the exports explicit so that it's consistent across the entire package.

### Checklist


lemonmade (approved):
LGTM 👍 I wish TypeScript could just figure this out automatically :/

PR #303, opened 2022-05-05
Swapped self for globalThis

js-goupil:
### Background

We're having issues working with self on POS. Our yarn type-check is complaining with the usage of self. Instead I'm swapping it for the widely adopted globalThis. This seems to resolve the complains.

### Solution

Swapping to `globalThis`. Is this correct? It seems weird to have to do this, but the type definition of self is `GlobalWorkScope & typeof globalThis`. 

### Checklist


samiulmonir06 (approved):
Looks good

PR #310, opened 2022-05-13
fixed missing props with omit and pick

jamesvidler:
### Background

This PR updates some documentation files fixed by a patch added https://github.com/Shopify/docs-tools/pull/19. It simply includes updates to some docs as well as using the new version of `docs-tools`.

To learn more and 🎩 tophatting instructions: https://github.com/Shopify/shopify-dev/pull/20193


rennyG (approved):
approving, will trust your tophat

PR #311, opened 2022-05-13
Separate text field components

vctrchu:
### Background

To give partners access to all the types of text fields that are available natively, we must separate the types which are a normal text field and a formatted one. Previously all the textfields were packed into one component but that has been deprecated and changed since. 

### Solution

I have created two separate components for the two types of text fields.

### 🎩

I haven't manually copied the files over to my pos and hackday projects yet since there is a lot of files that would require changing. 
Questions:
- Do you think its worth copying or as long as the code looks good just release another version and go from there? 
- Are there any other places I have to update when creating new components on the ui-extensions?

### Checklist


packages/retail-ui-extensions/src/components/FormattedTextField/FormattedTextField.ts
```diff
@@ -0,0 +1,34 @@
...
+  | 'date'
+  | 'zip'
+  | 'password'
+  | 'email'
+  | 'url'
+  | 'phone';
+
+export interface FormattedTextFieldProps {
+  inputType?: InputType;
+  title?: string;
```
js-goupil, line 21: I do see a lot of attributes here that are shared with `TextField.ts`. We could probably stick them into an interface called `BaseTextFieldProps`, and `FormattedTextFieldProps` and the other one can extend the base props. We could put the base props in the shared directory of components

packages/retail-ui-extensions/src/components/TextField/TextField.ts
```diff
@@ -1,31 +1,42 @@
...
-  | 'currency'
-  | 'giftcard'
-  | 'expiry'
-  | 'date'
-  | 'zip'
-  | 'password'
-  | 'email'
-  | 'url'
-  | 'phone';
+interface ActionProps {
```
vctrchu, line 3: I'm not sure but do I need to export these..?

packages/retail-ui-extensions/src/components/FormattedTextField/FormattedTextField.ts
```diff
@@ -1,7 +1,8 @@
 import {createRemoteComponent} from '@remote-ui/core';
+import {TextFieldProps} from 'components/TextField';
 import type {AutoCapitalizationType} from '../shared';
 
-export type InputType =
+type InputType =
```
js-goupil: I think we should keep this exported, partners might want to have access to this type.

packages/retail-ui-extensions/src/components/TextField/TextField.ts
js-goupil: I think we should export these, in case partners want to do logic outside of the component's render. Make sure you add the export at all the index files

js-goupil (commented):
Starting to come together! I still think we should create a base prop to share the props that both TextField and FormattedTextField have in common 😛

samiulmonir06 (commented):
Looks Good, Left some questions.

samiulmonir06 (approved):
Looks Good. 👍

PR #322, opened 2022-05-20
brought over public packages, re-generated docs

jamesvidler:
### Background

### Solution

### 🎩

- ...

### Checklist


jamesvidler:
Ran into an issue tophatting this:
Screen Shot 2022-05-20 at 4 27 02 PM


```
"Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem."
```

Will dig into this further on Tuesday.

packages/checkout-ui-extensions/src/components/View/View.ts
```diff
@@ -149,6 +254,41 @@ export interface ViewProps extends Bordered {
...
+  /**
+   * A label that describes the purpose or contents of the element. When provided,
+   * it will be announced to buyers using assistive technologies and will
+   * provide them with more context.
+   */
+  accessibilityLabel?: string;
+  /**
+   * Set the semantic of the component’s content
+   */
+  accessibilityRole?:
```
lemonmade, line 273: @jffortier do you know what the tuple version of this type means?

yarn.lock
```diff
@@ -10116,6 +10148,15 @@ react-is@^17.0.1:
   resolved "https://registry.yarnpkg.com/react-is/-/react-is-17.0.1.tgz#5b3531bd76a645a4c9fb6e693ed36419e3301339"
   integrity sha512-NAnt2iGDXohE5LI7uBnLnqvLQMtzhkiAOLXTmv+qnF9Ky7xAPcX8Up/xWIhxvLVGJvuLiNc4xQLtuqDRzb4fSA==
 
+"react-reconciler@>=0.26.0 <0.27.0", react-reconciler@^0.26.0:
```
lemonmade, line 10151: A bunch of these can technically be deduplicated, I'd probably run`npx yarn-deduplicate` with the `--packages` flag to avoid these new duplicates.

lemonmade (approved):

kumar303:
> The actual error message wasn't logged in the console, but this is the error here before it was swallowed up

@marvinhagemeister filed an issue for showing the actual error: https://github.com/Shopify/ui-extensions/issues/325

kumar303 (approved):
I didn't tophat but this looks good to me

packages/checkout-ui-extensions/src/components/View/View.ts
jffortier: Yes!

- when one value is set, we render the correct HTML element if one exist for that specific role. ex: `<View accessibilityRole="header">` will render `<header>`
- when a second value is set, we then set the `role` property on the element. ex: `<View accessibilityRole={['header', 'banner']}>` will render `<header role="banner">`

It can be useful for many use cases.  In Self Serve, they need the list/listItem semantic _and_ the stylistic flexibility of View for their UI. They also needed a list item that had the role of "separator".

```tsx
<View accessibilityRole="list" {...styleOnlyPossibleWithView}>
   <View accessibilityRole="listItem">...</View>
   <View accessibilityRole={{'listItem', 'separator']} border="base" />
   <View accessibilityRole="listItem">...</View>
</View>
```

Which renders
```html
<ul>
   <li>...</li>
   <li role="separator" class="css-class-for-border" />
   <li>...</li>
</ul>
```

lemonmade: Hmm, interesting. I think we should probably add some of this content as documentation for the prop, unless a deeper explainer is somewhere else. It wasn't clear to me what they meant, and I still find it a bit hard to understand from this type what values are actually useful to use in the two-value format (e.g., `['listItem', 'listItem']` is valid here but presumably not a thing you'd ever want in practice?)

jffortier: > I think we should probably add some of this content as documentation for the prop, unless a deeper explainer is somewhere else. 

Yeah, I remember providing that feedback in the PR, but now can't find the detailed explanation in the code. Not sure what happened. I'll add one.

>  I still find it a bit hard to understand from this type what values are actually useful to use in the two-value format (e.g., ['listItem', 'listItem'] is valid here but presumably not a thing you'd ever want in practice?)

We can provide examples in that description. Yes, `['listItem', 'listItem']` is valid with this API. We currently allow it (eg:   this will render `<li role="listItem">`, but could decide to not do anything with the second value when its identical to the first.

marvinhagemeister:
> > The actual error message wasn't logged in the console, but this is the error here before it was swallowed up
> 
> @marvinhagemeister filed an issue for showing the actual error: #325

Good catch! Looks like there is a deeper issue where error messages get swallowed completely

PR #329, opened 2022-05-27
Publish 0.15.1

jamesvidler:
- @shopify/checkout-ui-extensions-react@0.15.1
- @shopify/checkout-ui-extensions@0.15.1
### Background

This is a package bump PR prior to publishing public packages. For details on what has changed in this new version see this PR https://github.com/Shopify/ui-extensions/pull/322.

### Solution

Followed the [publishing packages instructions](https://checkout.docs.shopify.io/docs/referencedocs/web/extensions/libraries).

### 🎩
 Instructions can be found in the original PR that brought over the changes: https://github.com/Shopify/ui-extensions/pull/322

### Checklist

### DIFF
--- a/packages/checkout-ui-extensions-react/package.json
+++ b/packages/checkout-ui-extensions-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/checkout-ui-extensions-react",
-  "version": "0.15.0",
+  "version": "0.15.1",
   "description": "React bindings for @shopify/checkout-ui-extensions",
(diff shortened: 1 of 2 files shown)


kumar303 (approved):
Should it be `0.16.0` since we [added new features](https://semver.org/)? Either way, looks good 🚀

lemonmade:
Agreed that we should avoid going to 0.16 until, ideally, the beta, just to minimize the template updating we need to do.

PR #333, opened 2022-05-30
Checkout UI ext react hooks docs

jamesvidler:
### Background

This patch along with https://github.com/Shopify/docs-tools/pull/21 will solve https://github.com/Shopify/docs-tools/issues/20.

### Solution

- Use some sort of identifier such as "Hook" to filter functions declarations that are React Hooks
- Loop through each React Hook and output a table row - one row per React Hook which shows its `name`, `function signature/type` and `description`. 
- The `function signature/type` should show the parameters for the function (and their types) along with the return type
- If the parameter or return is a mapped type it should link to the type via #{typeName} so a user can navigate to it within the document

### 🎩

- ...

### Checklist

### DIFF
--- a/scripts/typedoc/shopify-dev-renderer/extension-points.ts
+++ b/scripts/typedoc/shopify-dev-renderer/extension-points.ts
@@ -2,7 +2,7 @@ import {resolve} from 'path';
 import * as fs from 'fs';
 import {createDependencyGraph, filterGraph} from '@shopify/docs-tools';
 
-import type {Paths, InterfaceType} from '../types';
+import type {Paths, InterfaceType, PropertySignature} from '../types';
 
 import {
   renderYamlFrontMatter,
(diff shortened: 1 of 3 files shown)


scripts/typedoc/shopify-dev-renderer/extension-points.ts
```diff
@@ -88,3 +143,35 @@ export async function extensionPoints(paths: Paths, options: Options = {}) {
...
+
+  // Sort alphabetically (tsdoc seems to get this confused)
+  hooks.sort((aa: any, bb: any) => {
+    if (aa.value.name > bb.value.name) {
+      return 1;
+    } else if (aa.value.name < bb.value.name) {
+      return -1;
+    } else {
+      return 0;
+    }
```
marvinhagemeister, line 174: Might be a cool place to use [String.prototype.localeCompare()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare)

```suggestion
    return aa.value.name.localeCompare(bb.value.name)
```

PR #346, opened 2022-06-07
Add extend and render functions for customer account

mathieulag:
### Background

Fixes https://github.com/Shopify/core-issues/issues/39368. Added `extend` method for vanilla JS implementation, and `render` method for React implementation of Customer Account.

### Solution

Most of the code was taken from the Checkout implementation, although we only kept what we actually need for now. We'll likely need a few of the things Checkout exposes in the extension API, but our plan is to expose as little as possible until we need it.

#### Shared component types

We've added a dependency on `checkout-ui-extensions` to be able to re-use the components that we'll share between the projects. We're already sharing the implementation, but this way we're also sharing the types we expose.

We're only exposing the `Button` for now as a first step, we'll add the other one we need as we go on.

At some point, it would make sense to extract these into a shared component type library, and this way is the simplest to get us there in the meantime.

### 🎩

N/A

### Checklist


robin-drexler (approved):

yarn.lock
```diff
@@ -2223,6 +2231,18 @@
     "@types/react-reconciler" "^0.26.0"
     react-reconciler ">=0.26.0 <0.27.0"
 
+"@remote-ui/react@^4.5.2":
```
lemonmade, line 2234: Could we dedupe `@remote-ui/{core,react}`? I don't think the duplicate versions are necessary in this case.

lemonmade (approved):
I was flipping back and forth on whether to ask for fully decoupling the checkout components vs sharing them as you have in this PR. Fully sharing them seems kind of dicey because it introduces coordination cost for us on changing the API definitions and on changing the implementations. That said, I think it is a nice forcing function to keep the developer's experience as the top priority...

I'm good to roll with sharing the component definitions, but want to make sure at least @danrosenthal or @jffortier are aware of, and OK with, that arrangement too.

yarn.lock
robin-drexler: Good catch. We should probably add a check to CI before we have too many duplicates.

packages/customer-account-ui-extensions/src/extension-points/extension-points.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * A UI extension will register for one or more extension points using `shopify.extend()`.
+ * An extension point in a UI extension is a plain JavaScript function.
+ * This function receives some API for interacting with the application,
+ * and is expected to return a value in a specific shape.
+ * The input arguments and the output type are different
+ * for each extension point.
+ */
+export interface ExtensionPoints {
+  'CustomerAccount::FullPage::RenderWithin': RenderExtension<
```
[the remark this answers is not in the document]
mathieulag: We went with that after talking to UX from the Checkout Extensibility team. They shared this [spreadsheet](https://docs.google.com/spreadsheets/d/1nErQrLhJ3CF4YcKXQi1r08qsOgqoPFpLzonnnf7SxRA/edit#gid=1145280301) with us, and we used the nomenclature from it to match.

You're correct it matches the Post Purchase extension, within the spreadsheet, the extension point is actually `Checkout::PostPurchase::RenderWithin`.

mathieulag:
Sounds good, I will book us some time to discuss this!

PR #349, opened 2022-06-09
Checkout Docs: handle templateLiteral types

jffortier:
### Background

Fix https://github.com/Shopify/checkout-web/issues/11043

Template literals are currently marked as [undocumented in docs-tools](https://github.com/Shopify/docs-tools/blob/main/src/dependency-graph.ts#L301-L303). When this PR ships that resolve the type and creates the object to digest here, we'll have them render correctly in the readme.

- Changes to types here will also be done at the source in checkout-web

### Solution

I had to change the use of `Unit<>` for simplicity and instead use the template literal directly. I don't think we were getting much value from using `Unit` anyway.

### 🎩
- In the docs-tools repo, checkout PR: 
- In the docs-tools repo, run `yarn link`
- In the ui-extensions repo, run `yarn link @shopify/docs-tools`
- In the ui-extensions repo, run `yarn docs:checkout` (preview the readme files in an online markdown viewer. I found the VSCode version is problematic and renders things differently)

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "rollup": "^2.52.2",
     "rollup-plugin-jsx": "^1.0.3",
     "uuid": "^8.3.2",
-    "@shopify/docs-tools": "^0.0.14"
+    "@shopify/docs-tools": "^0.0.15"
   },
   "resolutions": {
     "@typescript-eslint/eslint-plugin": "^4.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2548,10 +2548,10 @@
   resolved "https://registry.yarnpkg.com/@sewing-kit/webpack-plugin-hash-output/-/webpack-plugin-hash-output-0.0.2.tgz#a41ddef000619ac20e64513edbbfc6f28afe7905"
   integrity sha512-BF3gmC9YZf4VY5RLAc5qUbxft16O614jnTtGr8gPRbJF4y3Qz3AVSxSB8dJDldzk4AH9lTQvH0e1Wkbgp5tR8g==
 
-"@shopify/docs-tools@^0.0.14":
-  version "0.0.14"
-  resolved "https://registry.yarnpkg.com/@shopify/docs-tools/-/docs-tools-0.0.14.tgz#fa77788eb2e3a40d525c5242b379de0d6b0aa1dc"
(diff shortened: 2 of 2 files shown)


packages/checkout-ui-extensions/src/components/Grid/Grid.ts
```diff
@@ -2,25 +2,29 @@ import {createRemoteComponent} from '@remote-ui/core';
...
   Spacing,
   InlineAlignment,
   BlockAlignment,
 } from '../shared';
 
-export type GridItemSize = 'auto' | 'fill' | number | Unit<'fr' | '%'>;
+/* eslint-disable eslint-comments/no-unlimited-disable */
+// eslint-disable-next-line
+export type GridItemSize = 'auto' | 'fill' | number | `${number}fr` | `${number}%`;
+/* eslint-enable eslint-comments/no-unlimited-disable */
```
igor10k, line 13: Why's eslint disabled here?

scripts/typedoc/shopify-dev-renderer/shared/index.ts
```diff
@@ -313,6 +313,32 @@ function propType(
       return `${value.value}`;
     case 'BooleanLiteralType':
       return `${value.value}`;
+    case 'TemplateLiteralType':
+      // eslint-disable-next-line no-case-declarations
```
igor10k, line 317: Can we wrap the case clause into `{}` instead of disabling the rule?

scripts/typedoc/shopify-dev-renderer/shared/index.ts
```diff
@@ -313,6 +313,32 @@ function propType(
...
+          if (first.start > second.start) {
+            return 1;
+          } else if (second.start > first.start) {
+            return -1;
+          }
+          return 0;
+        },
+      );
+
+      return `&#96${sortedValues
```
igor10k, line 329: How about creating a const for `&#96` to improve readability? I think ideally the case should end with something like
```
return `${BACKTICK}${result}${BACKTICK}`;
```

igor10k (approved):
I didn't dig into the details like what's inside `node.value.cooked` and `node.value.raw` so I'll trust you on those. And left some nits. Aside from that looks good.

packages/checkout-ui-extensions/src/components/Grid/Grid.ts
jffortier: There's a mismatch of version of Prettier between checkout-web and ui-extensions. In ui-extensions, its a version that throw errors for template literals. This is out of scope of this PR and since it affects more than our package, I believe it would be preferable to do this separately.

PR #355, opened 2022-06-10
Don't unpack repeating types in place

igor10k:
### Background

When the same type is used more than once in a union it can get really messy.

Screen Shot 2022-06-10 at 1 37 51 PM

### Solution

When a type is used more than once in a union don't unpack it in place. Instead render its name as link and scroll to its description on click.

Screen Shot 2022-06-10 at 1 38 07 PM

### Checklist


jffortier (approved):
That works well. As a next step, do you think we'll be able to fetch the comments associated to each values in the union and render them in the table?

PR #360, opened 2022-06-14
Render union type values' comments

igor10k:
### Background

We have some useful comments for some union type values that are currently not rendered. For example,

```ts
export type AccessibilityRole =
  /** Used to indicate the primary content. */
  | 'main'
  /** Used to indicate the component is a header. */
  | 'header'
  /** Used to display information such as copyright information, navigation links, and privacy statements. */
  | 'footer'
  /** Used to indicate a generic section. */
  | 'section'
  /** Used to designate a supporting section that relates to the main content. */
  | 'complementary'
  ...
```

### Solution

Render them as a table.

For example,

Screen Shot 2022-06-14 at 12 47 35 PM

The corresponding `docs-tools` PR to make this possible: https://github.com/Shopify/docs-tools/pull/29

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@sewing-kit/plugin-package-flexible-outputs": "^0.1.0",
     "@sewing-kit/plugin-react": "^0.1.0",
     "@sewing-kit/plugin-typescript": "^0.1.0",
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
This is a great idea! The table looks much better 👍

PR #361, opened 2022-06-14
customer account components

robin-drexler:
### Background

Adds (selected) checkout components + one exclusive Self Serve components (`PageBlock`) to the `customer-account-ui-extensions` and `customer-account-ui-extensions-react` package.

We'll be adding documentation etc. later. We'll also have a chat soon on how to (better) organize components that are shared between checkout and customer account. 

We also added some missing type exports to the checkout react package: https://github.com/Shopify/ui-extensions/pull/361/files#diff-caeca760aafbe6b1565e10f0fe4dd403ac60588a909237fafe0568ca722cedf6R40

### 🎩

- n/a

### Checklist


packages/checkout-ui-extensions-react/src/index.ts
```diff
@@ -37,6 +37,9 @@ export type {
   UnitSuffix,
   Spacing,
   Size,
+  Status,
```
robin-drexler, line 40: Those were exported in the non-react version, but are missing here.

packages/customer-account-ui-extensions/src/components/PageBlock/PageBlock.ts
```diff
@@ -0,0 +1,30 @@
...
+   * - `base` means blockStart, inlineEnd, blockEnd and inlineStart paddings are `base`
+   *
+   * - [`base`, `none`] means blockStart and blockEnd paddings are `base`, inlineStart and inlineEnd paddings are `none`
+   *
+   * - [`base`, `none`, `loose`, `tight`] means blockStart padding is `base`, inlineEnd padding is `none`, blockEnd padding is `loose` and  blockStart padding is `tight`
+   */
+  padding?: CSSShorthand<Spacing> | Responsive<CSSShorthand<Spacing>>;
+}
+
+export const PageBlock = createRemoteComponent<'PageBlock', PageBlockProps>(
```
jffortier, line 28: I understand its a component specific to Self Serve, but do you think 3rd party developers will need this in their extensions? Could this be wrapping extension points instead, so that developers choosing theses extension point are guaranteed to be consistent with the rest of the page?

Just curious to understand. Not a blocker.

packages/customer-account-ui-extensions/src/components/shared-components.ts
```diff
@@ -1,2 +1,92 @@
...
+  Breakpoint,
+  CSSShorthand,
+  Responsive,
+  Status,
+  Size,
+  Spacing,
+  Alignment,
+  Appearance,
+  Unit,
+  UnitSuffix,
```
jffortier, line 91: ditto

packages/customer-account-ui-extensions/src/components/shared-components.ts
```diff
@@ -1,2 +1,92 @@
...
+  ViewCoordinate,
+  AccessibilityRole,
+  ViewLikeAccessibilityRole,
+  NonPresentationalAccessibilityRole,
+  Autocomplete,
+  AutocompleteField,
+  AutocompleteGroup,
+  Breakpoint,
+  CSSShorthand,
+  Responsive,
```
jffortier, line 84: ditto

packages/customer-account-ui-extensions/src/components/PageBlock/PageBlock.ts
robin-drexler: A full page app can potentially have multiple `PageBlocks` so in this case, it makes sense to expose it. But yeah, we'll also spend some more time sketching out the exact components we need/want to expose soon.

PR #362, opened 2022-06-15
@shopify/web-pixesl-extension package

antoinegrant:
### Background

As part of the Web Pixel Manager project we need to introduce anWeb Pixel Extension npm package to facilitate the development.

Resolves: https://github.com/Shopify/ce-customer-behaviour/issues/1258

### Solution

Followed common patterns found in similar packages.

### Checklist


packages/web-pixel-ui-extensions/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../config/typescript/tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "build/ts",
+    "baseUrl": "src",
+    "rootDir": "src",
+    "isolatedModules": true
+  },
+  "include": ["src/**/*.ts"],
+  "references": [{"path": "../checkout-ui-extensions"}]
```
lemonmade, line 10: I don't think you need this reference

packages/web-pixel-ui-extensions/src/extension-points/web-pixel-api.ts
```diff
@@ -0,0 +1,23 @@
+export type Version = 'unstable';
```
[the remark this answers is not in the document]
lemonmade: Why won't this package be the source of those types?

packages/web-pixel-ui-extensions/src/extension-points/web-pixel-api.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * The following APIs are provided to the Pixel extension point.
+ */
+export interface WebPixelApi {
+  analytics: {
+    subscribe(event: string, callback: (data: any) => void): void;
+  };
+  config: Record<string, unknown>;
```
lemonmade, line 10: Nit: we called this `configuration` in checkout extensions

packages/web-pixel-ui-extensions/src/extension-points/render-extension.ts
```diff
@@ -0,0 +1,11 @@
+/**
```
lemonmade, line 1: You don't need this type

packages/web-pixel-ui-extensions/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "exports": {
+    ".": {
+      "esnext": "./index.esnext",
+      "import": "./index.mjs",
+      "require": "./index.js"
+    },
+    "./": "./"
+  },
+  "dependencies": {
+    "@remote-ui/core": "^2.1.10"
```
lemonmade, line 24: I don't believe you actually need this dependency

packages/web-pixel-ui-extensions/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "module": "index.mjs",
+  "esnext": "index.esnext",
+  "types": "index.d.ts",
+  "exports": {
+    ".": {
+      "esnext": "./index.esnext",
+      "import": "./index.mjs",
+      "require": "./index.js"
+    },
+    "./": "./"
```
lemonmade, line 21: You can remove this, it's deprecated and the other packages shouldn't have it either

packages/web-pixel-ui-extensions/src/extension-points/web-pixel-api.ts
```diff
@@ -0,0 +1,23 @@
+export type Version = 'unstable';
+
+/**
+ * The following APIs are provided to the Pixel extension point.
+ */
+export interface WebPixelApi {
+  analytics: {
+    subscribe(event: string, callback: (data: any) => void): void;
```
lemonmade, line 8: Will we have strong typing on this at some point?

lemonmade: Also, will there eventually be a way to unsubscribe, or do we not think these extensions need that?

packages/web-pixel-ui-extensions/README.md
```diff
@@ -0,0 +1,28 @@
+# Web Pixel Extension
+
+Provides an `extend` method to register a Web Pixel Extension.
+
+## Usage
+
+Registering a Web Pixel Extension:
+
+```js
+import {extend} from '@shopify/web-pixel-ui-extensions';
```
lemonmade, line 10: In a recent conversation with Build, we kind of landed on this convention where things that say `ui` in their name denote a particular kind of extension, in the model of post-purchase/ checkout/ subscription UI extensions. We excluded pixels from that list, opting to instead call it `web pixels` in the CLI selector where you generate a new one. I think including `ui-extensions` in the package name blurs that separation, and I'm feeling like `@shopify/web-pixels` would be better. cc/ @MeredithCastile

packages/web-pixel-ui-extensions/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+extend('WebPixel::Render', ({analytics, context, config}) => {
```
lemonmade, line 12: Do we think there will be more than one extension point for web pixels? I like that under the hood this uses the same registration setup as the other extension types, but I feel like it is unnecessarily verbose for consumers if they are likely to only ever implement one extension point. I wonder if it would be better to just get closer to what I am proposing in https://github.com/Shopify/ui-extensions-private/discussions/1653#discussioncomment-1927495, and use a signature that does not include the extension point, since in your case only one is relevant. Something like this is what I was imagining as a bridge to the world where we just use exports to denote extension registrations:

```ts
import {register} from '@shopify/web-pixels';

register(({analytics, context, config}) => {
  // do your thing in here
});
```

antoinegrant: So the only reason I went with `extend` was to follow the pattern and the rational was to bring consistency in the developer experience. If you think that having a different way of registering the Web Pixel Extension is acceptable then I'm on board.

packages/web-pixel-ui-extensions/src/extension-points/web-pixel-api.ts
antoinegrant: @sorta-shopify Should we make that change or would it put us back too much given the different places that needs to be updated!?

packages/web-pixel-ui-extensions/src/extension-points/web-pixel-api.ts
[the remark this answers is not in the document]
antoinegrant: I'd like to keep the types within the pixel manager repo and have this package pull the latest types. Unless you have another strategy that you'd like to share, I'm all ears.

packages/web-pixel-ui-extensions/src/extension-points/web-pixel-api.ts
sorta-shopify: I think it may not be too bad to change, and now might be the time do it since I have open PRs for related stuff. I would lean towards making this change now I think

packages/web-pixel-ui-extensions/README.md
lemonmade: Keeping `extension` is up to you, I think it works fine either way.

packages/web-pixel-ui-extensions/src/extension-points/web-pixel-api.ts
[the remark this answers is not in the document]
lemonmade: In checkout, we make the package in this repo be the source of truth, and we depend on it in our private repos that care about types it contains. I like this model better because it means the "real" value is something that both we and extensions can see, and it prevents any weird setups where consumers have to look at multiple repos to understand types or, worse yet, where some subset of the types they care about are in private repos.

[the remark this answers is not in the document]
antoinegrant: Sounds good, we're on board to have this package be the source of truth for types. Our hesitation was based on the fact that we generate schemas and their types for standard events and we weren't clear on wether we wanted that generation tool to live here. Seems like the best choice to have the schema definition here as it will serve as documentation for partners.

packages/web-pixel-ui-extensions/README.md
marvinhagemeister: Prefer the `register()`-API compared to `extend()` here too, unless we'll end up with multiple extension points for web pixels. The way I understand it, is that web pixels are not tied to a specific layout point and registering them is more of a "global" thing. For layout points the `extend()`-API makes more sense.

vividviolet: `register` is closer to what the developer is actually doing so I'm good with renaming `extend`.  Would this mean we'd support calling `register` with or without an extension point?

lemonmade: @vividviolet for Pixels, `register` should not take an extension point, since there is only one extension. In extensions, I think we should introduce this function alongside the "list your extension point => module mapping in your configuration file, we will do the rest" change in my proposal, which would mean it would also not accept an extension point identifier.

lemonmade: No, I don't think so. `register` in a vanilla extension could be omitted entirely when we move to using exports to declare registrations:

```ts
import {register} from '@shopify/checkout-ui-extensions';

function extension(root, {extensionPoint}) {
  root.appendChild(root.createText(`Rendering in ${extensionPoint}`));
}

// Equivalent
export default extension;
export default register(extension);
```

Having the hook of `register` lets us give better typing hints, and a hook for doing framework-specific bindings, e.g.:

```ts
import {registerReact} from '@shopify/checkout-ui-extensions-react';

export default registerReact(({extensionPoint}) => <>Rendering in {extensionPoint}</>);
```

packages/web-pixels-extension/src/types/standard-events-V0/index.ts
```diff
@@ -0,0 +1,59 @@
+// Code generated by jtd-codegen for TypeScript v0.2.1
```
antoinegrant, line 1: These type are generated with the following command:

```bash
$ cd packages/web-pixels-extension/
$ jtd-codegen ./standard-events-schemas/StandardEventsV0.jtd.json --typescript-out ./src/types/standard-events-V0
```

The plan is to add the documentation on how to generate new event schema versions in the Pixel Manager private repo.

sorta-shopify (commented):
Code looks good to me 👍 I'll let people better acquainted with extensions approve

antoinegrant:
@lemonmade I made the suggested changes, does that reflect what we had in mind?

packages/web-pixels-extension/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+## Usage
+
+Registering a Web Pixels Extension:
+
+```js
+import {register} from '@shopify/web-pixels-extension';
+
+register((
```
lemonmade, line 18: Formatting of this looks off

packages/web-pixels-extension/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "esnext": "index.esnext",
+  "types": "index.d.ts",
+  "exports": {
+    ".": {
+      "esnext": "./index.esnext",
+      "import": "./index.mjs",
+      "require": "./index.js"
+    }
+  }
+}
```
lemonmade, line 22: Missing newline

packages/web-pixels-extension/src/types/standard-events-V0/index.ts
```diff
@@ -0,0 +1,59 @@
...
+export interface StandardEventsV0CheckoutStarted {
+  event_name: 'checkout_started';
+  event_schema_version: string;
+  page_url: string;
+}
+
+export interface StandardEventsV0CollectionView {
+  event_name: 'collection_view';
+  event_schema_version: string;
+  page_url: string;
```
lemonmade, line 28: It's unfortunate we aren't making these into more JS-friendly identifiers

packages/web-pixels-extension/src/types/runtime-api.ts
```diff
@@ -0,0 +1,60 @@
...
+  readonly cookie: {
+    get: (name?: string) => Promise<string>;
+    set: (cookieOrName: string, value?: string) => Promise<string>;
+  };
+  readonly origin: {
+    get: () => Promise<string>;
+  };
+  readonly sendBeacon: (
+    url: string | URL,
+    data?: BodyInit | null,
```
lemonmade, line 58: Are you sure you can actually take anything that can be used to initialize a request? I think this includes things like streams that can't be proxied to the parent page.

packages/web-pixels-extension/README.md
```diff
@@ -0,0 +1,40 @@
...
+  {configuration, analytics, client}) => {
```
lemonmade, line 19: Why was `client` chosen as the name for this field? I find it a little confusing, personally, since it's actually talking about a bridge to what we often refer to as the `host` (basically, the opposite of `client`).

lemonmade (approved):

packages/web-pixels-extension/README.md
```diff
@@ -0,0 +1,40 @@
+# Web Pixels Extension
+
+Provides an `extend` method to register a Web Pixels Extension.
```
joshdavey, line 3: a "register" method?

packages/web-pixels-extension/src/types/runtime-api.ts
antoinegrant: Good catch, this should of been updated to `Record<string, unknown>`

packages/web-pixels-extension/src/types/standard-events-V0/index.ts
antoinegrant: CamelCase as opposed to snake_case

kencwng: we picked camel_case to follow Google Analytics' convention (along with many other partners like Facebook): https://developers.google.com/analytics/devguides/collection/ga4/ecommerce?client_type=gtag#view_item_list

packages/web-pixels-extension/README.md
lemonmade: I think `browser` or `page` is more evocative of what you are getting at with `client`, in that case. If it were me, I might just put all those objects as siblings of `configuration`/ `analytics`, if there aren't too many of them (that's basically what we do with checkout extensions, where every <thing> an extension can do is a top-level object in this "API object")

packages/web-pixels-extension/src/types/standard-events-V0/index.ts
antoinegrant: Yeah, I think we opted for snake_case to have a consistent event object between client and server. Are we still in favour of keeping consistent?

packages/web-pixels-extension/README.md
joshdavey: cool. let's do browser.

PR #365, opened 2022-06-17
[Checkout] Update UI extensions packages

jffortier:
### Background

### Solution

### 🎩

- ...

### Checklist


packages/checkout-ui-extensions/documentation/conditional-props.md
```diff
@@ -15,7 +15,7 @@ import {Style} from '@shopify/checkout-ui-extensions';
 If you're using React:
 
 ```jsx
-import {Style} from '@shopify/checkout-ui-extennsions-react';
```
jffortier, line 16: Good catch!

PR #373, opened 2022-06-21
Improve method rendering in literal type

igor10k:
### Background

There's a redundant colon in the method definition in literal types.

### Solution

Get rid of the colon for function properties.

### Checklist


jamesvidler (approved):
@igor10k great work! I tophatted this change with your `docs-tools` patch and verified this resolves the issue with the `buyerJourney` type.

jffortier (approved):

marvinhagemeister (approved):

PR #378, opened 2022-06-21
Web pixels extension fix readme

antoinegrant:
### Background

`async` was missing from the function passed to `register` method.

### Solution

Add the keyword `async` before the function declaration.

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/web-pixels-extension/README.md
+++ b/packages/web-pixels-extension/README.md
@@ -15,7 +15,7 @@ Registering a Web Pixels Extension:
 ```js
 import {register} from '@shopify/web-pixels-extension';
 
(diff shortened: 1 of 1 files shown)


sorta-shopify (approved):
Pretty straightforward. Is the package.json change expected though?

antoinegrant:
> The example fix looks good to me, too.
> 
> > Is the package.json change expected though?
> 
> I don't think you need to bump the package version. Developers will be reading the readme on GitHub, most likely.

It should not be updated. I'll fix that.

marvinhagemeister (approved):

PR #380, opened 2022-06-22
Workaround for extension errors not shown during initial render

jffortier:
### Background

Address issue https://github.com/Shopify/ui-extensions/issues/325

### Solution

This PR is a port of @marvinhagemeister done at the source.
https://github.com/Shopify/checkout-web/pull/11674

I've also set the same logs in the `customer-account-ui-extensions-react` which is a recent copy of the Checkout render function.

### Checklist

### DIFF
--- a/packages/customer-account-ui-extensions-react/src/render.tsx
+++ b/packages/customer-account-ui-extensions-react/src/render.tsx
@@ -33,6 +33,9 @@ export function render<ExtensionPoint extends RenderExtensionPoint>(
           },
         );
       } catch (error) {
+        // Workaround for https://github.com/Shopify/ui-extensions/issues/325
+        // eslint-disable-next-line no-console
+        console.error(error);
         reject(error);
       }
     });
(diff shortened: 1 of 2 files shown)


packages/customer-account-ui-extensions-react/src/render.tsx
```diff
@@ -33,6 +33,9 @@ export function render<ExtensionPoint extends RenderExtensionPoint>(
           },
         );
       } catch (error) {
+        // Workaround for https://github.com/Shopify/ui-extensions/issues/325
+        // eslint-disable-next-line no-console
+        console.error(error);
```
kumar303, line 38: Good idea to patch `customer-account-ui-extensions-react` 👍

robin-drexler (approved):

marvinhagemeister:
@jffortier Thank you so much for adding these workarounds here too! Wasn't aware that we had similar codepaths here 👍

PR #384, opened 2022-06-22
Improve type unpacking behavior

igor10k:
### Background

Some types are unpacked when they shouldn't be which affects readability.

image

### Solution

Improve the algo that finds repeated types.

Screen Shot 2022-06-22 at 5 24 51 PM

### Checklist


marvinhagemeister (approved):

jffortier:
This is great. I found the way we type some concepts results in some slightly annoying abstractions. The main impact is that devs would jump from type to types to finally understand the values accepted. I think one layer is acceptable, but in this case, its a bit too much. WDYT?

ex:
image

Note that I changed the `inlineAlignment` value to not map to `Alignment` before the beta because it was a dead end with the script at the time. It would be a bit more useful here if we'd revert that change and use `type InlineAlignment = Alignment`.

I think we can change the type of `blockAlignment` the same way to remove that second table, but I dont think thats a long term solution. This basically make our code worst (and probably larger) to make the docs happy.

This is were UX and the docs team could be helpful in setting patterns for these types of types.

jffortier (approved):

PR #413, opened 2022-07-25
Fix typo in public docs

andra-la:
### Background

Typo in the README and public docs for 'heading' checkout ui component. 

https://shopify.dev/api/checkout-extensions/components/heading

25-56-vhq65-hdtco

### Solution

Fix typo. 

### Help me

Are these docs automatically generated for Shopify.dev on a daily basis? Or is it manual?


marvinhagemeister (approved):
Good catch! Thank you so much for the PR 🙌

PR #417, opened 2022-07-28
Expose types for PaymentIcon

hsauce:
### Background

### Solution

### 🎩

- ...

### Checklist


packages/customer-account-ui-extensions/src/components/PaymentIcon/PaymentIcon.ts
```diff
@@ -0,0 +1,25 @@
+import {createRemoteComponent} from '@remote-ui/core';
+
+export interface PaymentIconProps {
+  /**
+   * The payment brand's icon to display.
+   */
+  paymentBrand: string;
```
hsauce, line 7: I can replace this with a string literal type that matches what is currently allowed in checkout-web-ui but there's no guarantee it will be kept up to date. [Link to payment icon mapping in checkout-web-ui](https://github.com/Shopify/checkout-web/blob/master/packages/checkout-web-ui/src/components/PaymentIcon/images/index.ts#L242).

mathieulag: I would probably still do the string literal type, we're gonna have to publicly document which one are available anyway. Partners won't have access to the implementation to check which one would work.

mathieulag (approved):

mathieulag:
Yep good point. Just added Tobias since all the other FEDs are off right now.

TobiasBales (approved):
LGTM

PR #428, opened 2022-08-08
Expose location to full page extensions

mathieulag:
### Background

Part of https://github.com/Shopify/core-issues/issues/42207. Exposes the current route data to full page extensions so they can do internal routing themselves.

### Solution

In addition to adding the `location` property in the API, this PR also adds a new hook called `useSubscription` that extensions can use in order to unwrap a `StatefulRemoteSubscribable` from the API. The hook simply transforms the subscribable into a reactive value within React's lifecycle.

### 🎩

Added our first test for the newly introduced hook.

### Checklist

### DIFF
--- a/packages/customer-account-ui-extensions-react/src/hooks/tests/subscription.test.ts
+++ b/packages/customer-account-ui-extensions-react/src/hooks/tests/subscription.test.ts
@@ -1,0 +1,73 @@
+import type {StatefulRemoteSubscribable} from '@remote-ui/async-subscription';
+import {act, renderHook} from '@testing-library/react-hooks';
+import {useSubscription} from '../subscription';
+
+describe('useSubscription()', () => {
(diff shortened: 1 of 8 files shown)


packages/customer-account-ui-extensions-react/src/hooks/tests/subscription.test.ts
```diff
@@ -0,0 +1,73 @@
+import type {StatefulRemoteSubscribable} from '@remote-ui/async-subscription';
+import {act, renderHook} from '@testing-library/react-hooks';
+import {useSubscription} from '../subscription';
+
+describe('useSubscription()', () => {
```
robin-drexler, line 5: ❤️ the tests

robin-drexler (approved):

emileber (approved):

PR #431, opened 2022-08-10
Add support for generic interfaces

igor10k:
⚠️ Do not merge until https://github.com/Shopify/docs-tools/pull/37 is merged

### Background

Currently the generic variables in interfaces are omitted as in:

```ts
export interface TextFieldProps<T extends string | number | undefined> {
```

or

```ts
export interface StepperProps extends Pick<TextFieldProps<number | undefined>, 'id'> {
```

### Solution

1. Make use of the update in `docs-tools`: https://github.com/Shopify/docs-tools/pull/37
2. Feed the generic variable as another export to the script.


jamesvidler:
@igor10k don't forget to bump the docs-tools version number.

PR #432, opened 2022-08-10
add Policy components to customer-account-ui-extensions packages

kimhawthorne-shopify:
### Background

### Solution

### 🎩

- ...

### Checklist


hsauce (approved):
LGTM
Screen Shot 2022-08-19 at 10 11 41 AM

emileber (approved):
The implementation looks fine! 👌 

I'm wondering though if we could still expose a generic Modal component and use it inside this `PolicyModal`?

As I understand it, extensions are limited in what they can render, so building the `PolicyModal` on the extension side is off the table, and I'm fine with that, but what about offering both?

kimhawthorne-shopify:
> The implementation looks fine! 👌
> 
> I'm wondering though if we could still expose a generic Modal component and use it inside this `PolicyModal`?
> 
> As I understand it, extensions are limited in what they can render, so building the `PolicyModal` on the extension side is off the table, and I'm fine with that, but what about offering both?

The Shopify Returns team has been adding only the components that we need for our beta, but I agree that I think extensions should have access to a plain ol' `Modal` component. Maybe as part of a separate story/PR?

packages/customer-account-ui-extensions/src/components/PolicyModal/PolicyModal.ts
```diff
@@ -0,0 +1,31 @@
...
+   * not when it exits the viewport.
+   * You’ll usually want to use this callback to set the `open` prop to `false`.
+   */
+  onClose: () => void;
+  /**
+   * A title rendered as a `Heading` at the top of the modal.
+   */
+  title: string;
+  /**
+   * Type of policy to render
```
mathieulag, line 23:
```suggestion
   * Type of policy to render.
```

mathieulag (approved):

PR #434, opened 2022-08-11
[Customer Account] Exclude tests from tsconfig.json

mathieulag:
### Background

I added a test in https://github.com/Shopify/ui-extensions/pull/428, and these files are not excluded from our tsconfig.json as they should be.

### Solution

Excluded tests from both `tsconfig.json`

### Checklist


robin-drexler:
> they should be

why should they be? 🤔

emileber (approved):

PR #435, opened 2022-08-11
Prepublish checkout ui extensions

igor10k:
`checkout-ui-extensions` and `checkout-ui-extensions-react` updates


jamesvidler (approved):
Looks good! Before we merge this, we'll need to also review the PR on `shopify-dev` to validate the doc generation on that front as well.

PR #437, opened 2022-08-12
Publish

emileber:
- @shopify/customer-account-ui-extensions-react@0.0.12
 - @shopify/customer-account-ui-extensions@0.0.12


robin-drexler (changes requested):
iirc We need to publish the git tags for deployment to work properly in this repo.

Like we did for 0.0.11 here: https://github.com/Shopify/ui-extensions/tree/%40shopify/customer-account-ui-extensions-react%400.0.11

robin-drexler (approved):

PR #453, opened 2022-08-25
#43320 add i18n api to the customer-account-ui

Safi1012:
### Background

✅ Adds the same `I18n` api to the `customer-account-ui` extension package
✅ Extend the existing `Localization` interface within the `customer-account-ui` extension package

The Self Serve team will later need this API in order to implement internationalization for their own extensions.

This task is a sub-part of the ticket: https://github.com/Shopify/core-issues/issues/42980

### Solution

- Our plan is to make use of the same API as the checkout team.

### 🎩

Tophatting is quite difficult as this PR is just extending the `StandardApi` interface of the `customer-account-ui` extension package.

### Checklist


emileber (approved):

packages/customer-account-ui-extensions/src/extension-points/standard-api.ts
```diff
@@ -13,11 +14,115 @@ export interface Language {
...
+   */
+  isoCode: CurrencyCode;
+}
+
 export interface Localization {
   /**
-   * The language the buyer sees in the customer account hub.
+   * The currency that the buyer sees for money amounts in the checkout.
+   */
+  currency: StatefulRemoteSubscribable<Currency>;
```
mathieulag, line 29: Do we have a currency in Self-Serve to provide? Where is it coming from?

packages/customer-account-ui-extensions/src/extension-points/standard-api.ts
```diff
@@ -13,11 +14,115 @@ export interface Language {
...
+   *
+   * This function behaves like the standard `Intl.NumberFormat()`
+   * with a style of `currency` applied. It uses the buyer's locale by default.
+   *
+   * @param options.inExtensionLocale - if true, use the extension's locale
+   */
+  formatCurrency: (
+    number: number | bigint,
+    options?: {inExtensionLocale?: boolean} & Intl.NumberFormatOptions,
+  ) => string;
```
mathieulag, line 97: I wonder what currency should this method use? I understand that in Checkout the have a currency, but in Self-Serve a currency only exists in specific context, ie within an order.

I would need to see the implementation to understand a bit better whether this works for our use case.

emileber: The function itself makes sense since a currency may be present within the data and an extension might want to render it.

Though I do not believe we have a default currency for the user right now, just like we do not have a different locale for the extension even though it could be added in the future.

Safi1012:
Thanks @emileber and @mathieulag for your feedback! I will leave this PR open until the actual implementation is finished so we know exactly whether the API matches the implementation.

packages/customer-account-ui-extensions/src/extension-points/standard-api.ts
Safi1012: Update: I decided to not include currency in our API for now, as I did not have a default currency for the user right now. However, we can always add that in the future.

Thanks @emileber

packages/customer-account-ui-extensions/src/extension-points/standard-api.ts
```diff
@@ -15,9 +16,87 @@ export interface Language {
 
 export interface Localization {
   /**
-   * The language the buyer sees in the customer account hub.
+   * The buyer’s time zone.
+   */
+  timezone: StatefulRemoteSubscribable<Timezone>;
+
+  /**
+   * The language the buyer sees in the checkout.
```
robin-drexler, line 24:
```suggestion
   * The language the buyer sees in the customer account hub
```

robin-drexler (approved):

Safi1012:
Since everyone agreed with this new API in our last localization sync meeting I will ship it, so the code is available to the other PR (https://github.com/Shopify/customer-account-web/pull/1023).

lemonmade (approved):
Thank you for copying what we did in checkout!

PR #457, opened 2022-08-26
add run extensions and initiate returns extension point

robin-drexler:
### Background

Part of https://github.com/Shopify/core-issues/issues/43001

Adds typings for the endpoint that we'll use to allow extensions to initiate a return.

Also removes `ApiForRenderExtension` type in favor of `ApiForExtension`.

Explanation can be found here: https://github.com/Shopify/ui-extensions/pull/457/files#r958631838

### 🎩 
Please check: https://github.com/Shopify/customer-account-web/pull/1009 description

### Checklist

### DIFF
--- a/packages/customer-account-ui-extensions/src/extension-points/extension-points.ts
+++ b/packages/customer-account-ui-extensions/src/extension-points/extension-points.ts
@@ -1,5 +1,5 @@
 import type {StatefulRemoteSubscribable} from '@remote-ui/async-subscription';
-import {RenderExtension} from './render-extension';
+import type {RenderExtension, RunExtension} from './extension-signature';
 import type {StandardApi} from './standard-api';
 
 type Components = typeof import('../components');
@@ -18,9 +18,17 @@ export interface ExtensionPoints {
     StandardApi & FullPageApi,
     AllComponents
   >;
+  'CustomerAccount::Returns::Initiate': RunExtension<
+    StandardApi & {orderId: string},
+    void
+  >;
   'CustomerAccount::KitchenSink': RenderExtension<
     StandardApi & {name: string},
     AllComponents
+  >;
+  'CustomerAccount::KitchenSinkRun': RunExtension<
+    StandardApi & {name: string},
+    string
   >;
 }
 
--- a/packages/customer-account-ui-extensions/src/extension-points/index.ts
+++ b/packages/customer-account-ui-extensions/src/extension-points/index.ts
@@ -1,9 +1,11 @@
 export type {ExtensionPoints, ExtensionPoint} from './extension-points';
-export type {RenderExtension} from './render-extension';
 export type {StandardApi, Language, Localization} from './standard-api';
 export type {
-  ApiForRenderExtension,
+  ApiForExtension,
   ArgumentsForExtension,
   RenderExtensionPoint,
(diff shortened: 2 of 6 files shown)


packages/customer-account-ui-extensions/src/extension-points/types.ts
```diff
@@ -55,3 +89,16 @@ type ExtractedApiFromRenderExtension<T> = T extends RenderExtension<
 export type ApiForRenderExtension<
   ID extends keyof RenderExtensions
 > = ExtractedApiFromRenderExtension<RenderExtensions[ID]>;
+
+/**
+ * For a given run extension point, returns the type of the API that the
+ * extension will receive at runtime. This API type is the first argument to
+ * the callback for that extension point.
+ */
+export type ApiForRunExtension<
```
robin-drexler, line 98: Check if we can remove this and also `ApiForRenderExtension` now that we do have `ApiForExtension`

packages/customer-account-ui-extensions/src/extension-points/index.ts
```diff
@@ -1,9 +1,11 @@
 export type {ExtensionPoints, ExtensionPoint} from './extension-points';
-export type {RenderExtension} from './render-extension';
 export type {StandardApi, Language, Localization} from './standard-api';
 export type {
-  ApiForRenderExtension,
```
robin-drexler, line 5: This is a breaking change, however the returns app is the only consumer and the fix is rather trivial 

```diff
diff --git a/src/type.ts b/src/type.ts
index eefe916..7362d33 100644
--- a/src/type.ts
+++ b/src/type.ts
@@ -1,6 +1,6 @@
-import type {ApiForRenderExtension} from '@shopify/customer-account-ui-extensions';
+import type {ApiForExtension} from '@shopify/customer-account-ui-extensions';
 
-export type ExtensionApi = ApiForRenderExtension<
+export type ExtensionApi = ApiForExtension<
   'CustomerAccount::FullPage::RenderWithin'
 >;
```

packages/customer-account-ui-extensions/src/extension-points/extension-points.ts
```diff
@@ -18,10 +18,18 @@ export interface ExtensionPoints {
     StandardApi & FullPageApi,
     AllComponents
   >;
+  'CustomerAccount::Returns::Initiate': RunExtension<
```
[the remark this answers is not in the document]
emileber: Ok, so after a quick chat, I now understand that this extension point is entirely temporary, and even if we go with a similar one in the future, we'll still create a new extension point and refactor the return flow to use it.

So this comment can be disregarded! 🎉

packages/customer-account-ui-extensions/src/extension-points/types.ts
```diff
@@ -33,11 +33,18 @@ export type RenderExtensionPoint = {
 }[keyof ExtensionPoints];
 
 /**
- * A mapping of each “render extension” name to its callback type.
+ * A union type containing all extension points that follow the pattern of
+ * accepting an `api` argument, and using those arguments to run code that does not render anything, but instead return
```
Safi1012, line 37: Thanks for the comment 🙌

Safi1012 (approved):
🎉

PR #466, opened 2022-09-08
added optional field  to web pixel RuntimeAPI

sandor-shopify:
### Background

ISSUE: https://github.com/Shopify/ce-customer-behaviour/issues/1461

we must rename `configuration` to `settings` in the Runtime API.

### Solution

Adding in an optional `settings` property to support the transition until we are confident in the solution downstream

### 🎩

- ...

### Checklist


sandor-shopify:
> LGTM, assuming we will remove `configuration` and change `settings?` to `settings` after we ship the web-pixels-manager change

yes, this is so that we can release the settings and update in the references in the gem. Afterwards we will remove the `configuration` field

sorta-shopify (approved):
👍

PR #476, opened 2022-09-13
Removed src from file imports

js-goupil:
### Background

There are still a few places in our react package where we specify `src` in imports from packages. Since the imports are coming from the vanilla package, we should just import it directly as everything in that package is exported properly. 

### Solution

Just removes the src from paths and import from the package.

### 🎩

Everything should build normally. It's still importing the same thing, but from the correct path (which points to the correct build output)

### Checklist

### DIFF
--- a/packages/retail-ui-extensions-react/src/context.ts
+++ b/packages/retail-ui-extensions-react/src/context.ts
@@ -2,7 +2,7 @@ import {createContext} from 'react';
 import {
   ApiForRenderExtension,
   RenderExtensionPoint,
-} from '@shopify/retail-ui-extensions/src';
+} from '@shopify/retail-ui-extensions';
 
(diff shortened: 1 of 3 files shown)


samiulmonir06 (approved):
Looks good

packages/retail-ui-extensions-react/src/context.ts
```diff
@@ -2,7 +2,7 @@ import {createContext} from 'react';
 import {
   ApiForRenderExtension,
   RenderExtensionPoint,
-} from '@shopify/retail-ui-extensions/src';
+} from '@shopify/retail-ui-extensions';
```
samiulmonir06, line 5: Is it going to fix [this issue](https://github.com/Shopify/pos-next-react-native/issues/13732)

heltisace (approved):

PR #480, opened 2022-09-15
Web pixels update types new schema

antoinegrant:
### Background
As part of the Web Pixels Manager Project, we have updated the schema to be more comprehensive. 

### Solution
This introduces the types based on the new schema. 

Note: both the new and the previous types apply until we make the full switch to new schema.

### Checklist


sorta-shopify (approved):
👍

PR #491, opened 2022-09-19
Change kind's @default value to secondary in documentation

MaxCloutier:
### Background
The default value of kind in the Button is `secondary` and not `primary` [more details in the issue here](https://github.com/Shopify/ui-extensions-private/issues/1674)

### Solution

Change the documentation to `secondary`


olavoasantos (approved):
LGTM

PR #492, opened 2022-09-19
Add CountryCode type and update Address Cart API functions

NathanJolly:
### Background

The POS expects a CountryCode on address addition. This is a copy of the type from https://github.com/Shopify/pos-next-react-native/blob/9da8142ab3df17c58f797f6ab6327236ad4d4d03/src/data/types/admin/CountryCode.ts

Additionally, the functions available to manage addresses are https://github.com/Shopify/pos-next-react-native/blob/ea4f14b47ec37907583e430483a32942798eac3d/src/datasource/Shopify/Customer/CustomerHook.ts#L202-L204 here, so I believe the extensions package should match this.

### Solution

Create a copy of the CountryCodes in the `retail-ui-extensions` package.
Update the Cart API's Address-related functions to match the POS hook.

### Checklist


js-goupil (approved):

js-goupil:
Once it merges we'll have to release a new version of the package and also update the react package to export this type.

PR #505, opened 2022-09-21
Propagate export to top of package

NathanJolly:
### Background

Now it was in src/extension-api/types/index, but not propagated to the top level index.

### Solution

Propagate it to the top! (Learning process... :) )

### 🎩

- ...

### Checklist


coderBackPacker (approved):
makes sense

PR #508, opened 2022-09-22
Change uuid type to match other api calls and POS redux state

coderBackPacker:
### Background
While working on implementation of cart API calls in POS, I noticed that `setLineItemDiscount` takes uuid as number. However all other calls uses uuid as string which also matched how redux state stores the lineItem uuid
Screen Shot 2022-09-22 at 10 14 38 AM

### Solution

This PR is a tiny change to change the type to string to match all other calls

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/extension-api/cart-api/cart-api.ts
+++ b/packages/retail-ui-extensions/src/extension-api/cart-api/cart-api.ts
@@ -7,7 +7,11 @@ export interface CartApi {
   cart: CartApiContent;
 }
 
-export type DiscountType = 'Percentage' | 'FixedAmount' | 'Code';
+export type DiscountType =
+  | 'Percentage'
+  | 'FixedAmount'
+  | 'PriceOverride'
+  | 'Code';
 
 export interface CartApiContent {
   /** Initial value of the Cart when the screen is rendered. */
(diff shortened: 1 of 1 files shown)


packages/retail-ui-extensions/src/extension-api/cart-api/cart-api.ts
```diff
@@ -82,7 +82,7 @@ export interface CartApiContent {
    * @param amount the percentage or fixed monetary amount deducted with the discout
    */
   setLineItemDiscount(
-    uuid: number,
+    uuid: string,
```
js-goupil, line 85: strange that this was a number in the first place 🤨

js-goupil (approved):

packages/retail-ui-extensions/src/extension-api/cart-api/cart-api.ts
```diff
@@ -82,7 +82,7 @@ export interface CartApiContent {
...
     type: DiscountType,
```
NathanJolly, line 86: Should we also address the DiscountType while we're here? This type includes DiscountCode, which is not valid for line item discounts. There should be a separate type that excludes DiscountCode to account for this.

coderBackPacker: good point

PR #509, opened 2022-09-22
Publish

NathanJolly:
- @shopify/retail-ui-extensions-react@0.15.0

### Background

Adding the CountryCode type

### Solution

Bumping 3 versions to match the vanilla package, now that they're in sync

### 🎩

- ...

### Checklist


js-goupil:
Should we wait for Rashad's changes?

NathanJolly:
I think having 0.15 of both for countrycode and then 0.16 for the discount changes is right. If I hadn't already published 0.15 of vanilla, i'd agree that holding off and combining would be best.

coderBackPacker (approved):

PR #511, opened 2022-09-22
Add missing exports to package.json for admin-ui-extensions to fix build issue

MaxCloutier:
Fixes: https://github.com/Shopify/shopify-cli-extensions/issues/383

### Background

Missing exports in the package.json had extensions build failing

### Solution

Added missing exports

### DIFF
--- a/packages/admin-ui-extensions/package.json
+++ b/packages/admin-ui-extensions/package.json
@@ -11,7 +11,17 @@
       "import": "./index.mjs",
       "require": "./index.js"
     },
-    "./*": "./*"
+    "./*": "./*",
+    "./extension-api": {
+      "esnext": "./extension-api.esnext",
+      "import": "./extension-api.mjs",
+      "require": "./extension-api.js"
+    },
+    "./extension-points": {
+      "esnext": "./extension-points.esnext",
+      "import": "./extension-points.mjs",
+      "require": "./extension-points.js"
+    }
   },
   "license": "MIT",
   "sideEffects": false,


lemonmade (commented):
Why are we allowing developers to import from a subpath of this package? Checkout has more APIs and components, and you still always import from the root (https://github.com/Shopify/ui-extensions/blob/main/packages/checkout-ui-extensions/package.json#L15-L21). IMO, this is the cleanest-looking approach for developers, it's not fun to have to manually import from a bunch of different package subpaths.

olavoasantos:
I don't think this was broken as the [broken flow that was reported](https://github.com/Shopify/shopify-cli-extensions/issues/383) was working prior to a recent release. It just had been a very long time since we published this library and we published a patch this week.

As I pointed out above, simply removing this API is clearly breaking dependents. And so, removing this API would consist on a breaking change and should be done through a major version release. Maybe we can align this for the next major release, but I don't think this is reason enough to do it now.

lemonmade:
@vividviolet thank you for the pointer to the React library that is making use of these subpaths. I agree this is a good pattern, and better than checkout's version where we manually re-export everything (https://github.com/Shopify/ui-extensions/blob/main/packages/checkout-ui-extensions-react/src/index.ts). Can you open an issue to adopt this in the other packages? It would be great to have an agreed-upon set of entrypoints into the packages (for example, I'd personally prefer "api" over "extension-api"; I have been playing with an adjusted structure for when we have a single extension type for the different surfaces https://github.com/Shopify/ui-extensions-internal-test/tree/main/packages/ui-extensions/src/surfaces/checkout).

Separately, any reason why we have the "you can grab everything" mapping, `"./*": "./*"` (https://github.com/Shopify/ui-extensions/blob/main/packages/admin-ui-extensions/package.json#L14)? Because of how Loom builds packages, this is almost never desirable (the build outputs are deeply nested, with a weird directory structure).

vividviolet:
>  I have been playing with an adjusted structure for when we have a single extension type for the different surfaces https://github.com/Shopify/ui-extensions-internal-test/tree/main/packages/ui-extensions/src/surfaces/checkout).

@lemonmade Cool! This makes sense to me. We'd have a set of shared components at the top that can be re-exported by the different surface areas.

I think [this](https://github.com/Shopify/ui-extensions-internal-test/blob/a8ecb41bac47e692dac9af0246abedffe57ee2ab/packages/ui-extensions/src/surfaces/checkout/shared.ts#L5) can be made into a generic too in order to share between the surface areas.

I create the issue to follow up around standardizing re-exports [here](https://github.com/Shopify/ui-extensions-private/issues/1809).

PR #531, opened 2022-10-03
Converting Navigation API into two separate apis

samiulmonir06:
### Background

https://github.com/Shopify/pos-next-react-native/issues/17043

We are planning to converting navigation-api into two separate apis (SmartGrid and Navigation) APIs. SmartGrid API will open the modal from SmartGrid tiles whereas navigation api will control navigation flow inside the app. Currently navigation api will only support dismissing modal.

### Solution

Created a separate API to handle SmartGrid api.

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/extension-api/smartgrid-api/smartgrid-api.ts
+++ b/packages/retail-ui-extensions/src/extension-api/smartgrid-api/smartgrid-api.ts
@@ -1,6 +1,6 @@
 export interface SmartGridApiContent {
   /** Opens a full screen with the Retail::SmartGrid::Modal extension point */
-  navigateToFullScreenModal(): void;
+  presentModal(): void;
 }
 
 /**
--- a/packages/retail-ui-extensions/src/extension-api/standard-api.ts
+++ b/packages/retail-ui-extensions/src/extension-api/standard-api.ts
@@ -1,7 +1,6 @@
 import type {CartApi} from './cart-api';
 import type {LocaleApi} from './locale-api';
 import type {NavigationApi} from './navigation-api';
-import type {SmartGridApi} from './smartgrid-api';
 import type {SessionTokenApi} from './session-token-api';
 
 export type StandardApi<T> = {[key: string]: any} & {
@@ -9,5 +8,4 @@ export type StandardApi<T> = {[key: string]: any} & {
 } & LocaleApi &
   CartApi &
   NavigationApi &
-  SmartGridApi &
   SessionTokenApi;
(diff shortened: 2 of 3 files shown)


packages/retail-ui-extensions/src/extension-api/navigation-api/navigation-api.ts
```diff
@@ -1,13 +1,10 @@
...
-  /** Opens a full screen with the Retail::SmartGrid::Modal extension point */
-  navigateToFullScreenModal(): void;
-
   /** Dismisses the modal highest on the stack */
   dismiss(): void;
 }
 
 /**
- * Access the navigation API to present your app in a full screen modal.
+ * Access the navigation API to dismiss your app from a full screen modal.
```
NathanJolly, line 7: I think the idea here is to make the navigation API something that can expand over time, not just dismiss.

Maybe something like
"Access the navigation API for navigation functionality from a full screen modal."

packages/retail-ui-extensions/src/extension-api/smartgrid-api/smartgrid-api.ts
```diff
@@ -0,0 +1,11 @@
+export interface SmartGridApiContent {
+  /** Opens a full screen with the Retail::SmartGrid::Modal extension point */
+  navigateToFullScreenModal(): void;
```
js-goupil, line 3: What do you think if we would rename this to `presentModal`?

vctrchu: Yeah that sounds better and is easier to understand

packages/retail-ui-extensions/src/extension-api/standard-api.ts
```diff
@@ -1,11 +1,13 @@
...
 import type {LocaleApi} from './locale-api';
 import type {NavigationApi} from './navigation-api';
+import type {SmartGridApi} from './smartgrid-api';
 import type {SessionTokenApi} from './session-token-api';
 
 export type StandardApi<T> = {[key: string]: any} & {
   extensionPoint: T;
 } & LocaleApi &
   CartApi &
   NavigationApi &
```
js-goupil, line 11: Let's also remove this from the StandardAPI, and only include it in the other [extension point ](https://github.com/Shopify/ui-extensions/blob/6a3fbaa0d197fc11a4bf8a6e143b9215228171dd/packages/retail-ui-extensions/src/extension-points/extension-points.ts#L13)

js-goupil (approved):
:chef_kiss

PR #533, opened 2022-10-03
Bump retail-ui-extensions to 0.18.0

RcKeller:
### Background

Bumps retail-ui-extensions to v18, thus publishing changes to the card schema in https://github.com/Shopify/ui-extensions/pull/532

### Solution

### 🎩

- ...

### Checklist


heltisace (commented):
Will react version get the changes above? I think what we need to do is custom bump vanilla to 0.19.0, and then incremental bump the react version.

PR #546, opened 2022-10-06
added Stepper props and component

coderBackPacker:
### Background

Added Stepper component to `retail-ui-extensions` and `retail-ui-extensions-react` packages. 
I added the minimum props I believe we need to make the component functional
```ts
export interface StepperProps {
  initialValue: number;
  onValueChanged: (value: number) => void;
  useDefaultMargin?: boolean;
  isDisabled?: boolean;
}
```

### 🎩

- ...

### Checklist


packages/retail-ui-extensions/src/components/Stepper/Stepper.ts
```diff
@@ -0,0 +1,12 @@
+import {createRemoteComponent} from '@remote-ui/core';
+
+export interface StepperProps {
+  initialValue: number;
+  onValueChanged: (value: number) => void;
+  useDefaultMargin?: boolean;
+  isDisabled?: boolean;
```
NathanJolly, line 7: Total nitpick, isn't the prop typically called just `disabled`? https://www.w3schools.com/jsref/prop_pushbutton_disabled.asp

coderBackPacker: You are absolutley right, I like `disabled` more. You might see a lot of is `isSomething` in POS code base though

packages/retail-ui-extensions/src/components/Stepper/Stepper.ts
```diff
@@ -0,0 +1,12 @@
+import {createRemoteComponent} from '@remote-ui/core';
+
+export interface StepperProps {
+  initialValue: number;
+  onValueChanged: (value: number) => void;
+  useDefaultMargin?: boolean;
```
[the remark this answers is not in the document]
coderBackPacker: It impacts the margin of the horizontalStack inside the stepper component. It is more cosmetic but we can remove it. It defaults to `false` anyway

PR #548, opened 2022-10-07
Made staff member id optional because it may not exist on POS

heltisace:
### Background

I'm connecting the new `SessionApi` to POS and `staffMemberId` can be `undefined`. 

### Solution

Made the parameter optional in `SessionApiContext`. 

### 🎩

- ...

### Checklist


js-goupil (approved):
Good catch. This is usually the case only when the screen is locked, but yeah since that screen is an overlay i imagine it can cause issues underneath

PR #550, opened 2022-10-07
Updated, cross-surface ui-extension library

lemonmade:
This PR implements an initial version of the cross-surface `@shopify/ui-extension` library, which was primarily discussed in this RFC: https://github.com/Shopify/ui-extensions-private/discussions/1770. It implements just a few, basic components and extension points that will allow us to validate it in checkout and the admin.

I intend to publish this as `2.0.0-beta.0`, and to continue publishing new `.beta` versions as we iterate on this idea. I already removed the only dependency on this package in https://github.com/Shopify/ui-extensions/pull/543, so it should be safe to mess around with this package as we see fit.

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout/extension.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/extension.ts
@@ -1,47 +1,7 @@
-import {createRemoteRoot} from '@remote-ui/core';
+import {createExtensionRegistrationFunction} from '../../utilities/registration';
 
-import type {
-  RenderExtension,
-  RenderExtensionWithRemoteRoot,
-} from '../../extension';
-
-import type {ExtensionPoint, ExtensionPoints} from './extension-points';
+import type {ExtensionPoints} from './extension-points';
 
 export * from '../../extension';
 
-export function extension<Point extends ExtensionPoint>(
-  _extensionPoint: Point,
-  extensionWithRemoteRoot: ExtensionPoints[Point] extends RenderExtension<
-    infer Api,
-    infer AllowedComponents
-  >
-    ? RenderExtensionWithRemoteRoot<Api, AllowedComponents>
-    : ExtensionPoints[Point],
-): ExtensionPoints[Point] {
-  async function extension(...args: any[]) {
-    if (
-      args.length === 1 ||
-      typeof args[0] !== 'object' ||
-      args[0] == null ||
-      !('channel' in args[0])
-    ) {
-      const result = await (extensionWithRemoteRoot as any)(...args);
-      return result;
-    }
-
-    const [{channel, components}, ...rest] = args;
-
-    const root = createRemoteRoot(channel, {components, strict: true});
-
-    const result = await (extensionWithRemoteRoot as any)(root, ...rest);
-    await root.mount();
-
-    return result;
-  }
-
-  return extension as any;
-}
-
-extension('Checkout::CartLineDetails::RenderAfter', (root, {target}) => {
-  root.appendChild(root.createComponent('Text', null, `Hello, ${target.id}`));
-});
+export const extension = createExtensionRegistrationFunction<ExtensionPoints>();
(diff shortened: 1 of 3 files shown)


packages/ui-extensions/src/api.ts
```diff
@@ -0,0 +1,8 @@
+export interface ExtensionInstance {
+  readonly id: string;
```
vividviolet, line 2: What is this `id` for?

packages/ui-extensions/src/components/TextField/TextField.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface TextFieldProps {
+  /**
+   * The current value for the field. If omitted, the field will be empty. You should
+   * update this value in response to the `onChange` callback.
+   */
+  value?: string;
+
+  /**
+   * Callback when the buyer has **finished editing** a field. Unlike `onChange`
```
vividviolet, line 11:
```suggestion
   * Callback when the user has **finished editing** a field. Unlike `onChange`
```

packages/ui-extensions/src/components/TextField/TextField.ts
```diff
@@ -0,0 +1,40 @@
...
+   * callbacks you may be familiar with from Polaris or other React component libraries,
+   * this callback is **not** run on every change to the input. Text fields are
+   * “partially controlled” components, which means that while the buyer edits the
```
vividviolet, line 14:
```suggestion
   * “partially controlled” components, which means that while the user edits the
```

vividviolet (commented):
Looks good to me - just a few questions/suggestions

packages/ui-extensions/src/components/TextField/TextField.ts
lemonmade: Thank you for catching these!

packages/ui-extensions/src/surfaces/checkout/extension.ts
```diff
@@ -0,0 +1,47 @@
...
+import type {
+  RenderExtension,
+  RenderExtensionWithRemoteRoot,
+} from '../../extension';
+
+import type {ExtensionPoint, ExtensionPoints} from './extension-points';
+
+export * from '../../extension';
+
+export function extension<Point extends ExtensionPoint>(
```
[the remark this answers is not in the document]
lemonmade: It uses different underlying `ExtensionPoint`/ `ExtensionPoints` types. The goal was that TypeScript would not let you do something like this:

```ts
import {extension} from '@shopify/ui-extension/admin';

export default extension('Checkout::...', () => {});
```

That said, I could get rid of that and create one big extension point namespace. Alternatively, I could keep the per-surface version of this function, but have a shared function that creates the "wrapped" extension. What do you think?

packages/ui-extensions/src/api.ts
lemonmade: This is mostly just to have a field on this object. In checkout extensions, we use this "instance" object to reflect to the extension details about itself, including metadata (the extesnion version, script URL, etc), and runtime information (whether the extension is rendered, what capabilities it has). I'd like to have a place for that information to live in this package, so I added this type. I didn't want to get into any of those other concepts that will likely live here, so I just opted for a single field that is kind of a baseline expectation: a unique identifier for the running "instance".

If you feel it's too confusing, I can remove it for now. Checkout does have this concept, so I will eventually add it back, but I don't need it for getting the editor extension point off the ground.

packages/ui-extensions/src/surfaces/checkout/extension.ts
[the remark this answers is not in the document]
lemonmade: I've moved this wrapping concept into a shared bit of code 👍

packages/ui-extensions/src/api.ts
vividviolet: We have this concept in Admin as well it's tied to the version `uuid` we get from GraphQL. Can a single extension version have multiple `id`?

lemonmade: Yes, but probably not in the admin? In checkout, you can both install a single extension in multiple spots, and a single extension may render multiple times, if it targets a piece of UI that repeats (currently, this is just the `CartLineDetails::RenderAfter` extension point.

vividviolet: I see - I think we can just leave it as `id` then. Admin can just use the `uuid` for this value

PR #570, opened 2022-10-24
Added a selectable element

js-goupil:
### Background

We need a way for developers to be able to wrap custom UI elements into a selectable UI. Currently, they only way for them to make something selectable is by having a Button, or using the lists. So for more complex UI like the one in the linked figma, it's currently not possible.

### Solution

Introducing selectable! You can use selectable to wrap custom UI into something selectable. On the POS side, we will implement it around [TouchableButton](https://github.com/Shopify/pos-next-react-native/blob/main/src/components/TouchableButton/TouchableButton.tsx).

Here is a code sample:

```tsx
const CustomListItem = ({title, subtitle, onPress}) => {
  return (
    <Selectable onPress={onPress} disabled={false}>
      <Stack direction="horizontal">
        <Stack direction="vertical" spacing={1}>
          <Text>{title}</Text>
          <Text variant="captionRegular">{subtitle}</Text>
        </Stack>
      </Stack>
    </Selectable>
  );
};
```

### 🎩

Real pain to tophat, but you'll have to link this dependency to your local projects. Just take my word for it, it's a simple enough change 😄 

### Checklist


samiulmonir06 (commented):
LGTM, left a comment

packages/retail-ui-extensions/src/components/Selectable/Selectable.ts
```diff
@@ -0,0 +1,10 @@
+import {createRemoteComponent} from '@remote-ui/core';
+
+export interface SelectableProps {
```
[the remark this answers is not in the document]
js-goupil: If you look at the way Stack is implemented, you'll see it doesn't have children but it definitely works!

packages/retail-ui-extensions/src/components/Selectable/Selectable.ts
```diff
@@ -0,0 +1,10 @@
...
+  onPress: () => void;
+  disabled?: boolean;
```
coderBackPacker, line 5: should we use a default value (true)?

coderBackPacker (approved):
I have one comment but everything else is clear and straight forward

packages/retail-ui-extensions/src/components/Selectable/Selectable.ts
js-goupil: By default it should be enabled, so we'll use false. But that's an implementation detail on POS (if they omit this variable)

PR #572, opened 2022-10-24
Remove 'timezone' from customer-account-ui-extensions and export `I18nTranslate`

Safi1012:
### Background

- Remove `timezone` from `packages/customer-account-ui-extensions/src/extension-points/standard-api.ts`, as we currently do not have support for `timezone` within customer-accounts-web (I mistakenly added this last week)
- Properly export `I18nTranslate` in `customer-account-ui-extensions` so the interface can be imported by consumers of this library.

### 🎩

There is not much to Tophat here.

### Checklist

- [x] I have updated relevant documentation (WIP)


robin-drexler:
Do we want to add those types here too: https://github.com/Shopify/ui-extensions/blob/main/packages/customer-account-ui-extensions-react/src/index.ts#L2

That will allow consumers of [customer-account-ui-extensions-react](https://github.com/Shopify/ui-extensions/tree/main/packages/customer-account-ui-extensions-react) to use them as well

robin-drexler (commented):
I feel like I've already commented this, but it got either lost, I never actually submitted it or I did that on the wrong PR, but:

Can we re-export `I18nTranslate` in the `customer-account-ui-extensions-react` package as well so that consumers of that package have access as well?

robin-drexler (approved):

PR #604, opened 2022-11-14
Added a new Section component

js-goupil:
### Background

As we dive deeper into the Stocky beta and have a closer look at some of the pain points and designs, it's becoming obvious that we need some sort of Card, or Section component. 

This new interface will allow developers to wrap their content into a Seciton, with takes a title, as well as an optional action to wrap their content to reflect something like[this](https://www.figma.com/file/A0B96slYSNa9fw3UrMWdt2/Retail---Components?node-id=14602%3A7832) on the POS side.

### Solution

The only modifiable things are the title, and the action in the props. The content will be automatically inserted as children. The idea is that the separators are optional.

### 🎩

Still needs to be implemented on the POS side.

### Checklist


vctrchu (approved):

coderBackPacker (approved):
LGTM

PR #614, opened 2022-11-17
Navigation API pop function

heltisace:
### Background

I was working on https://github.com/Shopify/pos-next-react-native/issues/17783 and it makes sense to include a `pop` function in our navigation, even if the navigation itself ends up being temporary. Having the option to only dismiss the whole modal doesn't make much sense. 

### Solution

Added the interface for the `pop` function. 

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/extension-api/navigation-api/navigation-api.ts
+++ b/packages/retail-ui-extensions/src/extension-api/navigation-api/navigation-api.ts
@@ -7,6 +7,9 @@ export interface NavigationApiContent {
   /** The parameters the screen was displayed with */
   params?: any;
 
+  /** Pops the currently presented screen */
+  pop(): void;
+
   /** Dismisses the modal highest on the stack */
   dismiss(): void;
 


vctrchu (approved):
Makes sense

packages/retail-ui-extensions/src/extension-api/navigation-api/navigation-api.ts
```diff
@@ -7,6 +7,9 @@ export interface NavigationApiContent {
   /** The parameters the screen was displayed with */
   params?: any;
 
+  /** Pops the currently presented screen */
+  pop(): void;
```
NathanJolly, line 11: Is there a case where you'd want to pass `params` here?

packages/retail-ui-extensions/src/extension-api/navigation-api/navigation-api.ts
```diff
@@ -7,6 +7,9 @@ export interface NavigationApiContent {
   /** The parameters the screen was displayed with */
   params?: any;
 
+  /** Pops the currently presented screen */
```
js-goupil, line 10: Should we use a more meaningful word than "Pops" for 3P developers? Most of these people likely are not mobile devs

js-goupil (approved):

packages/retail-ui-extensions/src/extension-api/navigation-api/navigation-api.ts
samiulmonir06: I think `pop` works as it is an underlaying functionality for any browser navigation.

js-goupil: @samiulmonir06 I didn't know that, then pop it must be 😄

js-goupil: For the record I wasn't saying we should rename the function, just the language in the comment

packages/retail-ui-extensions/src/extension-api/navigation-api/navigation-api.ts
heltisace: Usually the `pop` function doesn't have `params` here, just a number of screens to `pop` which is something we maybe should add in the future (adding it right now might lead to some weird stuff happening when dismissing the modals).

So for now I would stick to the same API as `react-navigation` (the navigation we use for POS) and when I start investigating a better approach for navigation, we can see if it makes sense to modify this.

NathanJolly: Makes sense to me.

PR #634, opened 2022-12-02
Update development tooling

lemonmade:
This PR updates the development tooling for the UI extension repo:

- Moves off the (very old) `@sewing-kit` dependencies to Shopify's `loom` project
- Updated TypeScript, Prettier, and ESLint to their latest versions
- Updated the CI workflows to be a bit easier to grok

There should not be any actual changes to the packages' functionality from this change, so I will not be bumping any dependencies after merging this. Sorry for the large amount of changes, they are pretty much exclusively Prettier's doing :/

**Update:** actually, I think I will have to publish a version to avoid other folks unnecessarily getting these bumps in  a future version change. I also had to do some changes to the `types`/ `typeVersions` field in every package's `package.json` to match how Loom expects you to declare types for a package. I will do a patch release for each package.


packages/admin-ui-extensions/package.json
```diff
@@ -4,20 +4,33 @@
   "main": "index.js",
   "module": "index.mjs",
   "esnext": "index.esnext",
-  "types": "index.d.ts",
+  "types": "./build/ts/index.d.ts",
+  "typesVersions": {
```
vividviolet, line 8: Can you clarify why we need to declare these again here?

.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules
+.loom
 .sewing-kit
```
vividviolet, line 3: Can we eventually get rid of this?

vividviolet (approved):
Tested the admin-ui-extensions-react library - everything works. Just need to fix the conflicts on the md files

.gitignore
lemonmade: We can, maybe... I left it so people who have the repo already do not suddenly have a bunch of formally-gitignored files that now show up as added.

PR #649, opened 2022-12-09
Pin remote UI versions

lemonmade:
### Background

`@remote-ui/core@2.2.0` and `@remote-ui/react@4.6.0` make use of new core APIs for manipulating the remote root. However, the current structure of most remote-ui uses at Shopify has the host owning the version of `@remote-ui/core`, and none of them have support for the new APIs yet. Most developers will get the new version of `@remote-ui/react` because it is only a minor version, but because `@remote-ui/react` assumes it has the correct version of `@remote-ui/core`, it will break when run in a host with an older version.

Until hosts are updated to use the new version of `@remote-ui/core`, this PR pins all versions of remote-ui to the ones before the new APIs were added. This prevents consumers installing these versions of the packages from accidentally getting code depending on `@remote-ui/core@2.2.0`.

### DIFF
--- a/packages/admin-ui-extensions-react/package.json
+++ b/packages/admin-ui-extensions-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/admin-ui-extensions-react",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "main": "index.js",
   "module": "index.mjs",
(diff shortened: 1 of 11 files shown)


vividviolet (approved):
Looks good to me - double checked each package.json

jffortier (approved):

PR #655, opened 2022-12-13
Navigation interface updates

heltisace:
### Background

A decision was made to move forward with https://github.com/Shopify/pos-next-react-native/issues/17439 approach. This PR contains the final changes required to present modals easier. Changed `push` to `navigate` to accommodate the limitations of the new approach. 

### Solution

1. Added a new `Modal` interface. 
2. Changed `push` to `navigate`.

### 🎩

- ...

### Checklist


packages/retail-ui-extensions/src/extension-api/types/modal.ts
```diff
@@ -0,0 +1,10 @@
+export interface Modal {
+  /**
+   * Path should be used to determine which Screen or Navigator to return for the Modal extension point.
+   */
+  path: string;
+  /**
+   * Params can be used when presenting a modal to pass any initial parameters or attach callbacks.
+   */
+  params?: any;
```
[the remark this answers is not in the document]
heltisace: I think at some point someone asked me this question before. 

`any` makes the most sense because the best other thing we could use otherwise would be a dictionary, I suppose, but it would end up being `string` to `any` type anyway, so using a straight `any` seems like the best option for now. And using `routes` for strict typing would be overly complicated at this point.

packages/retail-ui-extensions/src/extension-api/navigation-api/navigation-api.ts
```diff
@@ -1,31 +1,26 @@
+import type {Modal} from '../types';
+
 export interface NavigationApiContent {
-  /** The name of the current screen on the stack.
-   * This property can be used to evaluate which component you want to mount.
+  /** Contains the information about the currently presented modal */
+  modal: Modal;
```
js-goupil, line 5: Should we call this something like `currentModal` instead?

js-goupil (approved):

packages/retail-ui-extensions/src/extension-api/navigation-api/navigation-api.ts
heltisace: That would make sense, yeah, renamed it.

PR #667, opened 2022-12-16
Add changesets

lemonmade:
This PR adds Changesets as a development dependency to this repo. I am not actually setting up any actions to use it yet, but the goal is to use changesets to implement our API-versioned libraries, by having it watch per-version branches like https://github.com/Shopify/ui-extensions/tree/2022-10.


kumar303 (approved):
I don't know much about Changesets other than [the auto-generated filenames](https://github.com/Shopify/cli/tree/main/.changeset) are confusing 😂 Is the CLI team happy with it?

lemonmade:
@kumar303 yeah, I have heard positive feelings about changesets from the CLI team, Web Foundations, and Polaris. I have also been relying on them for my own open source projects and they are a wonderfully lightweight solution to managing package versioning.

PR #668, opened 2022-12-16
adding 2 new extensions points for rendering components on PDP and VDP

kmdavis:
### Background

#596 

### Solution

Per discussion with @vividviolet, we're adding these new extension points to the new `@shopify/ui-extensions` package instead of the old `@shopify/admin-ui-extensions` package. At the moment, there's no special api needed, so, only the standard api is available, but I stubbed out api's for both, in case we add something later.

### 🎩

Additional changes in other repos are needed in order to tophat these, but I tophatted via a prototype using this draft pr: #654 along with corresponding changes in several other repos (exact same branch name in each)

### Checklist


byrichardpowell (approved):
:+1:

PR #670, opened 2022-12-16
Subscribable changes for Retail UI Extensions

heltisace:
### Background

We need to be able to subscribe to `Cart` and `Locale` changes from multiple `Screen` objects. Because all the objects have the same API context, we need to implement a subscriptions queue. 

Due a current `remote-ui` limitation we cannot out-of-the-box provide the API object with a `Subscribable` value that allows multiple subscriptions. Multiple hooks were added as a temporary solution. 

### Solution

- Changed the `Cart` and `Locale` APIs to rely on `RemoteSubscribable`. 
- Added `StatefulRemoteSubscribable` hooks. 

### 🎩

- ...

### Checklist


heltisace:
I have the POS code ready to go and I tested everything using `yalc` on my local environment.

PR #686, opened 2023-01-16
Add changeset actions

lemonmade:
### Background

This PR adds changeset actions to publish versioned branches containing the `@shopify/ui-extensions` package. It covers the three current version branches:

- [2022-10](https://github.com/Shopify/ui-extensions/tree/2022-10)
- [2023-01](https://github.com/Shopify/ui-extensions/tree/2023-01)
- [unstable](https://github.com/Shopify/ui-extensions/tree/unstable)

For the stable branches, the behavior is the following:

- When a PR is made targetting the branch, it will get a warning if it changes source code but does not provide a changeset
- Every merge into the branch will be tracked by changesets, and when there are unmerged changes, it will create a `changeset-release/<VERSION>` branch. That branch will contain a preview of what is ready to be released
- When we are comfortable with a release, we can merge the changeset release PR in. Changesets will track this and will deploy the new versions of the packages, using a tag matching the branch name

For the `unstable` branch, the behavior is a little different:

- When a PR is made targetting the branch, it will get a warning if it changes source code but does not provide a changeset
- When this branch contains changesets, any new changes applied to the branch will result in a ["snapshot" release](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md), tagged as `unstable`.

This PR just sets up the changeset files on `main`. I have not added the scripts that they will call (`yarn deploy`/ `yarn deploy:unstable`), because I think these would be a bit confusing while we still have Lerna on `main`. Instead, I will just add those scripts to the version branches after this PR merges.

### DIFF
--- a/.github/workflows/changesets-reminder.yml
+++ b/.github/workflows/changesets-reminder.yml
@@ -1,0 +1,36 @@
+name: Changelog Reminder
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    branches:
+      - 2022-10
+      - 2023-01
+      - unstable
+    paths:
+      - 'packages/*/src/**'
+      - 'packages/*/package.json'
+      - '!*.test.*'
+      - '!*.md'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  remind:
+    name: Changeset Reminder
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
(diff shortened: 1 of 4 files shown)


vividviolet (commented):
Can we omit the 2023-01 branch and use unstable until 2023-01 is released? I'm having trouble keeping it all in my head because we still have `main` as well. Since 2023-01 is unreleased it's technically "unstable". We can then follow the process of cloning unstable and renaming it to 2023-01 when we're ready to release.

vividviolet:
Oh sorry, I was suggesting we don't have the 3 branches. I'm fine with defining the actions ahead of time

lemonmade:
👍 Deleted the `2023-01` branch.

marvinhagemeister (approved):

PR #696, opened 2023-01-19
Updated the Icon to include all POS icon types

heltisace:
### Background

After the discussion with the UI/UX team, it was decided that POS UI Extensions should expose all the POS icon types. 

### Solution

I went to `icons.ts` in the POS repository and copied all of the keys. This got read of the two not implemented icons, and I'm not sure why they exposed in the first place: `not-stocked` and `sold-out`. Let me know if there was a reason why we had those types. In case we need them in the future, we can always expose them when they are actually implemented. 

### 🎩

- ...

### Checklist


NathanJolly (approved):
Cool.

PR #702, opened 2023-01-24
WIP: Feature/customer accounts add storage standard api

DiagoSmith:
### Background

### Solution

### 🎩

- ...

### Checklist


Safi1012 (approved):
Nice job 🎉 

The `StorageApi` matches exactly your implementation 👌

I'm not entirely sure about the `Extension` part (your linked PR doesn't include the implementation), but I'm confident that we will have the implementation soon and that API will match since it's the same from checkout.

emileber (approved):
This branch needs to be rebased onto main as the other part of our API types split refactor is now merged.

Other than that, I obviously agree to add the storage API as we've paired on this 😄

PR #705, opened 2023-01-25
adding bundles extension point

kmdavis:
### Background

Adding a new extension point `Admin::Product::Bundles::Render`

### Solution

We were going to implement this in the new `ui_extensions` spec, but due to time constraints, we're de-pivoting to implement in the old `@shopify/admin-ui-extensions` package (but the `ui_extensions` spec will still be used)

### 🎩

- ...

### Checklist


vividviolet (approved):
Looks good. Make sure that when you deploy this you make it a minor release

PR #716, opened 2023-01-30
Rename useExtensionApi to useApi

vividviolet:
Simplify `useExtensionApi` to `useApi` as this library will only be used for extensions


kumar303 (approved):
Looks good to me ✨

PR #717, opened 2023-01-30
Fix checkout surface exports

vividviolet:
While testing this library using the CLI, I got this error when trying access the esm exports

image

This PR fixes the exports for Checkout to have the correct paths.


lemonmade (approved):
Thank you!

lemonmade:
Try out the changesets flow and let me know what you think!

PR #722, opened 2023-02-01
Call root.mount after resolving the render callback

vividviolet:
### Background

### Solution

### 🎩

- ...

### Checklist


lemonmade (approved):
Thank you Trish! Would you mind checking the built outputs to see what the async function gets compiled out to? I am not sure if we have actually added these elsewhere in the library, and at least for checkout we need to make sure that some combination of the library build and/or the eventual CLI build of the extension compile this down to something that works in browsers without async/await.

PR #724, opened 2023-02-02
Added isLoading props to section and screen

js-goupil:
### Background

Pretty straight forward, but it allows the Screen and Section components to display a loading state before data is available.

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/components/Section/Section.ts
+++ b/packages/retail-ui-extensions/src/components/Section/Section.ts
@@ -7,7 +7,6 @@ export interface SectionHeaderAction {
 
 export interface SectionProps {
   title?: string;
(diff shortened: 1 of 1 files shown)


heltisace (approved):

packages/retail-ui-extensions/src/components/Section/Section.ts
```diff
@@ -7,6 +7,7 @@ export interface SectionHeaderAction {
 
 export interface SectionProps {
   title?: string;
+  isLoading?: boolean;
```
js-goupil, line 10: @heltisace is this actually a thing? I did not know we could display loading indicators like this in Sections/Cards

NathanJolly (approved):

PR #725, opened 2023-02-02
Added a boolean to trigger animation for lists on load more

js-goupil:
Resolves https://github.com/Shopify/pos-next-react-native/issues/20746

### Background

When the List component reaches the end, we have a callback `onEndReached`. Here the developer can fetch more data. We recommend that when this happens, they flip `isLoadingMore` to `true`, so that a loading indicator can happen at the bottom of the list.


heltisace:
Would it be in the middle then? If so, are you okay with that from the UX point of view? Asking because we decided to cancel https://github.com/Shopify/pos-next-react-native/issues/20572.

PR #736, opened 2023-02-10
Sync Web Pixels Schema (update CurrencyCode and CountryCode)

kencwng:
### Background

This PR is just syncing with the `web-pixel-manager` schema

- Make sure web-pixel-manager and ui-extension repos are checked out:
- In web-pixel-manager repo, run `yarn schema:all`  - this should cause the changes to be copied over to ui-extension repo automatically
- Go to ui-extension and run `yarn loom lint --fix`

### Solution

N/A

### Checklist


packages/web-pixels-extension/src/schemas/pixel-events.jtd.json
```diff
@@ -364,7 +364,7 @@
       },
       "properties": {
         "currencyCode": {
-          "ref": "CurrencyCode",
+          "type": "string",
```
kencwng, line 367: Actually, why are we exposing this JSON file in our NPM package? Should we remove this? https://socket.dev/npm/package/@shopify/web-pixels-extension/files/0.3.0/src/schemas/pixel-events.jtd.json

sorta-shopify: Hmm, yeah I think we can probably remove it. All we really need to expose are the typescript types IIRC.

PR #737, opened 2023-02-14
chore(customer-account): export capability type

DiagoSmith:
### Background

Shoutout to @Safi1012 with the eagles eyes for spotting this.
We weren't exporting this type properly, despite needing to use it in the hub (see:https://github.com/Shopify/customer-account-web/pull/1722)

### Solution

### 🎩

- ...

### Checklist


Safi1012 (approved):
Thank you for updating this! 🚀

emileber (approved):

PR #754, opened 2023-02-21
Updated POS UI Extension points namespace

pkhanal:
### Background

https://github.com/Shopify/pos-next-react-native/issues/21278

### Solution

 - Updated pos ui extension points namespace
    - `Retail::SmartGrid::Tile` -> `pos.home.title.render`
    -  `Retail::SmartGrid::Modal` -> `pos.home.modal.render`


pkhanal:
> @pkhanal are we aligned on the naming here? Let's make sure we are before we merge. Other than that looks good.

Looks like we have some feedback on the proposed naming. I will hold onto merging it for now until we agree on the naming. If we need to make additional change, i will update this PR.

heltisace:
@pkhanal I believe we are aligned on this. I'm good to merge.

heltisace (approved):

PR #759, opened 2023-02-22
Segmentation / Introduce `CustomerSegmentationTemplate` component

loic-d:
### Background
Fixes https://github.com/Shopify/core-issues/issues/51784
We are building a new rendering UI extension for Customer Segmentation. We're introducing a new component called `CustomerSegmentationTemplate` that will be used to render segmentation templates in the Admin:

Screenshot 2023-02-22 at 3 09 10 PM

### Solution

This PR adds the required interfaces for `CustomerSegmentationTemplate` in `@shopify/ui-extensions/admin` and `@shopify/ui-extensions-react/admin`. See code examples in PR.

### 🎩

🍏 CI. You can also generate a build (`yarn build`) and consume these new changes in your app extension.

### Checklist

### DIFF
--- a/packages/ui-extensions-react/src/surfaces/admin/components/CustomerSegmentationTemplate/examples/CustomerSegmentationTemplate.example.tsx
+++ b/packages/ui-extensions-react/src/surfaces/admin/components/CustomerSegmentationTemplate/examples/CustomerSegmentationTemplate.example.tsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import {
+  reactExtension,
+  CustomerSegmentationTemplate,
+} from '@shopify/ui-extensions-react/admin';
+
+function App({i18n, enabledFeatures, category}) {
+  if (category == 'reEngageCustomers') {
+    const productsPurchasedOnTagsEnabled = enabledFeatures.includes('productsPurchasedByTags');
+    const templateQuery = productsPurchasedOnTagsEnabled
+      ? 'products_purchased(tag: (product_tag)) = true'
+      : 'products_purchased(id: (product_id)) = true';
+    const templateQueryToInsert = productsPurchasedOnTagsEnabled
+      ? 'products_purchased(tag:'
+      : 'products_purchased(id:';
+  
+    return (
+      <CustomerSegmentationTemplate
+        title={i18n.translate('product.title')}
+        description={i18n.translate('product.description')}
+        icon='productsMajor'
+        templateQuery={templateQuery}
+        templateQueryToInsert={templateQueryToInsert}
+      />
+    );
+  }
+  
+  if (category == 'location') {
+    return (
+      <CustomerSegmentationTemplate
+        title={i18n.translate('location.title')}
+        description={i18n.translate('location.description')}
+        icon='locationMajor'
+        templateQuery="customer_cities CONTAINS 'US-NY-NewYorkCity'"
+      />
+    );
+  }
+
+  return null;
+}
+
+reactExtension('admin.customers.segmentation-templates.render', ({i18n, __category, __enabledFeatures}) => <App i18n={i18n} enabledFeatures={__enabledFeatures} category={__category} />);
--- a/packages/ui-extensions/src/surfaces/admin/api/customer-segmentation-template/customer-segmentation-template.ts
+++ b/packages/ui-extensions/src/surfaces/admin/api/customer-segmentation-template/customer-segmentation-template.ts
@@ -1,0 +1,30 @@
+import type {StandardApi} from '../standard/standard';
+import type {I18n} from '../../../../api';
+import type {ExtensionPoint as AnyExtensionPoint} from '../../extension-points';
+
+/* List of enabled query language features during a progressive rollout */
+type CustomerSegmentationFeature =
+  /* Allows merchants to segment on products purchased by tags. For example: products_purchased(tag: 'Red hats') = true */
+  | 'productsPurchasedByTags'
+  /* Enables count aggregates on functions. For example: shopify_email.opened(count_at_least: 5) = true */
+  | 'aggregateFilters';
+
+type TemplateCategory =
+  | 'all'
+  | 'firstTimeBuyers'
+  | 'highValueCustomers'
+  | 'reEngageCustomers'
+  | 'abandonedCheckout'
+  | 'purchaseBehaviour'
+  | 'location';
+
+export interface CustomerSegmentationTemplateApi<
+  ExtensionPoint extends AnyExtensionPoint,
+> extends StandardApi<ExtensionPoint> {
+  /* Utilities for translating content according to the current `localization` of the admin. */
+  i18n: I18n;
+  /** @private */
+  __category: TemplateCategory;
+  /** @private */
+  __enabledFeatures: CustomerSegmentationFeature[];
+}
--- a/packages/ui-extensions/src/surfaces/admin/extension-points.ts
+++ b/packages/ui-extensions/src/surfaces/admin/extension-points.ts
@@ -1,10 +1,19 @@
 import type {RenderExtension} from '../../extension';
 
-import type {AnyComponent} from './shared';
-import type {StandardApi} from './api';
+import type {AnyComponent, Components} from './shared';
+import type {StandardApi, CustomerSegmentationTemplateApi} from './api';
+import {AnyComponentBuilder} from '../../shared';
+
+type CustomerSegmentationTemplateComponent = AnyComponentBuilder<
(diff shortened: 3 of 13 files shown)


packages/ui-extensions/src/surfaces/admin/api/customer-segmentation-template/customer-segmentation-template.ts
```diff
@@ -0,0 +1,46 @@
...
+  AggregateFilters,
+}
+
+export interface CustomerSegmentationTemplateApi<
+  ExtensionPoint extends AnyExtensionPoint,
+> extends StandardApi<ExtensionPoint> {
+  /** List of enabled segmentation query language features. */
+  enabledFeatures: [CustomerSegmentationFeature];
+  /** Utilities for translating content according to the current `localization` of the admin. */
+  i18n: I18n;
```
loic-d, line 15: Should we create a shared interface with Checkout? We really only need `i18n.translate()` for now.

olavoasantos: We should definitely align in terms of API but IIRC this is currently **not** available in Admin.

loic-d: @jonathanhamel4 has a PoC for Admin

packages/ui-extensions-react/src/surfaces/admin.ts
```diff
@@ -1,4 +1,5 @@
 export {extend, extension} from '@shopify/ui-extensions/admin';
+export {CustomerSegmentationFeature} from '@shopify/ui-extensions/admin';
```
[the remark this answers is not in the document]
jonathanhamel4: We sometimes rollout gradually new filters and these new filters are often accompanied by new templates to give it more visibility. These enabledFeatures cover those, mostly. Another example is right now, we are running an experiment on the QL that spreads to the templates as well.

packages/ui-extensions/src/surfaces/admin/extension-points.ts
```diff
@@ -1,10 +1,39 @@
...
+import type {StandardApi, CustomerSegmentationTemplateApi} from './api';
+import {AnyComponentBuilder} from '../../shared';
+
+type CustomerSegmentationTemplateComponent = AnyComponentBuilder<
+  Pick<Components, 'CustomerSegmentationTemplate'>
+>;
 
 export interface ExtensionPoints {
   Playground: RenderExtension<StandardApi<'Playground'>, AnyComponent>;
+  'Admin::Customers::SegmentationTemplatesFirstTimeBuyers::Show': RenderExtension<
```
lemonmade: Are we going to wait until the conventions for extension point names is sorted out? `Show` is a term that they are using in that proposal currently, but it is different from checkout and I am not sure we have alignment on that part of the naming proposal.

packages/ui-extensions/src/surfaces/admin/api/customer-segmentation-template/customer-segmentation-template.ts
```diff
@@ -0,0 +1,46 @@
+import type {StandardApi} from '../standard/standard';
+import type {ExtensionPoint as AnyExtensionPoint} from '../../extension-points';
+
+export enum CustomerSegmentationFeature {
```
lemonmade, line 5: I think there needs to be some documentation about what these are.

loic-d: I can add more context. These are internal feature flags for our own Segmentation templates. Other 3P apps shouldn't rely on them and only use [QL public features](https://help.shopify.com/en/manual/customers/customer-segmentation/customer-segments/customer-segmentation-reference/customer-segment-shopify).

Once we open this to 3Ps, can we expose a new API interface without this property?

packages/ui-extensions/src/shared.ts
```diff
@@ -11,3 +11,76 @@ export type ComponentsBuilder<ComponentTypes> = {
 
 export type AnyComponentBuilder<ComponentTypes> =
   ComponentsBuilder<ComponentTypes>[keyof ComponentsBuilder<ComponentTypes>];
+
```
loic-d, line 14: cc/ @vividviolet I moved `I18n` here. WDYT?

packages/ui-extensions/src/surfaces/admin/extension-points.ts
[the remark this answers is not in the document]
loic-d: @lemonmade we just haven't thought about it. currently, each extension point maps to a category of templates. 

Screenshot 2023-02-23 at 4 30 54 PM

Our extension could accept a `templateCategory` parameter as part of the API and return the right templates accordingly. This would also solve the problem of the "all" category. We'd have to run the extension for each category (instead of running extensions for each extension point).

How does that sound? Would you also use a union type instead of an enum to represent the categories?

packages/ui-extensions-react/src/surfaces/admin/components/CustomerSegmentationTemplate/examples/CustomerSegmentationTemplate.example.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import {
+  reactExtension,
+  CustomerSegmentationTemplate,
+} from '@shopify/ui-extensions-react/admin';
+
+function App({i18n, enabledFeatures, category}) {
+  if (category !== 're_engage_customers') {
```
loic-d, line 8: cc/ @jonathanhamel4

packages/ui-extensions/src/surfaces/checkout/api.ts
```diff
@@ -1,3 +1,4 @@
+export type {I18n, I18nTranslate} from '../../shared';
```
loic-d, line 1: We're now re-exporting these types from the shared utility

loic-d:
@vividviolet @lemonmade this is ready for a second review:
- moved `I18n` to a shared file for both checkout and admin surfaces
- use a single extension point and provide the template category as part of the api
- use union type instead of enums
- improve examples and inline comments

the outstanding question is for `enabledFeatures`. these are really meant to be temporary (weeks) while we rollout new features. they shouldn't be used by 1P or 3P apps other than our own Segmentation app. once we address some technical limitations of our customer data platform, we shouldn't even need them anymore. are you ok with the approach I took in this PR?

packages/ui-extensions/src/shared.ts
vividviolet: Can you put it in `ui-extensions/src/api.ts` instead since it's an api?

packages/ui-extensions/src/surfaces/admin/extension-points.ts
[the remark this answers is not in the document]
loic-d: We (Segmentation app) have to handle all categories. Admin will loop over a list of categories and render the extension with the current category as `api`. If our extension doesn't handle a category, the page will be blank. We'll make sure to handle all categories. This is maybe where a non-rendering extension could have been matter because we could have returned a map of templates where the key is the category.

For other 1P (Shopify Email, etc.) and 3P apps, their templates will be grouped under an "Apps" entry point. We will provide `apps` as category but they can basically ignore it.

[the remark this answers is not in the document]
vividviolet: I think it would be very messy to go down the route of passing them a category they might ignore. It would result in the wrong template being shown for the category or blank content. It actually seems a lot better to separate these to different extension points they can specifically opt in to use.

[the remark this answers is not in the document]
loic-d: The category is really just for us. Can we have different private and public APIs?
On the Admin side we were thinking of populating categories just for our app (app ID or handle lookup).
Any other extension should go to the Apps category and the provided `templateCategory` doesn't matter.

packages/ui-extensions/src/surfaces/checkout/api/standard/standard.ts
```diff
@@ -350,79 +349,6 @@ export interface AppMetafieldEntry {
 
 export type Version = 'unstable';
 
-/**
- * This defines the i18n.translate() signature.
- */
-export interface I18nTranslate {
```
lemonmade, line 356: @kumar303 @marvinhagemeister do you know if any of the docs extraction stuff is depending on the exact file structure of our packages? This is on the shared UI extension package, so it won't impact checkout yet, but we will need to make sure it works as part of versioning.

PR #771, opened 2023-03-02
CustomerSegmentationTemplate / Introduce `dateAdded` prop

loic-d:
### Background

We currently display a "New" badge for recently introduced templates. Apps can feed us the date of when the template was added and we will display the badge if the date is within a month.

I've made the prop optional because but we could just set it in the past for all existing templates. I'm open to suggestions.

image

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentationTemplate/CustomerSegmentationTemplate.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentationTemplate/CustomerSegmentationTemplate.ts
@@ -32,14 +32,16 @@ export interface CustomerSegmentationTemplateProps {
   title: string;
   /* Localized description of the template. */
   description: string;
-  /* Icon identifier for the template. This property is ignored for non-1P Segmentation templates as we fallback to the app icon */
+  /* Icon identifier for the template. This property is ignored for non-1P Segmentation templates as we fallback to the app icon. */
   icon?: Source;
-  /* ShopifyQL code snippet to render in the template with syntax highlighting */
+  /* ShopifyQL code snippet to render in the template with syntax highlighting. */
   templateQuery: string;
   /* ShopifyQL code snippet to insert in the segment editor. If missing, `templateQuery` will be used. */
   templateQueryToInsert?: string;
-  /* List of customer standard metafield used in the template's query */
+  /* List of customer standard metafield used in the template's query. */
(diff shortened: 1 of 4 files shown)


packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentationTemplate/CustomerSegmentationTemplate.ts
```diff
@@ -32,14 +32,16 @@ export interface CustomerSegmentationTemplateProps {
...
-  /* ShopifyQL code snippet to render in the template with syntax highlighting */
+  /* ShopifyQL code snippet to render in the template with syntax highlighting. */
   templateQuery: string;
   /* ShopifyQL code snippet to insert in the segment editor. If missing, `templateQuery` will be used. */
   templateQueryToInsert?: string;
-  /* List of customer standard metafield used in the template's query */
+  /* List of customer standard metafield used in the template's query. */
   standardMetafieldDependencies?: CustomerStandardMetafieldDependency[];
+  /* Date when the template was first introduced. A "New" badge will be rendered for recently introduced templates. */
+  dateAdded?: Date;
```
loic-d, line 44: Thoughts on this being optional/required?

lemonmade (commented):
This API seems a little odd, shouldn't we be able to determine this ourself (since we know when extensions are created)?

loic-d:
@lemonmade can we retrieve that information using the Admin GraphQL API? I don't see any relevant attribute on an `Extension`

loic-d:
@lemonmade we need more granularity actually. For example:
- Extension adds templates A, B, C. They are flagged as new
- Extension adds templates D, E one month later. Only D, E should be flagged as new.

packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentationTemplate/CustomerSegmentationTemplate.ts
carolopolo: optional makes sense to me 👍🏼

lemonmade:
@loic-d could these be separate extensions instead?

loic-d:
@lemonmade I don't think this is a scalable solution. We have almost 20 first party templates at the moment and this list will continue to grow. It will be harder to maintain and I think it will lead to duplicate boilerplate code being shipped as part of each extension's bundle 🤔

carolopolo (approved):
LGTM. i think this makes sense given our needs.

PR #774, opened 2023-03-02
[CustomerSegmentationTemplate] Update `description` prop type

carolopolo:
### Background

- Segmentation templates in web currently accept an array of strings for the `description`. Some templates contain several sentences, and others contain 1 sentence, so this PR allows a union type to facilitate both use cases in a backwards compatible way.

### Solution

Please refer to background section

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentationTemplate/CustomerSegmentationTemplate.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentationTemplate/CustomerSegmentationTemplate.ts
@@ -30,9 +30,9 @@ type CustomerStandardMetafieldDependency = 'facts.birth_date';
 export interface CustomerSegmentationTemplateProps {
   /* Localized title of the template. */
   title: string;
-  /* Localized description of the template. */
-  description: string;
-  /* Icon identifier for the template. This property is ignored for non-1P Segmentation templates as we fallback to the app icon. */
+  /* Localized description(s) of the template. */
+  description: string | string[];
+  /* Icon identifier for the template. This property is ignored for non-1P Segmentation templates as we fallback to the app icon */
   icon?: Source;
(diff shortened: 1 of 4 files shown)


packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentationTemplate/CustomerSegmentationTemplate.ts
```diff
@@ -30,8 +30,8 @@ type CustomerStandardMetafieldDependency = 'facts.birth_date';
 export interface CustomerSegmentationTemplateProps {
   /* Localized title of the template. */
   title: string;
-  /* Localized description of the template. */
-  description: string;
+  /* Localized description(s) of the template. */
+  description: string | string[];
```
loic-d, line 34: Should it just be a `string[]`? This way we don't have to check if you're getting a string or array in the component's implementation?

loic-d (approved):
LGTM!

packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentationTemplate/CustomerSegmentationTemplate.ts
carolopolo: sure! no strong opinion/i'm indifferent - @jonathanhamel4 brought up the advantage of a union type: it would mean it's backwards compatible. but i also think it is nice to keep it simple :)

loic-d: let's go with the union type (you both suggested it)

jonathanhamel4 (approved):
lgtm. Agreed we could rename descriptionX to paragraphX, makes it a bit clearer that there will be a line break between them. Else, lgtm.

PR #775, opened 2023-03-03
Add toast API to ui extensions

NathanJolly:
### Background
https://github.com/Shopify/pos-next-react-native/issues/19327
Adds the toast API.

### Solution
Add it in the pattern of other API's

### 🎩 
I've tested this API on POS by hand and it looks good. Is this all I need to change for this part to work? Did I miss any indexes or places? Nothing to add to the `react` package since there's no React-specific stuff going on here, right?

### DIFF
--- a/packages/retail-ui-extensions/src/extension-api/toast-api/toast-api.ts
+++ b/packages/retail-ui-extensions/src/extension-api/toast-api/toast-api.ts
@@ -1,0 +1,15 @@
+export interface ToastApiContent {
+  /**
+   * Show a toast.
+   * @param content The text content to display.
+   * @param duration The duration of the toast in ms. Defaults to 5000.
+   */
+  show: (content: string, duration?: number) => void;
+}
+
+/**
+ * Show a toast.
+ */
+export interface ToastApi {
+  toast: ToastApiContent;
+}
--- a/packages/retail-ui-extensions/src/extension-api/index.ts
+++ b/packages/retail-ui-extensions/src/extension-api/index.ts
@@ -6,6 +6,7 @@ export type {SmartGridApi, SmartGridApiContent} from './smartgrid-api';
 export type {StandardApi} from './standard-api';
 
 export type {SessionApiContent, SessionApi} from './session-api';
+export type {ToastApiContent, ToastApi} from './toast-api';
 
 export type {
   Address,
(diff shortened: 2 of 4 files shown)


packages/retail-ui-extensions/src/extension-api/toast-api/toast-api.ts
```diff
@@ -0,0 +1,15 @@
+export interface ToastApiContent {
+  /**
+   * Show a toast.
+   * @param content The text content to display.
+   * @param duration The duration of the toast in ms. Defaults to 5000.
+   */
+  show: (content: string, duration?: number) => void;
```
vctrchu, line 7: Do you think there is value giving partners to ability to `dismiss` their toast?

NathanJolly: I didn't really think so, since guidance on Toasts suggest to only use them for less important things. Between duration and all toasts having an "X" button, I was hard pressed to think of a case where we'd need to `dismiss`. Your thoughts?

vctrchu (approved):
That sounds good. No strong opinions, just thought I would throw that out. ✅

packages/retail-ui-extensions/src/extension-api/toast-api/toast-api.ts
```diff
@@ -0,0 +1,15 @@
+export interface ToastApiContent {
```
olavoasantos, line 1: I see this API has a different shape than the one used in [admin-ui-extensions](https://shopify.dev/docs/api/admin-extensions/api/toastapi) and other surfaces.

To reduce fragmentation, could we use something closer to Admin?

```ts
export interface ShowToastOptions {
  duration?: number;
}

export interface ToastApiContent {
  show: (content: string, options?: ShowToastOptions) => void;
}
```

packages/retail-ui-extensions/src/extension-api/toast-api/toast-api.ts
developit, line 7: Using an optional positional argument for `duration` will make it very difficult to add future arguments to `toast.show()`. If this is switched to using named parameters / a description object, that will allow for future parameters without changing the method's arity.

```suggestion
  show: ({ content: string, duration?: number }) => void;
```

developit: actually - @olavoasantos's proposed API is preferable over this one.

PR #781, opened 2023-03-08
Removed SearchBar component, added it as a Screen prop

heltisace:
partially_resolves https://github.com/Shopify/pos-next-react-native/issues/20041

### Figma

The smaller SearchBar here is the desired one for when the `SearchBar` is shown on top of the screen:
https://www.figma.com/file/A0B96slYSNa9fw3UrMWdt2/Retail---Components?node-id=7446%3A453&t=CXgRerv7KeR7c8OB-1

### Background

When we reviewed the current UI Extensions components with the UI and UX team, we received the feedback that the SearchBar on top of the screen should be an `InlineSearchBar`. 

### Solution

As at the moment this is the main use case we want to support, I removed the `SearchBar` component, added the props to the `Screen`, and will update the implementation on POS to render an `InlineSearchBar` instead. 

### 🎩

- ...

### Checklist


packages/retail-ui-extensions/src/components/Screen/Screen.ts
```diff
@@ -1,8 +1,20 @@
 import {createRemoteComponent} from '@remote-ui/core';
 
+export interface ScreenSearchBarProps {
+  initialValue?: string;
+  onTextChange?: (value: string) => void;
+  onSearch: (value: string) => void;
+  onFocus?: () => void;
+  visible?: boolean;
```
heltisace, line 8: This is a new prop I added. Let me know if you think it's useful.

js-goupil: Can we double check the naming convention we use for these booleans with core-app? ie `isVisible` vs `visible`. Just so we get it right the first time

packages/retail-ui-extensions/src/components/Screen/Screen.ts
```diff
@@ -1,8 +1,20 @@
...
+  editable?: boolean;
+  placeholder?: string;
+  showBarcodeIcon?: boolean;
+}
+
 export interface ScreenProps {
   name: string;
   isLoading?: boolean;
+  searchBar?: ScreenSearchBarProps;
```
js-goupil, line 17: I'm guessing undefined here will be the default, and not render a search bar in this case?

js-goupil (approved):
Looks gucci

packages/retail-ui-extensions/src/components/Screen/Screen.ts
agorovyi: @heltisace could you please clarify what this `visible` prop will do? Is this the prop that we discussed that should make the `SearchBar` smaller?

agorovyi (approved):
Looks good!

PR #789, opened 2023-03-10
New navigation Modal for Retail UI Extensions

js-goupil:
### Background

This PR simplifies our approach to presenting a separate `Modal` on POS. Partners will now be able to wrap a `Navigator`, or `Screen` in a `Modal`, which will modally present a new stack or single screen. This is a big improvement over the current approach, as now the modals can all live in the same instance of a UI Extension, as opposed to POS presenting a new modal running a new context with the same extension point.

This means that all modals will render immediately (much faster), and sharing context and state between the two modals is all done within the extension, which is much easier to manage for the partner.

### Solution

This would result in an app stack looking like this:

```tsx
const SmartGridModal = () => {
  const api = useExtensionApi();

  return (
    <Navigator>
      <Modal name="HomeModal">
        <Navigator>
          <Screen name="Home">
            <Button title="Go to Product List!" onPress={() => api.navigation.navigate('ProductsList')} />
            <Button title="Go to Settings Details!" onPress={() => api.navigation.navigate('SettingsModal')} />
          </Screen>
          <Screen name="ProductsList">
            <Text>Products List!</Text>
          </Screen>
        </Navigator>
      </Modal>
      <Modal name="SettingsModal">
        <Navigator>
          <Screen name="SettingsRoot">
            <Button title="Go to Settings Details!" onPress={() => api.navigation.navigate('SettingsDetails')} />
          </Screen>
          <Screen name="SettingsDetails">
            <Text>Settings details!</Text>
          </Screen>
        </Navigator>
      </Modal>
    </Navigator>
  );
}
```

In this example, there are two modals, `HomeModal` and `SettingsModal`. Both wrap a navigator with some screens. Since HomeModal is the first in the parent Navigator, it's the root screen that will show up when the extension point opens. From the Home screen, you can push the product list onto its stack, or you can modally present the SettingsModal. The Settings Modal also has to screens, where you can navigate between them within this stack.

### 🎩

- ...

### Checklist


packages/retail-ui-extensions/src/extension-api/standard-api.ts
```diff
@@ -1,9 +1,11 @@
 import type {CartApi} from './cart-api';
 import type {LocaleApi} from './locale-api';
 import type {SessionApi} from './session-api';
+import {ToastApi} from './toast-api';
 
 export type StandardApi<T> = {[key: string]: any} & {
   extensionPoint: T;
 } & LocaleApi &
   CartApi &
+  ToastApi &
```
js-goupil, line 10: @NathanJolly I think we forgot to include the ToastApi in the api object interface. We probably want the ToastApi to be available on both extension points right?

packages/retail-ui-extensions/src/extension-api/navigation-api/navigation-api.ts
```diff
@@ -1,20 +1,10 @@
-import type {Modal} from '../types';
-
 export interface NavigationApiContent {
-  /** Contains the information about the currently presented modal */
-  currentModal: Modal;
-
-  /** Presents a new modal on top of the current one.
-   * @param modal this interface contains the information about the path and any initial parameters.
-   */
-  presentModal(modal: Modal): void;
```
js-goupil, line 10: We no longer need this. The `navigate` function will go to the correct modal if you pass in the modal's name. This matches more closely what react-navigation does.

packages/retail-ui-extensions/src/extension-api/smartgrid-api/smartgrid-api.ts
```diff
@@ -1,10 +1,7 @@
-import type {Modal} from '../types';
-
 export interface SmartGridApiContent {
-  /** Presents a new modal on top of the Smart Grid.
-   * @param modal this interface contains the information about the path and any initial parameters.
+  /** Presents the `pos.home.modal.render` extension point on top of the Smart Grid.
    */
-  presentModal(modal: Modal): void;
+  presentModal(): void;
```
heltisace, line 4: Two notes about what this new approach misses compared to the old one: 

1. Sometimes the user might want to start with a different screen depending on the tile state. Perhaps we should have a `destination` here.
2. Sending `params` here could also be useful in case a `Tile` needs to share some state with the `Modal`.

heltisace (commented):
The changes make sense. Left a comment regarding showing modals from tiles.

packages/retail-ui-extensions/src/extension-api/smartgrid-api/smartgrid-api.ts
js-goupil: Hmmmm yeah fair enough

packages/retail-ui-extensions/src/extension-api/standard-api.ts
NathanJolly: Thank you! Correct: Both extension points.

NathanJolly (approved):

vctrchu (approved):
LGTM

packages/retail-ui-extensions/src/extension-api/smartgrid-api/smartgrid-api.ts
js-goupil: @heltisace `navigationApi`'s `navigate` method takes in the same parameters as well. It takes a screen/modal name and a params. IMO we should either share the destination type with it OR, we don't create a type, we just have a `destinationName: string, params: any`. I'm leaning more on just having regular primitive function parameters, rather than a typed object. It's just two parameters. What do you think?

henrytao-me (approved):
LGTM

heltisace (approved):
Nice, let's ship it

PR #797, opened 2023-03-14
Revert API type changes

js-goupil:
### Background

Turns out the previous changes broke all the API typing. The api auto completes as any in the UI Extension. 

### Solution

Reverts the changes that broke the typing

### 🎩

- ...

### Checklist


heltisace (approved):
We checked with yalc and this version works

PR #806, opened 2023-03-17
[segmentation] moving category from api to props

jonathanhamel4:
### Background

Moves the category from the api to the props. This will make web control the rendering of the `all` while allowing developers to add a template to a single category. When the category is missing, Web will render the template in the `All` or the respective `app` section only. 

We currently have to update the api to show the new category. We could also leverage the context to control the rendering. Open to suggestions.

### 🎩

Nothing to tophat.

### Checklist

### DIFF
--- a/.changeset/hot-jobs-return.md
+++ b/.changeset/hot-jobs-return.md
@@ -1,0 +1,6 @@
+---
+'@shopify/ui-extensions': patch
+'@shopify/ui-extensions-react': patch
+---
+
+Moving the Segmentation TemplateCategory from the api to the CustomerSegmentationTemplate component props.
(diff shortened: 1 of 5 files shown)


.changeset/hot-jobs-return.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/ui-extensions': patch
+'@shopify/ui-extensions-react': patch
```
jonathanhamel4, line 3: nothing is used yet, I probably wouldn't worry too much about this being a patch?

loic-d (approved):
LGTM 👍

carolopolo (approved):
LGTM!

PR #814, opened 2023-03-20
Publish

h74zhou:
- @shopify/web-pixels-extension@0.4.1


sorta-shopify (approved):
👍 Tags created, looks good to me 👍

PR #818, opened 2023-03-23
feat: expose modal component to customer account ui extensions

DiagoSmith:
### Background

Allows customer-account ui extensions to make use of the Modal component
[Customer Account Hub PR here](https://github.com/Shopify/customer-account-web/pull/1900) (tophatting instructions also there). 

### Checklist


DiagoSmith:
> Don't we need them in the vanilla package as well?
> 
> https://github.com/Shopify/ui-extensions/blob/main/packages/customer-account-ui-extensions/src/components/shared-components.ts#L23

Good catch, updated

robin-drexler (approved):

PR #823, opened 2023-03-28
Allow React extensions to return an element asynchronously

lemonmade:
### Background

This PR expands the ability for an extension to return its initial UI asynchronously by bringing it to the React library. Now, when a function registers an async function that returns a React element, we will wait for that promise to resolve before mounting the resulting element. The example below shows how a developer could use this to indicate that their extension is not ready to render, until a direct API access call is resolved:

```tsx
import {reactExtension, Text} from '@shopify/ui-extensions-react/checkout';

export default reactExtension('Checkout::Dynamic::Render', async ({query}) => {
  const data = await query(`query Recommendations { productRecommendations { id title } }`);
  return <Extension initialData={data} />;
});

function Extension({initialData}) {
  return <Text>{JSON.stringify(initialData)}</Text>
}
```

### Checklist

### DIFF
--- a/.changeset/green-colts-attend.md
+++ b/.changeset/green-colts-attend.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions-react': patch
+---
+
+Allow React extensions to return a React element asynchronously


packages/ui-extensions-react/src/surfaces/admin/render.tsx
```diff
@@ -25,18 +25,22 @@ import {ExtensionApiContext} from './context';
...
+  ) => ReactElement<any> | Promise<ReactElement<any>>,
 ): ExtensionPoints[ExtensionPoint] {
   // TypeScript can’t infer the type of the callback because it’s a big union
   // type. To get around it, we’ll just fake like we are rendering the
   // Playground extension, since all render extensions have the same general
   // shape (`RenderExtension`).
-  return extension<'Playground'>(target as any, (root, api) => {
-    return new Promise((resolve, reject) => {
+  return extension<'Playground'>(target as any, async (root, api) => {
+    const element = await render(api as ApiForRenderExtension<ExtensionPoint>);
```
robin-drexler, line 37: Shouldn't that be in the `try/catch`? If `render` throws we wouldn't reach the console.error, would we?

lemonmade: Good call, I will move the try/catch around the whole function body.

lemonmade: err actually @marvinhagemeister do you know if the `console.error` is even needed anymore? I feel like we should be doing that in the sandbox instead.

PR #825, opened 2023-03-28
Added new props to ListItem

js-goupil:
### Background

This PR updates the props available, according to the new [List designs and row permutations](https://www.figma.com/file/htKrn9jejIfeyXrjQpHNOn/List-Component-MVP?node-id=7-103977&t=6eQpR68Z9iFf7qHR-0).

### Solution

We are no longer going to be supporting subtitleColor, and instead the badge component will be breathing new life into the rows. We're also adding support for additional subtitles.

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/components/List/List.ts
+++ b/packages/retail-ui-extensions/src/components/List/List.ts
@@ -54,7 +54,7 @@ export interface ListProps {
    */
   title?: string;
   /**
-   * The array of `ListItem` which will be converted into rows for this list.
+   * The array of `ListRow` which will be converted into rows for this list.
    */
   data: ListRow[];
   /**


NathanJolly (approved):

packages/retail-ui-extensions/src/components/List/List.ts
```diff
@@ -1,22 +1,76 @@
...
+  label: string;
+  /**
+   * The subtitles to display beneath the main label. Up to 3 subtitles can be displayed.
+   */
+  subtitle?: readonly [string, string?, string?];
+  /**
+   * A colored badge that is displayed underneath the `title` and `subtitles`.
+   */
+  badge: TagProps;
+  image?: {
```
vctrchu, line 17: Did you exclude a comment for image here on purpose?

packages/retail-ui-extensions/src/components/List/List.ts
```diff
@@ -1,22 +1,76 @@
...
 export interface ListProps {
-  data: ListItem[];
-  showDisclosureIndicator: boolean;
-  isLoadingMore?: boolean;
+  /**
+   * A large display title at the top of the `List`.
+   */
   title?: string;
+  /**
+   * The array of `ListItem` which will be converted into rows for this list.
```
agorovyi, line 57: Do we still call this `ListItem`? Should this be `ListRow`?

```suggestion
   * The array of `ListRow` which will be converted into rows for this list.
```

packages/retail-ui-extensions/src/components/List/List.ts
```diff
@@ -1,22 +1,76 @@
...
 }
 
 export interface ListProps {
-  data: ListItem[];
-  showDisclosureIndicator: boolean;
-  isLoadingMore?: boolean;
+  /**
+   * A large display title at the top of the `List`.
+   */
   title?: string;
```
agorovyi, line 55: Just a question. Is this `title` used in section header? Do we want to also add an action with an `onPress` that appears in the right side of the section header?

js-goupil: No this is more for the large title at the top of the infinite List. What you're referring to we do currently have in our Section component, it's a section header action.

PR #834, opened 2023-03-30
Made disclosure indicator on right side optional

js-goupil:
### Background

Adds the disclosure indicator to the list row right side. Optional, defaults to false.

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/components/List/List.ts
+++ b/packages/retail-ui-extensions/src/components/List/List.ts
@@ -27,6 +27,11 @@ export interface ListRowRightSide {
    * The main label that will be displayed on the right side of the row.
    */
   label?: string;
+  /**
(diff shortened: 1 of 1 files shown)


packages/retail-ui-extensions/src/components/List/List.ts
```diff
@@ -27,6 +27,11 @@ export interface ListRowRightSide {
    * The main label that will be displayed on the right side of the row.
    */
   label?: string;
+  /**
+   * Show a chevron. Set this to true if pressing on the row navigates to another screen.
+   * @defaultValue `false`
+   */
+  showDisclosureIndicator?: boolean;
```
heltisace, line 34: Maybe something like this is better? What's a better name for this? 

```suggestion
  chevron?: boolean;
```

heltisace (approved):

PR #838, opened 2023-03-30
include address1, address2, zip in mailingaddress

eric-blue:
### Background

https://github.com/Shopify/ce-customer-behaviour/issues/1719

### Solution

### 🎩

https://screenshot.click/30-07-o51xz-ljydp.mp4

### Checklist


eric-blue:
what docs need to be updated here?

PR #840, opened 2023-03-31
Bump package from 0.4.2 > 0.4.3

eric-blue:
- @shopify/web-pixels-extension@0.4.3

### Background

https://github.com/Shopify/ui-extensions/pull/838

### Solution

### 🎩

- ...

### Checklist


sorta-shopify (approved):
Found the tag, looks good to me 👍 

> image

PR #856, opened 2023-04-10
Add products to collection viewed

eric-blue:
### Background

part of https://github.com/Shopify/ce-customer-behaviour/issues/2051

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/web-pixels-extension/src/types/index.ts
+++ b/packages/web-pixels-extension/src/types/index.ts
@@ -2,6 +2,7 @@ export type {ExtensionApi} from './ExtensionApi';
 export type {RegisterInit, InitData} from './RegisterInit';
 export type {PublicApi} from './PublicApi';
 export type {
+  Attribute,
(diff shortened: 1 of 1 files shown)


sorta-shopify (approved):
👍

packages/web-pixels-extension/src/types/PixelEvents/index.ts
```diff
@@ -581,6 +581,7 @@ export interface Collection {
    * A globally unique identifier.
    */
   id: string;
+  productVariants: ProductVariant[];
```
h74zhou, line 584: If you run `yarn:schem all` does this get removed? 

I don't believe you change it here. This file should be generated.

eric-blue: this was generated by my edits to Collection running `yarn:schema all`

PR #858, opened 2023-04-11
Renamed POS Tag to Badge

js-goupil:
### Background

This applies the name in the retail for polaris guidelines, but also matches with what admin is using as the name.

### 🎩

- ...

### Checklist


NathanJolly (approved):
Is there a ticket for the doc update, too?

PR #863, opened 2023-04-12
Retail/pos rename

NathanJolly:
### Background
Partly resolves https://github.com/Shopify/pos-next-react-native/issues/23232

### Solution
Basically copy-paste `retail-ui-extensions` to `pos-ui-extensions` and `retail-ui-extensions-react` to `pos-ui-extensions-react`, then update supporting files.

NB: I ran into a typescript problem while building for the first time. Since `-react` depends on `non-react`, I had to add the `non-react` package first, then `yarn build`, then add the `-react` package, then `yarn build` again. Is this expected? Will it cause problems with other folks first builds?

### 🎩
`yarn build` and verify that the new `pos-ui-extensions(-react)` packages build.

### Checklist


.github/CODEOWNERS
```diff
@@ -27,6 +27,8 @@
 /packages/post-purchase-ui-extensions-react/      @Shopify/checkout-extensibility
 /packages/retail-ui-extensions/			  @Shopify/retail-succeed-extensibility
 /packages/retail-ui-extensions-react/		  @Shopify/retail-succeed-extensibility
+/packages/pos-ui-extensions/			  @Shopify/retail-succeed-extensibility
```
heltisace, line 30: Let's update these to @Shopify/retail-succeed-build. Same goes for `retail-ui-extensions`.

heltisace (approved):
LGTM

heltisace:
If you merge this, don't deploy it yet. We might need to include a few minor fixes before we cut the GA version. I got a few reports from partners.

PR #880, opened 2023-04-19
add dynamic customer account render extension point

robin-drexler:
### Background

Uses new naming scheme. We'll adjust the other extension points soon as well.

https://www.figma.com/file/rEv6iMXEo2uhtZOndFnG4c/%F0%9F%91%A4-Self-Serve-%7C-Customer-Account-Hub?node-id=6158-177766&t=BHgbTZtgTXG49tAA-0

### Solution

### 🎩

- ...

### Checklist


Safi1012 (approved):
🚀🚀🚀

PR #882, opened 2023-04-19
Creates a Scanner API in POS UI Extensions

js-goupil:
### Background

This is the second part of the ticket to create the package side interfaces for the new POS Scanner API.

### Solution

Pretty simple, as this is just creating an API wrapper for the scanner API.

The one thing I'd like to point out is that currently, our tech design indicates that the data from the scan event, as well as the supported sources are both wrapped in the same subscribable. I wonder if instead we should have two subscribables. It seems like a strange thing to wrap two not closely related things in the same subscription. Thoughts?

### Checklist


packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
```diff
@@ -0,0 +1,24 @@
+import type {RemoteSubscribable} from '@remote-ui/async-subscription';
+
+/** The scanner source the POS device supports */
+export type SupportedSource = 'camera' | 'external' | 'embedded';
+
+export interface ScannerSubscriptionResult {
+  /** The string data from the last scanner event received */
+  data?: string;
+  /** An array containing all supported scanner types on the POS device. */
+  supportedSources: SupportedSource[];
```
heltisace, line 10: This one should be the source of the specific scan. So where the data comes from.

packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
```diff
@@ -0,0 +1,24 @@
+import type {RemoteSubscribable} from '@remote-ui/async-subscription';
+
+/** The scanner source the POS device supports */
+export type SupportedSource = 'camera' | 'external' | 'embedded';
```
heltisace, line 4: Should the partners even know about the `embedded` source? Or should we just create a word that works for both `external` and `embedded`? 

If we do want to have three types, perhaps we should rename to embedded to `pos-go` for 3P because embedded sounds ambiguous. 

Curious to hear what everyone else thinks about this.

packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
heltisace: We can add this to the actual API object, this array shouldn't mutate, right? Or will it actually show if an external scanner is connected? If it does, then perhaps we should just have a separate subscribable for this?

packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface ScannerSubscriptionResult {
+  /** The string data from the last scanner event received */
+  data?: string;
```
NathanJolly, line 8: What does a case look like where `data` would be undefined? Wouldn't a successful scan always provide something?

heltisace: Agreed, this shouldn't be undefined.

packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
vctrchu: supported sources is an array that returns the available scanners that device has available. It is not related to the type of scanner data that gets returned. Hmm maybe bunching this array into the same hook is causing confusion and we should have a separate hook for it.

packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
NathanJolly: Totally, now I understand the intent of the array!

vctrchu: @heltisace Although external and embedded scanners use the same listener, the availability for each type of scanner needs to be distinct. POS Go is technically the only current device that supports embedded scanning however that could change in the future.

packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
NathanJolly: I agree. It's fine here. I think ConnectedSources or AvailableSources might be a less confusing term than SupportedSources, but it's not a strong opinion.

packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
```diff
@@ -0,0 +1,24 @@
+import type {RemoteSubscribable} from '@remote-ui/async-subscription';
+
+/** The scanner source the POS device supports */
```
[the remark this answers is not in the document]
js-goupil: Screenshot 2023-04-20 at 9 55 08 AM

More or less for the inline docs. I know the code is pretty self documenting, but it's always good to make sure everyone knows what's going on in case it's not clear 😄

packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
```diff
@@ -0,0 +1,30 @@
+import type {RemoteSubscribable} from '@remote-ui/async-subscription';
+
+/** The scanner source the POS device supports. */
+export type ScannerSource = 'camera' | 'external' | 'embedded';
+
+export interface ScannerSubscriptionResult {
+  /** The string data from the last scanner event received. */
+  data?: string;
+  /** The scanning source from which the scan event came. */
+  scannerSource: ScannerSource;
```
vctrchu, line 10:
```suggestion
  source: ScannerSource;
```
How I see it, is its implied that its a scanner source if its inside the ScannerSubscriptionResult. If we follow this flow, why isn't data named scannerData?

packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export interface ScannerApiContent {
+  /** Creates a subscription to scan events
+   * Provides an initial value and a callback to subsribe to value changes. Currently supports only one subscription.
+   * You can utilize `makeStatefulSubscribable` on a `RemoteSubscribable` to implement multiple subscriptions.
+   * Using `makeStatefulSubscribable` or the corresponding hooks counts as a subscription.
+   */
+  scannerDataSubscribable: RemoteSubscribable<
+    ScannerSubscriptionResult | undefined
```
js-goupil, line 20: @heltisace @vctrchu @NathanJolly Since the entire object is now the result of a scan, what do you think about this instead? The subscribable returns undefined if there hasnt been a scan yet. data and source are now both mandatory values

heltisace: I'm good with that but I have a question about the name. The Result contains both `data` and `source`. Can we name this variable and the hooks somehow better?

vctrchu (approved):

packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
```diff
@@ -0,0 +1,32 @@
+import type {RemoteSubscribable} from '@remote-ui/async-subscription';
+
+/** The scanner source the POS device supports. */
+export type ScannerSource = 'camera' | 'external' | 'embedded';
+
+export interface ScannerSubscriptionResult {
+  /** The string data from the last scanner event received. */
+  data: string;
```
vctrchu, line 8: i think data and source might have to be undefined, im getting problems with the extension running if i destructure these from the useScannerDataSubscription..

vctrchu: Yeah it marks the whole api as undefined causing the extension to not load. You will have to do something like this   `const { data, source } = useScannerDataSubscription() || {};` which isn't ideal for the developer experience.

heltisace: We paired on this, what Victor showed makes sense. I agree that making those optional is not ideal but creates a better development experience.

packages/retail-ui-extensions/src/extension-api/scanner-api/scanner-api.ts
vctrchu: Fixed it here https://github.com/Shopify/ui-extensions/pull/888

PR #883, opened 2023-04-19
include pixels schema json to be consumed in web

eric-blue:
### Background

following https://github.com/Shopify/web-pixels-manager/pull/330, we want to reintroduce the `pixel-events.jtd.schema` so it [can be consumed](https://github.com/Shopify/ce-customer-behaviour/issues/2479) in the forthcoming code editor autocomplete [feature](https://github.com/Shopify/ce-customer-behaviour/issues/1536) in `/web`

### Solution

exposes the schema publicly, which is already available in our dev docs so no risk there

### 🎩

...

### Checklist


sorta-shopify (approved):
👍

PR #909, opened 2023-05-01
chore(customer-accounts): deprecate pageblock & section

DiagoSmith:
### Background

Part of https://github.com/Shopify/core-issues/issues/49020 

TLDR: PageBlock and Section should not be exposed to 3p or 1p extensibility partners. 

### Solution

### 🎩

- ...

### Checklist


robin-drexler (approved):
🪦

peterleilei86 (approved):
⚰️

PR #950, opened 2023-05-22
Update Navigator to allow initial screen

NathanJolly:
### Background
Part of https://github.com/Shopify/pos-next-react-native/issues/24123

### Solution

Add an `initialScreenName` prop to the Navigator component.

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/components/Navigator/Navigator.ts
+++ b/packages/retail-ui-extensions/src/components/Navigator/Navigator.ts
@@ -1,3 +1,12 @@
 import {createRemoteComponent} from '@remote-ui/core';
 
-export const Navigator = createRemoteComponent<'Navigator'>('Navigator');
+/**
+ * @property `initialScreenName` sets the initial Screen by its `name` property.
+ */
+export interface NavigatorProps {
+  initialScreenName?: string;
+}
+
+export const Navigator = createRemoteComponent<'Navigator', NavigatorProps>(
+  'Navigator',
+);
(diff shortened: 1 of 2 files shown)


heltisace:
If we want to restrict the Partners to using the `Navigator` as a place where we set what screens to use, they should be able to just pass the `Screens` in different order depending on the status of the cart. So this change would primarily provide only utility value and not unlock new behaviors.

What do you think the downside be of adding `params` as a value when showing a `Modal`? I know `Tile` doesn't have many inputs, but they still could pre-fetch some data from their backend they could pass through using `params`. And later, when we unlock new extension points, `params` could prove to be even more useful.

heltisace:
We could also add both this change because it does follow https://reactnavigation.org/docs/stack-navigator/#initialroutename like you mentioned, and the `params`. I'm just trying to find a way to restrict the Partners the least in how they develop.

NathanJolly:
@heltisace 

> If we want to restrict the Partners to using the `Navigator` as a place where we set what screens to use, they should be able to just pass the `Screens` in different order depending on the status of the cart. So this change would primarily provide only utility value and not unlock new behaviors.

Correct. This solution is far more explicit and obvious.

> What do you think the downside be of adding `params` as a value when showing a `Modal`? I know `Tile` doesn't have many inputs, but they still could pre-fetch some data from their backend they could pass through using `params`. And later, when we unlock new extension points, `params` could prove to be even more useful.

**Downsides:** 
- _Solution in search of a problem._ The problem is that the partner wants an explicit way to tell the navigator what screen to start on. This approach solves that problem.
- _Pre-fetching should be discouraged._ Encouraging partners to use precious device resources while their modal is not loaded is a poor idea. They shouldn't be taking more resources than needed while their extension is not actively being used.

PR #954, opened 2023-05-24
[Customer Account] Remove unused Card props

lsit:
### Background

Part of https://github.com/Shopify/core-issues/issues/55239
Related to https://github.com/Shopify/customer-account-web/pull/2258

Removes the Card props `fillHeight` and `includeDividers` from `customer-account-ui-extensions`. These props are no longer available in `customer-account-web`. They aren't used by the `buyer-returns` or `buyer-subscriptions` extensions so it's safe to remove them.

### Solution

### 🎩

See https://github.com/Shopify/customer-account-web/pull/2258 for details

### Checklist

### DIFF
--- a/packages/customer-account-ui-extensions/src/components/Card/Card.ts
+++ b/packages/customer-account-ui-extensions/src/components/Card/Card.ts
@@ -7,14 +7,6 @@ export interface CardProps {
    */
   background?: Background;
   /**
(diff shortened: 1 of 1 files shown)


emileber (approved):
Removing props would be considered breaking changes, but since this library is not supposed to be used yet (other then our own teams), I guess it doesn't really matter.

So LGTM 🚀

PR #958, opened 2023-05-24
Add badge and toggleSwitch to List

arigny:
### Background

Part of [24526](https://github.com/Shopify/pos-next-react-native/issues/24526)

This PR adds the `badge` and `toggleSwitch` properties to the `ListRowProps` of the `List` component that will be updated in version `1.2.0`.

### Solution

These types were copied from `ListRowProps` in `polaris-for-retail`.

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/components/List/List.ts
+++ b/packages/retail-ui-extensions/src/components/List/List.ts
@@ -1,5 +1,28 @@
 import {createRemoteComponent} from '@remote-ui/core';
 import {BadgeProps} from '../Badge';
+import type {
+  AccessibilityProps as RNAccessibilityProps,
+  AccessibilityPropsAndroid,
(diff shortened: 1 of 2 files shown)


packages/retail-ui-extensions/src/components/List/List.ts
```diff
@@ -19,6 +42,7 @@ export interface ListRowLeftSide {
      * A link to an image to be displayed on the far left side of the row.
      */
     source: string;
+    badge?: number;
```
NathanJolly, line 45: A comment here to describe what the badge does would be helpful

arigny: Yeah I missed it and was hoping I could push it before anyone noticed whoops

NathanJolly (approved):

PR #959, opened 2023-05-24
Create a DeviceAPI for name and deviceId

arigny:
### Background

Part of [23437](https://github.com/Shopify/pos-next-react-native/issues/23437)

Creates a DeviceAPI with `name` and `deviceId` properties.

### Solution

Create the DeviceAPI with `name` and `deviceId` properties and propagate it to the top of the `retail-ui-extensions` package.

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/extension-api/device-api/device-api.ts
+++ b/packages/retail-ui-extensions/src/extension-api/device-api/device-api.ts
@@ -1,0 +1,17 @@
+export interface DeviceApiContent {
+  /**
+   * The name of the device
+   */
+  name: string;
+  /**
+   * The string ID of the device
+   */
+  getDeviceId(): Promise<string>;
+}
+
+/**
+ * Access information about the device, like name and ID
+ */
+export interface DeviceApi {
+  device: DeviceApiContent;
+}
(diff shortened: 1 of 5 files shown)


vctrchu (approved):

packages/retail-ui-extensions/src/extension-api/device-api/device-api.ts
```diff
@@ -0,0 +1,17 @@
+export interface DeviceApiContent {
+  /**
+   * The name of the device
+   */
+  name: string;
```
js-goupil, line 5: What will this name entail? Like the operating system + model?

js-goupil (approved):
LGTM, however Alo yoga literally today was asking if they could get the model, as well as OS information, if it's possible to add that on.

packages/retail-ui-extensions/src/extension-api/device-api/device-api.ts
heltisace: I think it might be just `iPad` etc.

arigny: Ah yeah, for model it looks like you'd need `getModel`

PR #977, opened 2023-05-30
Remove react native and tests from List

heltisace:
### Background

https://github.com/Shopify/pos-next-react-native/issues/24526 added react-native dependency. The dependency mis-matches POS which crashes the app. It also is too big for ui-extensions to ingest.

### Solution

Removed accessibility and tests for now. This should be a separate discussion. 

### 🎩

- ...

### Checklist


yarn.lock
heltisace, line 1: I think they excluded the yarn.lock changes for some reason and then someone else merged them for us. This is also how we missed the extent of this.

PR #1000, opened 2023-06-08
Publish

js-goupil:
- @shopify/retail-ui-extensions-react@1.2.0-rc.4
 - @shopify/retail-ui-extensions@1.2.0-rc.4

### Background

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/retail-ui-extensions-react/package.json
+++ b/packages/retail-ui-extensions-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/retail-ui-extensions-react",
-  "version": "1.2.0-rc.3",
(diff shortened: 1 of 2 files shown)


heltisace:
Let's hold off merging this while we decide on how to handle Pin Pad. It might need to be reverted if it should be in 1.3, if this needs to get pushed.

heltisace:
@vctrchu is confident that the implementation of Pin Pad is solid so we will include it in POS UI Extensions 1.2.0 this time around. In the future, we will try to plan the feature work for POS UI Extensions more in advance.

@js-goupil we are good to merge.

PR #1001, opened 2023-06-08
Add missing pinpad react exports

vctrchu:
### Background

Apologies, I will have to make rc.5 to acommodate for these missed exports. I'll make sure to confirm if the props require a type so that they are exposed properly in the package next time.
### Solution

### 🎩

- ...

### Checklist


heltisace (approved):

heltisace (commented):
Not UI interfaces should be kept in Vanilla package

heltisace (approved):
These are UI interfaces so they do apply here.

js-goupil (approved):

PR #1022, opened 2023-06-16
CI: Create internal workflow

olavoasantos:
### Background

We require a new fork of the current implementation that will allow us to work on upcoming major changes, such as the React 18 update. As there may be breaking changes involved, we do not want to use the existing `unstable` branch, as this could potentially block further development. Therefore, we need a separate fork that will allow us to experiment with these changes without affecting the existing codebase. This will enable us to test and refine the new features before integrating them into the main development flow, ensuring a smoother transition and minimizing the risk of disruption to ongoing development.

### Solution

This PR introduces a new CI workflow that automatically deploys snapshots of the `internal` branch on demand. Similar to the `unstable` branch, every merge into the `internal` branch triggers the creation of a new set of versions that are published to NPM. A GitHub action monitors this branch and, upon detecting any changes, creates a snapshot release that publishes the changed packages as they exist at that point in time. Snapshot versions are identified by the format `0.0.0-internal-{{TIMESTAMP}}` and are published to NPM with the `internal` tag. Packages published to this tag should **not** be consumed externally.

### Checklist


vividviolet (approved):
Should this not be merged to main?

MitchLillie (approved):
LGTM, is this ready to ship?

vividviolet:
Yup, just merge it

PR #1030, opened 2023-06-21
Upgrade to React 18, including remote-ui deps

MitchLillie:
### Background

Resolves https://github.com/Shopify/app-ui/issues/175

### Solution

### 🎩

- ...

### Checklist


package.json
```diff
@@ -42,7 +42,7 @@
     "eslint": "^8.28.0",
     "nodemon": "^2.0.4",
     "prettier": "^2.8.0",
-    "react": ">=17.0.0 <18.0.0",
+    "react": ">=18.0.0 <19.0.0",
```
vividviolet, line 45: I think this should support version 17 too, no?

MitchLillie:
🎩 ed these changes in web & turkifier, looking good.

PR #1064, opened 2023-07-10
Add additional admin components

Fionoble:
### Background

This PR adds some of the missing components to the admin surface. These include:
- Box
- BlockStack
- Checkbox
- Form
- InlineStack
- TextArea

This PR also removes the following components:
- HorizontalStack
- VerticalStack

Follow up PRs will be in regards to adding the proper documentation files.

### Checklist


packages/ui-extensions-react/src/surfaces/admin/components/BlockStack/BlockStack.ts
```diff
@@ -0,0 +1,9 @@
+import {BlockStack as BaseBlockStack} from '@shopify/ui-extensions/admin';
+import type {BlockStackProps} from '@shopify/ui-extensions/admin';
+import {createRemoteReactComponent} from '@remote-ui/react';
+
+export const BlockStack = createRemoteReactComponent<
+  'BlockStack',
+  BlockStackProps
```
olavoasantos, line 7: NABD... Do we need to set the tag and props in these component definitions?

PR #1080, opened 2023-07-13
Convenience hooks for StandardApi in customer-account-ui-extensions

oleksandroliynyk:
### Background

Closes https://github.com/Shopify/core-issues/issues/55866
Closes https://github.com/Shopify/core-issues/issues/49473

**TL;DR**
Added hooks to expose functionality available in the `StandardApi` for `customer-account-ui-extensions`
Exposed `extensionPoint` API from the `StandardApi`

### Solution

Exposed hooks for each functionality in the StandardApi

### 🎩
Use this spin instance with all new hooks showcased: 
https://development-store-1.account.shopify.customer-accounts-ui-extension-dev-cyax.oleksandr-oliynyk.eu.spin.dev/orders?placement-reference=ORDERS2
(user: dev@shopify.com)

You should see example hooks used 
image

**OR:**

1. Create a spin instance as described in https://github.com/shopify/customer-accounts-ui-extension-dev/ . It will have both the UI extension and a consumer app to use them in
2. In spin, navigate to `ui-extensions`, checkout the branch and run `yarn build-consumer customer-account-web && yarn build-consumer customer-accounts-ui-extension-dev` 
3. No you can use new hooks in the `customer-accounts-ui-extension-dev` project in spin

### Checklist


packages/customer-account-ui-extensions-react/src/hooks/authenticated-redirect.ts
```diff
@@ -0,0 +1,13 @@
+import {
+  RenderExtensionPoint,
+  StandardApi,
+} from '@shopify/customer-account-ui-extensions';
+
+import {useApi} from './api';
+
+// Returns authenticated redirect api utilities for the extension point.
+export function useAuthenticatedRedirect<
```
robin-drexler, line 9: that's on me since we didn't remove it yet, but `authenticatedRedirects` are dead and we should not expose a hook. Can you please remove it?

packages/customer-account-ui-extensions-react/src/hooks/features.ts
```diff
@@ -0,0 +1,18 @@
+import {RenderExtensionPoint} from '@shopify/customer-account-ui-extensions';
+import {useApi} from './api';
+
+/**
+ * Returns a list of feature names for features enabled on a shop
+ */
+export function useFeatures<
```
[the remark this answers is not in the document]
emileber: I had started on a different approach, but then we figured just removing the type was good enough for now, so I just closed my other PR.

packages/customer-account-ui-extensions-react/src/hooks/i18n.ts
```diff
@@ -0,0 +1,15 @@
...
+import {useApi} from './api';
+
+/**
+ Returns utilities for translating content and formatting values
+    according to the current localization of the user.
+ */
+export function useI18n<
+  ID extends RenderExtensionPoint = RenderExtensionPoint,
+>(): I18n {
+  return useApi<ID>().i18n;
```
robin-drexler, line 14: As you yourself mentioned, in checkout they have more fine-grained hooks (e.g. translate) that  do some more work here. I think we should check why and potentially do the same, unless there's a good reason not to. I'd be ok if this happens in a follow-up PR if that's better for you. 

https://github.com/Shopify/ui-extensions/blob/633503bc8d64a3bea7b5b9b2537b3d91b7e3c7a6/packages/checkout-ui-extensions-react/src/hooks/translate.ts#L13

robin-drexler (approved):
Thanks for doing this and for adding tests + providing a great 🎩  experience!

brianshen1990 (approved):

PR #1088, opened 2023-07-14
Rename extension points to extension targets in admin surface

Fionoble:
### Background

Resolves part of: https://github.com/Shopify/app-ui/issues/226

### Solution

Rename ExtensionPoints to ExtensionTargets

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/globals.ts
+++ b/packages/ui-extensions/src/surfaces/admin/globals.ts
@@ -1,8 +1,8 @@
-import type {ExtensionPoints} from './extension-points';
+import type {ExtensionTargets} from './extension-targets';
 
 export interface ShopifyGlobal {
-  extend<ExtensionPoint extends keyof ExtensionPoints>(
-    extensionPoint: ExtensionPoint,
-    extend: ExtensionPoints[ExtensionPoint],
+  extend<ExtensionTarget extends keyof ExtensionTargets>(
+    extension: ExtensionTarget,
+    extend: ExtensionTargets[ExtensionTarget],
   ): void;
 }
--- a/packages/ui-extensions-react/src/surfaces/admin/context.ts
+++ b/packages/ui-extensions-react/src/surfaces/admin/context.ts
@@ -1,8 +1,8 @@
 import {createContext} from 'react';
 import type {
   ApiForRenderExtension,
-  RenderExtensionPoint,
+  RenderExtensionTarget,
 } from '@shopify/ui-extensions/admin';
 
 export const ExtensionApiContext =
-  createContext<ApiForRenderExtension<RenderExtensionPoint> | null>(null);
+  createContext<ApiForRenderExtension<RenderExtensionTarget> | null>(null);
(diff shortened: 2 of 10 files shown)


packages/ui-extensions/src/surfaces/admin/globals.ts
```diff
@@ -1,8 +1,10 @@
-import type {ExtensionPoints} from './extension-points';
+import type {ExtensionTargets} from './extension-targets';
 
 export interface ShopifyGlobal {
-  extend<ExtensionPoint extends keyof ExtensionPoints>(
-    extensionPoint: ExtensionPoint,
-    extend: ExtensionPoints[ExtensionPoint],
+  extend<ExtensionTarget extends keyof ExtensionTargets>(
+    extensionTarget: ExtensionTarget,
+    extend: {
```
vividviolet, line 6: Why do we need to update this?

vividviolet (commented):
Looking good to me - thanks for doing this. We will only update Admin for now because I believe Checkout still needs to do a non-breaking release for `23-04` from this branch.

Another option is to create that branch right now as a snapshot? Thoughts @jamesvidler?

packages/ui-extensions/src/surfaces/admin/globals.ts
Fionoble: Maybe we don't I wasn't actually sure

Fionoble: Fixed

packages/ui-extensions/src/surfaces/admin/globals.ts
```diff
@@ -1,8 +1,10 @@
-import type {ExtensionPoints} from './extension-points';
+import type {ExtensionTargets} from './extension-targets';
 
 export interface ShopifyGlobal {
-  extend<ExtensionPoint extends keyof ExtensionPoints>(
-    extensionPoint: ExtensionPoint,
-    extend: ExtensionPoints[ExtensionPoint],
+  extend<ExtensionTarget extends keyof ExtensionTargets>(
+    extension: {
+      target: ExtensionTarget;
```
vividviolet, line 6: Sorry, this is my bad, this has to be `extension: ExtensionTarget`

vividviolet (approved):

packages/ui-extensions/src/surfaces/admin/globals.ts
jamesvidler: Sorry for the confusion, but why would it be `extension: ExtensionTarget`? If anything `extend` should be `extension`,  since `extend` is replaced with `extension` in our JS API. 

Wouldn't this be more appropriate to just be `target: ExtensionTarget`?

vividviolet: yes, that's true that it can be `target:ExtensionTarget`. Either way the first param should be a string. Sorry @Fionoble do you mind updating? 

> If anything extend should be extension, since extend is replaced with extension in our JS API.
No, `extension` is the wrapper around `extend`. There's a bunch of stuff it does under the hood. `extend` is the low level API here: https://github.com/Shopify/ui-extensions/blob/a34277d5d1ce5c7fe5fda4d5d388a6db161c6618/packages/ui-extensions/src/utilities/registration.ts#L73

jamesvidler: Ok, I'm going to follow-suit in my [corresponding PR for checkout-web](https://github.com/Shopify/checkout-web/pull/23530).

PR #1095, opened 2023-07-17
Add IconName values for Admin Icon component

Fionoble:
### Background

Adds an enum for all the IconNames that an Icon component can use. This helps with documentation as well. 

### Checklist


elanalynn (approved):
LGTM!

PR #1106, opened 2023-07-18
Add NewTextField component and deprecation comments for old TextField

vctrchu:
Part of https://github.com/Shopify/polaris-for-retail/issues/182

### Background

This begins the migration the previous implementation of the original TextField implementation to use the new PFR component instead. Since we are releasing rc versions, we can temporary release a NewTextField and put deprecation comments for the existing TextField until we release version 2.0.0.

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/components/TextField/TextField.ts
+++ b/packages/retail-ui-extensions/src/components/TextField/TextField.ts
@@ -1,5 +1,5 @@
 import {createRemoteComponent} from '@remote-ui/core';
-import type {BaseTextFieldProps} from '../shared';
+import type {BaseTextFieldProps, InputProps} from '../shared';
 
 export interface ActionProps {
   type: 'action';
@@ -29,10 +29,17 @@ export type EmbeddedElementProps =
   | SuccessProps
   | PasswordProps;
 
+export interface NewTextFieldProps extends InputProps {}
 export interface TextFieldProps extends BaseTextFieldProps {
   rightElementStyle?: EmbeddedElementProps;
 }
 
-export const TextField = createRemoteComponent<'TextField', TextFieldProps>(
+/**
+ * @deprecated
+ * This TextField component will be replaced by NewTextField in version 2.0.0.
+ * Please migrate to using NewTextField as soon as possible.
+ */
+export const TextField = createRemoteComponent<
   'TextField',
-);
+  TextFieldProps | NewTextFieldProps
+>('TextField');
--- a/packages/retail-ui-extensions/src/index.ts
+++ b/packages/retail-ui-extensions/src/index.ts
@@ -94,6 +94,7 @@ export type {
   SectionProps,
   InputType,
   TextFieldProps,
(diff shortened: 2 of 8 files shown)


agorovyi:
I think due to `TextField` introducing breaking changes it will have to be added in `retail-ui-extensions` 2.0 release

packages/retail-ui-extensions/src/index.ts
```diff
@@ -92,11 +93,7 @@ export type {
   SectionProps,
   InputType,
   TextFieldProps,
-  ActionProps,
-  InfoProps,
-  SuccessProps,
-  PasswordProps,
-  EmbeddedElementProps,
```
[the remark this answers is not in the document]
vctrchu: Ya I'm adding them back rn, thanks

NathanJolly (approved):

packages/retail-ui-extensions/src/components/TextField/TextField.ts
```diff
@@ -1,23 +1,39 @@
 import {createRemoteComponent} from '@remote-ui/core';
 import type {BaseTextFieldProps} from '../shared';
-
+/**
+ * @deprecated
```
agorovyi, line 4: I think adding deprecation note to component itself should be sufficient as it will go away together with all its props.

vctrchu: Screenshot 2023-07-19 at 10 38 28 AM

That would be nice but I dont believe thats the case in Typescript

agorovyi:
> Should we be calling this "NewTextField"? I'm not really sure what else we could do instead though 🤔

@js-goupil The only other way is do nothing now about TextField and just introduce the breaking changes in 2.0.

The approach in this PR provides a "heads up" warning and alternative component to migrate to make upcoming breaking changes less cumbersome.

vctrchu:
> Is there any nice way to do a type union here and not introduce a new component?
> 
> Something like
> 
> image
> and then do some detection on the POS side based on which set of props the partner sets? If they set the old props, they get the old TextField, and if they set the new props, they get the new TextField
> 
> That way, the change wouldn't be breaking, there would only be one textfield component, and we could still deprecate the old props to warn folks they'll eventually be retired.

Great idea I like this, Im onboard for this approach since it involves less new moving parts and leverages our existing one

agorovyi:
> > Is there any nice way to do a type union here and not introduce a new component?
> > Something like
> > image
> > and then do some detection on the POS side based on which set of props the partner sets? If they set the old props, they get the old TextField, and if they set the new props, they get the new TextField
> > That way, the change wouldn't be breaking, there would only be one textfield component, and we could still deprecate the old props to warn folks they'll eventually be retired.
> 
> Great idea I like this, Im onboard for this approach since it involves less new moving parts and leverages our existing one

@vctrchu before your merge this PR let's check what the implementation on the POS side will look like

agorovyi (approved):
Looks good!

PR #1114, opened 2023-07-19
Update `Divider`'s `direction` values from `horizontal`/`vertical` to `block`/`inline`

billfienberg:
### Background

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components/Divider/Divider.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components/Divider/Divider.ts
@@ -4,9 +4,9 @@ export interface DividerProps {
   /**
    * Specify the direction of the divider.
    *
-   * @defaultValue 'horizontal'
+   * @defaultValue 'block'
    */
-  direction?: 'horizontal' | 'vertical';
+  direction?: 'block' | 'inline';
 }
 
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/src/surfaces/admin/components/Divider/Divider.ts
```diff
@@ -4,9 +4,9 @@ export interface DividerProps {
   /**
    * Specify the direction of the divider.
    *
-   * @defaultValue 'horizontal'
+   * @defaultValue 'block'
```
olavoasantos, line 7: This should be `inline`

olavoasantos: It's `inline`. If you check the implementation logic on the component, the default behaviour is `inline`.

billfienberg:
```suggestion
   * @defaultValue 'inline'
```

billfienberg: Fixed in e6b5b62622858f392e8fc6f53117cfc59f16ebff

PR #1115, opened 2023-07-19
Make source optional for POS UI extension Lists

heltisace:
### Background

In case the image display strategy is always, a Partner can provide a badge with a number even when the source is empty. 

- https://github.com/Shopify/pos-next-react-native/issues/26169
- https://github.com/Shopify/pos-next-react-native/pull/26173#issuecomment-1642783877

### Solution

Make source optional. 

### 🎩

- ...

### Checklist


heltisace:
@agorovyi @NathanJolly let me know what you think

rtymchyk (approved):

PR #1116, opened 2023-07-20
Fix `PixelEvents`, `EventBus` and `StandardEvents` types

emileber:
### Background

On all standard events, `data` is an object with some specific data. `'page_viewed'` of type `PixelEventsPageViewed` is the only one that doesn't include any explicit additional data. However, the implementation still provides an empty object.

### Solution

First, I added `PixelEventsPageViewedData` as an empty object to `page_viewed` event type. Then, I figured that some other types were slightly off, so I moved `custom_event` out of `PixelEvents` and now `CustomEvent` replaces `PixelEventsCustomEvent`.

I then updated `StandardEvents` so that the "all events" keys' possible values were properly typed. Which led me to also update the `EventBus` interface since it's not possible to publish to "all events", it's only possible to subscribe to those.

### Checklist

### DIFF
--- a/packages/web-pixels-extension/src/types/PixelEvents/index.ts
+++ b/packages/web-pixels-extension/src/types/PixelEvents/index.ts
@@ -151,22 +151,7 @@ export interface PixelEventsCollectionViewed {
   timestamp: Timestamp;
 }
 
-/**
- * This event represents any custom events emitted by partners or merchants via
- * the `publish` method
- */
-export interface PixelEventsCustomEvent {
-  clientId: ClientId;
-  context: Context;
-  customData: CustomData | null;
-  id: Id;
-
-  /**
-   * The name of the customer event
-   */
-  name: 'custom_event';
-  timestamp: Timestamp;
-}
+export interface PixelEventsPageViewedData {}
 
 /**
  * The `page_viewed` event logs an instance where a customer visited a page.
@@ -175,6 +160,7 @@ export interface PixelEventsCustomEvent {
 export interface PixelEventsPageViewed {
   clientId: ClientId;
   context: Context;
+  data: PixelEventsPageViewedData;
   id: Id;
 
   /**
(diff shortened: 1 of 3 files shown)


packages/web-pixels-extension/src/types/PixelEvents/index.ts
```diff
@@ -338,12 +346,6 @@ export interface PixelEvents {
    */
   collection_viewed: PixelEventsCollectionViewed;
 
-  /**
-   * This event represents any custom events emitted by partners or merchants
-   * via the `publish` method
-   */
-  custom_event: PixelEventsCustomEvent;
```
emileber, line 345: Removed this key from `PixelEvents` since it's not a standard event. `PixelEventsCustomEvent` would now be `CustomEvent`

scripts/consumer-helper.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-AVAILABLE_PACKAGES=('admin-ui-extensions' 'admin-ui-extensions-react' 'checkout-ui-extensions' 'checkout-ui-extensions-react' 'post-purchase-ui-extensions' 'post-purchase-ui-extensions-react' 'checkout-ui-extensions-run' 'customer-account-ui-extensions' 'customer-account-ui-extensions-react')
+AVAILABLE_PACKAGES=('admin-ui-extensions' 'admin-ui-extensions-react' 'checkout-ui-extensions' 'checkout-ui-extensions-react' 'post-purchase-ui-extensions' 'post-purchase-ui-extensions-react' 'checkout-ui-extensions-run' 'customer-account-ui-extensions' 'customer-account-ui-extensions-react' 'web-pixels-extension')
```
emileber, line 3: Adds a missing package for the build-consumer command.

sorta-shopify (approved):
Looks good to me! 👍

emileber:
Release PR:
- https://github.com/Shopify/ui-extensions/pull/1252

PR #1132, opened 2023-07-21
useApi(): infer target from dummy argument

developit:
### Background

This makes it possible to use `useApi()` in JS with full type inference. Previously this required extensive jsdoc annotations that nobody would bother doing.

### Solution

Inference from argument, the same as how `reactExtension(target)` works.


MitchLillie (approved):
LGTM!

PR #1133, opened 2023-07-21
Add data to block and action types

MitchLillie:
### Background

Adds `data` which was missing from action and block extensions, and also removes `i18n` since that is present in the standard API.

### Solution

### 🎩

- ...

### Checklist


olavoasantos (approved):
LGTM. You're missing `changesets` though.

PR #1136, opened 2023-07-24
Make @shopify/ui-extensions a peer dependency

lemonmade:
### Background

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/.changeset/soft-flies-repair.md
+++ b/.changeset/soft-flies-repair.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions-react': patch
+---
+
+Make @shopify/ui-extensions a peer dependency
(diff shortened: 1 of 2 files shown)


vividviolet (commented):
@lemonmade do you know if this might make the conflicting remote-ui versions between vanilla and React worse? I could do some testing on with pnpm, npm and yarn.

lemonmade:
I don't think it makes the situation any worse, since the React library can still declare as strict a peer dependency range as it wants.

developit (approved):

lemonmade:
@vividviolet yes, I will include it in this collection of backported changes I am queuing up: https://github.com/Shopify/ui-extensions/pull/1135/files

PR #1189, opened 2023-08-01
Extension editor

brianshen1990:
### Background

Part of https://github.com/Shopify/core-issues/issues/58453 
Add UI Extension Editor interface, just like what checkout did to indicate whether it is rendered within an editor so 3rd part developers can simulate data in this case

customer-account-web side: https://github.com/Shopify/customer-account-web/pull/2664 

### Solution

### 🎩

See tophat in this PR https://github.com/Shopify/customer-account-web/pull/2664 

### Checklist

### DIFF
--- a/packages/customer-account-ui-extensions/src/extension-points/standard-api/extension.ts
+++ b/packages/customer-account-ui-extensions/src/extension-points/standard-api/extension.ts
@@ -2,6 +2,12 @@ import type {StatefulRemoteSubscribable} from '@remote-ui/async-subscription';
 
 export type Capability = 'network_access';
 
+export interface Editor {
+  /**
+   * Indicates whether the extension is rendering in the checkout editor.
+   */
+  type: 'checkout';
+}
 export interface Extension {
   /**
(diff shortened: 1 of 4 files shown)


packages/customer-account-ui-extensions/src/extension-points/standard-api/extension.ts
```diff
@@ -40,4 +46,11 @@ export interface Extension {
    * to make network calls.
    */
   capabilities: StatefulRemoteSubscribable<Capability[]>;
+
+  /**
+   * Information about the editor where the extension is being rendered.
+   *
+   * The value is undefined if the extension is not rendering in an editor.
+   */
+  editor?: Editor;
```
brianshen1990, line 55: Same as checkout https://github.com/Shopify/ui-extensions/blob/c866b23cbbeb28f21db045f5142b796ed790d18d/packages/ui-extensions/src/surfaces/checkout/api/standard/standard.ts#L89

lihaokx (approved):
LGTM

PR #1194, opened 2023-08-02
Admin authentication

AaronSpringut:
### Background

Folks kept asking if they could use these with their apps' backends, or if they could get a session token. 
### Solution

Documented how fetch requests will automatically be authenticated and added a code sample.

### 🎩

- Generate the docs
- Copy them to shopify-dev
- Preview the changes

### Checklist

- [X ] I have :tophat:'d these changes
- [ X, but this one feels like a trick question for this commit] I have updated relevant documentation

### DIFF
--- a/packages/ui-extensions/docs/surfaces/admin/staticPages/admin-extensions.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/admin/staticPages/admin-extensions.doc.ts
@@ -100,7 +100,7 @@ const data: LandingTemplateSchema = {
         title: "Make requests to your app's backend",
         tabs: [
           {
-            code: './examples/direct-api.jsx',
+            code: './examples/authenticated-fetch.jsx',
             language: 'tsx',
             title: 'Get Product Data',
           },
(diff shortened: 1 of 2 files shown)


packages/ui-extensions/docs/surfaces/admin/staticPages/admin-extensions.doc.ts
```diff
@@ -84,6 +90,23 @@ const data: LandingTemplateSchema = {
         ],
       },
     },
+    {
+      type: 'Generic',
+      title: 'App Authentication',
+      sectionContent:
+        "Admin UI extensions can also make authenticated calls to your app's backend. When you make a `fetch()` request to your app's backend, an authorization token will be automatically added to the call's headers. There's no need to manually manage your app's session tokens.",
```
Fionoble, line 97:
```suggestion
        "Admin UI extensions can also make authenticated calls to your app's backend. When you make a `fetch()` request to your app's backend, an authorization token will be automatically added to the request headers. There's no need to manually manage your app's session tokens.",
```

Fionoble (approved):
Just the one change

PR #1196, opened 2023-08-03
stevep/update-cost-api

Steve-D-Powell:
### Background

Related to #23770

Added new API endpoints for the standard API `cost` endpoint for checkout UI extensions.

### 🎩

```
https://shopify-dev.cost-api-hook-docs.stephen-powell.us.spin.dev/docs/api/checkout-ui-extensions/unstable/apis/standardapi
```

### Checklist


packages/ui-extensions-react/README.md
jamesvidler, line 1: @Steve-D-Powell can you exclude these changes from your PR? 

It looks like these have been modified directly in this repo by someone else and we don't want to overwrite them. If you can, would you be able to incorporate these changes back into our private `packages/ui-extensions-react/README.md` so we don't need to worry about this in the future?

packages/ui-extensions/docs/surfaces/checkout/screenshots/supported-locations-shipping.png
jamesvidler, line 1: I'm not sure why this change is in here.

packages/ui-extensions-react/README.md
Steve-D-Powell: @jamesvidler I manually copied the original contents of these files back so they won't be overridden. 

But when I look at the files in checkout-web (which I think is the private packages, right? Sorry still new to this) they are the updated version there: 

* https://github.com/Shopify/checkout-web/blob/main/packages/ui-extensions-react/README.md
* https://github.com/Shopify/checkout-web/blob/main/packages/ui-extensions/README.md

packages/ui-extensions/docs/surfaces/checkout/screenshots/supported-locations-shipping.png
jamesvidler: Oh, there is a fix here for the extension target name being used. This is ok to include 👍 .
Screenshot 2023-08-04 at 12 57 51 PM

packages/ui-extensions-react/src/surfaces/checkout/components.ts
```diff
@@ -73,6 +73,9 @@ export type {ListProps} from './components/List/List';
 export {ListItem} from './components/ListItem/ListItem';
 export type {ListItemProps} from './components/ListItem/ListItem';
 
+export {LoginWithShop} from './components/LoginWithShop/LoginWithShop';
+export type {LoginWithShopProps} from './components/LoginWithShop/LoginWithShop';
+
```
jamesvidler, line 78: This is a new component. We should add @mikewilson01 as a reviewer to ensure they are aware that will be discoverable in the new package and usable, even if the feature or documentation isn't ready yet.

jamesvidler (approved):
This looks good from my end, tophatted the docs as well. Assuming the below can be accomplished:

We should be including two additional changesets on this PR:
- `LoginWithShop` should be **minor** (it is a new feature)
- Documentation updates **patch**

packages/ui-extensions-react/src/surfaces/checkout/components.ts
mikewilson01: Hi @jamesvidler - thanks for pointing this out, yes this is safe to ship. It's purposely not (publicly) documented as it's intended to be used for a single merchant. It's currently gated behind a beta flag, so though it's discoverable, it's not usable.

jamesvidler: Great 👍

mikewilson01 (approved):

.changeset/short-wolves-laugh.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/ui-extensions': minor
+'@shopify/ui-extensions-react': minor
+---
+
+Uses the functionality of Login with Shop to power logging in or signing up with Shop. Upon success, redirects to Shop Pay checkout.
```
jamesvidler, line 6: Language nit: 
```suggestion
New `LoginWithShop` UI component for checkout. Redirects the buyer to Shop Pay.
```

jamesvidler (approved):
Small language nit.

PR #1198, opened 2023-08-04
changset should use unstable branch as base

jamesvidler:
### Background

When running `yarn changset` it should be using the `unstable` branch as the base branch.

### Solution

Updates the config.

### Checklist


lemonmade (approved):
Should the versioned branches also be updated to point at different branch names here?

jamesvidler:
> Should the versioned branches also be updated to point at different branch names here?

That's a good question. I'll look into it.

PR #1226, opened 2023-08-14
Adds `CustomerSegmentTemplate` component

lauramann:
Resolves https://github.com/Shopify/customer-data-platform/issues/5767

### Background

We already have a CustomerSegmentationTemplate ui extension component (see https://github.com/Shopify/ui-extensions/pull/759) that we use to render templates from our [1p app](https://github.com/Shopify/segmentation-templates-app) in the admin. 

However, we want to extend this functionality to other 1p and 3p apps, and we have different requirements for this new public component than the one we use currently (we don't want other 1p/3p apps to be able to pass a custom icon or category, we want to support custom metafields). 

### Solution

Since web uses the `unstable` version of ui-extensions, we're adding a second component that will be [documented](https://shopify.dev/docs/api/admin-extensions/components) for 1p/3p apps to use, and we will continue to use our other CustomerSegmentationTemplate for our 1p app. 

Adds the component in both `ui-extensions/components` and `ui-extensions-react/components`.

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/api/customer-segment-template/customer-segment-template.ts
+++ b/packages/ui-extensions/src/surfaces/admin/api/customer-segment-template/customer-segment-template.ts
@@ -9,7 +9,7 @@ type CustomerSegmentationFeature =
   /* Enables count aggregates on functions. For example: shopify_email.opened(count_at_least: 5) = true */
   | 'aggregateFilters';
 
-export interface CustomerSegmentationTemplateApi<
+export interface CustomerSegmentTemplateApi<
   ExtensionTarget extends AnyExtensionTarget,
 > extends StandardApi<ExtensionTarget> {
   /* Utilities for translating content according to the current `localization` of the admin. */
--- a/packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentTemplate/CustomerSegmentTemplate.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentTemplate/CustomerSegmentTemplate.ts
@@ -1,0 +1,33 @@
+import {createRemoteComponent} from '@remote-ui/core';
+
+/**
+ * Reserved namespace and key for the customer standard metafield used in the template's query.
+ * More info - https://shopify.dev/docs/apps/custom-data/metafields/definitions/standard
+ */
+type CustomerStandardMetafieldDependency = 'facts.birth_date';
+
+export interface CustomerSegmentTemplateProps {
+  /* Localized title of the template. */
+  title: string;
+  /* Localized description(s) of the template. */
+  description: string | string[];
+  /* Code snippet to render in the template with syntax highlighting. */
+  templateQuery: string;
+  /* Code snippet to insert in the segment editor. If missing, `templateQuery` will be used. */
+  templateQueryToInsert?: string;
+  /* List of customer standard metafields or custom metafields used in the template's query. */
+  dependencies?: {
+    standardMetafields?: CustomerStandardMetafieldDependency[];
+    customMetafields?: string[];
+  };
+  /* ISO 8601-encoded date and time string. A "New" badge will be rendered for recently introduced templates. */
+  dateAdded?: string;
+}
+
+/**
+ * Customer segmentation templates are used to give merchants a starting point to create segments.
+ */
+export const CustomerSegmentTemplate = createRemoteComponent<
+  'CustomerSegmentTemplate',
+  CustomerSegmentTemplateProps
+>('CustomerSegmentTemplate');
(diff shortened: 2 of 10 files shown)


packages/ui-extensions/src/surfaces/admin/api/customer-segment-template/customer-segment-template.ts
```diff
@@ -9,7 +9,7 @@ type CustomerSegmentationFeature =
   /* Enables count aggregates on functions. For example: shopify_email.opened(count_at_least: 5) = true */
   | 'aggregateFilters';
 
-export interface CustomerSegmentationTemplateApi<
+export interface CustomerSegmentTemplateApi<
```
lauramann, line 12: Also renamed the API to match, but both components use the same API.

packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentationTemplate/CustomerSegmentationTemplate.ts
```diff
@@ -51,6 +51,11 @@ export interface CustomerSegmentationTemplateProps {
   templateQueryToInsert?: string;
   /* List of customer standard metafield used in the template's query. */
   standardMetafieldDependencies?: CustomerStandardMetafieldDependency[];
```
PrescilliaA, line 53: @LTiger14 we are going to keep this field until we swap it out in segmentation templates and web.

lauramann:
Note: Documentation will come after (tracked [here](https://github.com/Shopify/customer-data-platform/issues/5713)).

packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentTemplate/CustomerSegmentTemplate.ts
```diff
@@ -0,0 +1,33 @@
...
+ * More info - https://shopify.dev/docs/apps/custom-data/metafields/definitions/standard
+ */
+type CustomerStandardMetafieldDependency = 'facts.birth_date';
+
+export interface CustomerSegmentTemplateProps {
+  /* Localized title of the template. */
+  title: string;
+  /* Localized description(s) of the template. */
+  description: string | string[];
+  /* ShopifyQL code snippet to render in the template with syntax highlighting. */
```
LTiger14, line 14: Nitpicky but it is it actually ShopifyQL? Notebook uses ShopifyQL we... have a different syntax lol 😅 just wondering if it can cause confusion

packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentTemplate/CustomerSegmentTemplate.ts
```diff
@@ -0,0 +1,33 @@
...
+  templateQuery: string;
+  /* ShopifyQL code snippet to insert in the segment editor. If missing, `templateQuery` will be used. */
+  templateQueryToInsert?: string;
+  /* List of customer standard metafields or custom metafields used in the template's query. */
+  dependencies?: {
+    standardMetafields: CustomerStandardMetafieldDependency[];
+    customMetafields: string[];
```
LTiger14, line 21: We should have them both as optional if not adding one of the two requires the other to be there as well:

```suggestion
    standardMetafields?: CustomerStandardMetafieldDependency[];
    customMetafields?: string[];
```

Now a nit: doing the above would make `dependencies: {}` valid which if ok but  we could go crazy about it.

See [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgApwK4BtkG8CwAUMicgA6ZYD8AXMgM5hSgDmA2gLoDcRAvkUVCRYiFADEocEAGs8RUshiSZtBk1acehfoUHho8JMgDCWTABMUBYqRBwAthDqNmIFloVgA9vdUAlCABHDGAoCHMAHnRsZAAfZAkpaQA+PgFdQgQvEEY1CAA3CBA6UwsUAF45GxI7RzoAcgBlSEKQeoAaeVJvezpcHV4gA)

```ts
interface Paul {
    paul?: string[];
}

interface Frank {
    frank?: string[];
}

interface Claude {
    name: string;
    tom?: Required<Paul | Frank>
}
```

packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentationTemplate/CustomerSegmentationTemplate.ts
```diff
@@ -51,6 +51,11 @@ export interface CustomerSegmentationTemplateProps {
...
+  /* List of customer standard metafields or custom metafields used in the template's query. */
+  dependencies?: {
```
LTiger14, line 55: Same here

LTiger14 (approved):
LGTM! Please just make sure to look at the `dependencies` 🙂

packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentTemplate/CustomerSegmentTemplate.ts
lauramann: good catch, I just removed ShopifyQL from them both (so it's just "code snippet...")

packages/ui-extensions/src/surfaces/admin/components/CustomerSegmentTemplate/CustomerSegmentTemplate.ts
lauramann: good catch! I made them both optional within it 👍

PR #1229, opened 2023-08-14
add ActionExtensionApi to customer account web ui

brianshen1990:
### Background

Part of https://github.com/Shopify/core-issues/issues/58096 

add ActionExtensionApi to customer account web ui
```
export interface ActionExtensionApi {
  close(): void;
}
```

### Solution

### 🎩

- ...

### Checklist


gracerivas (approved):
LGTM

Hey Brian, nice addition, would be nice if we had a description in the PR of what and why an ActionExtensionApi added here, just to gain immediate context, I didn't understand the issue until I saw the example in the admin. TY!

lsit (approved):

PR #1254, opened 2023-08-17
Fix tab names for Direct API examples

vividviolet:
Fix the tab names being cut off in the docs and also switch to sentence casing to match convention

**Before**
image

**After**
image


MitchLillie (approved):
LGTM!

PR #1255, opened 2023-08-17
Add back button hijacking callback to screen component

vctrchu:
Part of https://github.com/Shopify/pos-next-react-native/issues/26548

### Background
Stocky wants to implement the ability to display a dialog when the back button is pressed resulting in the screen in not navigating back. Currently `onNavigateBack` only performs some sort of logic after the screen is unmounted however we need another callback to run while mounted. 

You can see a draft POS Screen implementation below which adds a secondary action depending if this prop is not undefined:

```ts
    <Container
      title={title}
      inset="none"
      scrollViewAsContainer={false}
      secondaryAction={
        onSecondaryActionPressed
          ? {
              text: i18n.translate('Screen.back'),
              onPress: () => {
                onSecondaryActionPressed();
              },
            }
          : undefined
      }
    >
```

### Solution

Open to discussion about naming and callback type. We need to make it clear on the differentiation between this callback and the onNavigateBack callback...

### DIFF
--- a/packages/retail-ui-extensions/src/components/Screen/Screen.ts
+++ b/packages/retail-ui-extensions/src/components/Screen/Screen.ts
@@ -12,7 +12,8 @@ export interface ScreenPresentationProps {
  * @property `title` the title of the screen which will be displayed on the UI.
  * @property `isLoading` displays a loading indicator when `true`. Set this to `true` when performing an asynchronous task, and then to false when the data becomes available to the UI.
  * @property `onNavigate` triggered when the screen is navigated to.
- * @property `onNavigateBack` triggered when the user navigates back from this screen.
+ * @property `onNavigateBack` triggered when the user navigates back from this screen. Runs after screen is unmounted.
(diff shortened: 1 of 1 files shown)


packages/retail-ui-extensions/src/components/Screen/Screen.ts
```diff
@@ -22,6 +23,7 @@ export interface ScreenProps {
   presentation?: ScreenPresentationProps;
   onNavigate?: () => void;
   onNavigateBack?: () => void;
+  onSecondaryActionPress?: () => void;
```
heltisace, line 26: Looking back at how these things are named, I with we had started with something like this instead:
- Naming `onSecondaryActionPress` as `onNavigateBack`.
- Naming `onNavigateBack` as `onNavigatedBack`. 
- Naming `onNavigate` as `onNavigatedHere`. 

As things stand, that would probably be a breaking change. For the present moment, what do we think about using a name like `overrideNavigateBack` instead? Although it will also be called not just for `back`, but a `close` action too, this follows the current naming of parameters and builds on it.

js-goupil (approved):

PR #1256, opened 2023-08-18
Add customer account action component

brianshen1990:
### Background

### Solution

### 🎩

- ...

### Checklist


packages/customer-account-ui-extensions/src/components/CustomerAccountAction/CustomerAccountAction.ts
```diff
@@ -0,0 +1,22 @@
+import {createRemoteComponent} from '@remote-ui/core';
+import type {ActionComponents} from 'component-set';
```
robin-drexler, line 2: Does this import work? And if it does, should we still import relatively?

I first thought `component-set` would be an npm package. 😅

brianshen1990: Screenshot 2023-08-18 at 15 13 03

both `../../component-set` and `component-set` works, updating it to relative path now

brianshen1990: updated in the commit https://github.com/Shopify/ui-extensions/pull/1256/commits/87e2a9838483929ce58859a1c820a06e3ffc42e2

packages/customer-account-ui-extensions/src/components/CustomerAccountAction/CustomerAccountAction.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export interface CustomerAccountActionProps {
+  /**
+   * A title rendered as a `Heading` at the top of the modal.
+   */
+  title: string;
+  /**
+   * Action modal's primary action, ie 'Save' or 'Accept'.
+   */
+  primaryAction?: ActionComponents;
```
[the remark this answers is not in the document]
robin-drexler: in admin they use `RemoteFragment`.

I wonder if we have to as well? https://github.com/Shopify/ui-extensions/blob/unstable/packages/ui-extensions/src/surfaces/admin/components/AdminAction/AdminAction.ts#L13

I'm genuinely not sure.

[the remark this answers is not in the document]
brianshen1990: changed, and keep everything same with admin 

The only issue is that with `remoteFragement`, on 3P developer side, it will always cause lint issue ;( 
Screenshot 2023-08-18 at 17 02 31

lsit (approved):

PR #1275, opened 2023-08-22
Expose `CustomerSegmentTemplate` icon and category types

PrescilliaA:
### Background

Right now we are duplicating the icon and category types on web, and this makes 2 sources of truth for template icon sources and categories to maintain. 

### Solution

Exposing both the array and type publicly will help clean up web, and result in less maintenance in the long term. 

### Checklist

### DIFF
--- a/.changeset/popular-eggs-remember.md
+++ b/.changeset/popular-eggs-remember.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+Expose CustomerSegmentTemplate category and icon types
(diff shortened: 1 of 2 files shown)


LTiger14 (approved):

.changeset/popular-eggs-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
```
lauramann, line 2: should this be a minor change or patch?

PR #1283, opened 2023-08-24
Move customer account ui extension from main to unstale

lihaokx:
### Background

### Solution

### 🎩

- ...

### Checklist


packages/ui-extensions-react/src/surfaces/customer-account.ts
```diff
@@ -6,76 +6,19 @@ export {
...
-  Alignment,
-  AllowedComponentsForRenderExtension,
-  ApiForRenderExtension,
-  Appearance,
-  AtLeastOne,
   ArgumentsForExtension,
-  Autocomplete,
-  AutocompleteField,
-  AutocompleteGroup,
-  Background,
```
lihaokx, line 19: Remove the type that are not exported in `customer-account-ui-extensions-react`

packages/ui-extensions-react/src/surfaces/customer-account/hooks/app-metafields.ts
```diff
@@ -25,7 +25,7 @@ type AppMetafieldFilterKeys = keyof AppMetafieldFilters;
  * @arg {AppMetafieldFilters} - filter the list of returned metafields
  */
 export function useAppMetafields<
-  Target extends RenderExtensionTarget = RenderExtensionTarget,
+  Target extends OrderStatusExtensionTarget = OrderStatusExtensionTarget,
```
lihaokx, line 28: Reason: `appMetafields` only exists on `OrderStatusApi` instead of other APIs. Only extensions on `OrderStatus` can call this hook. Thus, we should limit the targets to the targets on OrderStatus

packages/ui-extensions-react/src/surfaces/customer-account/hooks/extension-data.ts
lihaokx, line 1: This is removed because it is duplicated with [this file](https://github.com/Shopify/ui-extensions/blob/unstable/packages/ui-extensions-react/src/surfaces/customer-account/hooks/extension.ts), and [useExtensionData](https://shopify.dev/docs/api/checkout-ui-extensions/unstable/react-hooks/metadata/useextensiondata) is deprecated.

packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
```diff
@@ -9,129 +9,12 @@ import type {
...
- * Stored data is only available to this specific app
- * at this specific extension target.
- *
- * The storage backend is implemented with `localStorage` and
- * should persist across the buyer's checkout session.
- * However, data persistence isn't guaranteed.
- */
-export interface Storage {
-  /**
-   * Read and return a stored value by key.
```
lihaokx, line 27: Move them to the shared.ts

packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
```diff
@@ -344,7 +148,7 @@ export interface Market {
   handle: string;
 }
 
-export interface Localization {
+export interface OrderStatusLocalization {
```
lihaokx, line 151: Reason: `Localization` exists on both OrderStatusApi and StandardApi. Rename it to resolve the conflict. The same reason for the changes follows.

packages/ui-extensions/src/surfaces/customer-account/extension.ts
```diff
@@ -19,3 +19,7 @@ export const extension =
  * @deprecated This is deprecated, use `extension` instead.
  */
 export const extend = extension;
+
+export interface RunExtension<Api, Return> {
+  (api: Api): Promise<Return>;
+}
```
brianshen1990, line 25: from here https://github.com/Shopify/ui-extensions/blob/main/packages/customer-account-ui-extensions/src/extension-points/extension-signature.ts 

only for return, should be removed soon

lihaokx: Sure. Will remove as well.

packages/ui-extensions/src/surfaces/customer-account/targets.ts
```diff
@@ -49,8 +63,65 @@ export interface ExtensionTargets {
...
+  >;
+  'customer-account.account-information.location.render-after': RenderExtension<
+    StandardApi<'customer-account.account-information.location.render-after'>,
+    AllComponents
+  >;
+  'customer-account.order-status.action.menu-item.render': RenderExtension<
+    StandardApi & {orderId: string},
+    AllComponents
+  >;
+  'customer-account.order-status.action.render': RenderExtension<
```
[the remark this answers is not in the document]
lihaokx: Added it !

packages/ui-extensions-react/src/surfaces/customer-account/hooks/buyer-identity.ts
```diff
@@ -73,7 +75,7 @@ export function usePhone<
  */
 export function usePurchasingCompany<
   Target extends RenderExtensionTarget = RenderExtensionTarget,
->(): PurchasingCompany | undefined {
+>(): OrderStatusPurchasingCompany | PurchasingCompany | undefined {
   const buyerIdentity = useApi<Target>().buyerIdentity;
```
brianshen1990, line 79: this one we might be able to do better un the return type: 
`Target extends OrderStatusExtensionTarget ? OrderStatusPurchasingCompany | undefined: PurchasingCompany | undefined`

packages/ui-extensions-react/src/surfaces/customer-account/hooks/extension-data.ts
```diff
@@ -1,15 +0,0 @@
-import type {
-  Extension,
-  RenderExtensionTarget,
-} from '@shopify/ui-extensions/customer-account';
-
-import {useApi} from './api';
-
-/**
```
brianshen1990, line 8: why this is removed? did we moved it somewhere else? if checkout added it, we should keep it maybe

packages/ui-extensions-react/src/surfaces/customer-account/hooks/router.ts
```diff
@@ -0,0 +1,25 @@
...
+
+// Returns the router utilities for the extension point.
+export function useRouter<
+  Target extends RenderExtensionTarget = RenderExtensionTarget,
+>(): StandardApi['router'] {
+  const api = useApi<Target>();
+
+  if ('router' in api) {
+    return api.router;
+  }
```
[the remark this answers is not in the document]
lihaokx: Done! Thanks

lsit (commented):
I think this looks fine from a component perspective. Hard to tell without being able to test. We do have [this PR](https://github.com/Shopify/ui-extensions/pull/1280) that almost ready to ship that adds more components. Would you rather us ship before or after this migration?

lihaokx:
> We do have https://github.com/Shopify/ui-extensions/pull/1280 that almost ready to ship that adds more components. Would you rather us ship before or after this migration?

There might be more minor changes after this one is shipped. Since this is not used now, the order does not matter. I can merge this one first, then change it in a later pr. @lsit

packages/ui-extensions-react/src/surfaces/customer-account/hooks/settings.ts
```diff
@@ -8,8 +11,9 @@ import {useSubscription} from './subscription';
  */
 export function useSettings<
   Settings extends ExtensionSettings,
+  Target extends RenderOrderStatusExtensionTarget = RenderOrderStatusExtensionTarget,
```
alxclark, line 14: I don't know where we have the list of apis that will come to customer account extensibility, but do you know if settings will become part of standard api eventually ?

packages/ui-extensions/src/surfaces/customer-account/targets.ts
```diff
@@ -12,7 +21,12 @@ import type {RenderExtension} from './extension';
  * The input arguments and the output type are different
  * for each extension target.
  */
-export interface ExtensionTargets {
+export interface ExtensionTargets
+  extends OrderStatusExtensionTargets,
+    CustomerAccountExtensionTargets {}
```
alxclark, line 26: Can we have a proper union/intersection here ? I don't think we would ever have a member live in the interface's body ?

packages/ui-extensions-react/src/surfaces/customer-account/hooks/buyer-identity.ts
brianshen1990: the `PurchasingCompany | undefined` from customer account standard api side

alxclark (approved):
I mostly focused on the order-status renames and changes looks good. In a perfect world we wouldn't have collisions between our types and apis that are available on standard vs extension point specific, but its pretty hard to back track some of those decisions now. When we start planning for the next stable version we'll be able to reshuffle some of those fields, set deprecations and hopefully remove some of the type weirdness.

Thanks for this massive job 🙌🏼 

I left a comment regarding wether or not we should backport some of the type changes already. It's not a blocker for this to go out, we can take that discussion async.

lsit (approved):
FYI https://github.com/Shopify/ui-extensions/pull/1280 has shipped

PR #1302, opened 2023-09-01
Fix Direct API query example for Admin

vividviolet:
A partner tried to use the `query` API as well as `fetch` and noticed that our docs were wrong


developit (approved):
these are likely my fault!

PR #1315, opened 2023-09-05
Add checkoutToken api to unstable release

dnagoda:
### Background

Resolves https://github.com/Shopify/temp-project-mover-megswim-20250326195306/issues/390

NOTE: I only included my changes. There's an issue with the current customer-account changes that blows up the build.

### Solution

Provides stable id in checkout that can be mapped to WebPixel checkout payloads (via `token` in checkout payload) and Orders resource (via Admin REST API Orders `checkout_token`).

### 🎩

- Create a spin instance that includes the `ui-extensions` repo and checkout this branch: `denii/unstable-release-2023-09-05`
- Create a new app with checkout ui extension locally: `yarn create @shopify/app` and `yarn generate extension`
- In the local extension, use the new api, both directly and through the hook:
```
const { checkoutToken } = useApi();

const value = useSubscription(
  checkoutToken,
);
```
or
```
const hookValue = useCheckoutToken();
```
- In your app directory, run the following to copy the `ui-extensions` code into your app:
```
VM=direct.$(spin show -o fqdn)
rsync -av --delete $VM:~/src/github.com/Shopify/ui-extensions/packages/ui-extensions node_modules/@shopify/ && \
rsync -av --delete $VM:~/src/github.com/Shopify/ui-extensions/packages/ui-extensions-react node_modules/@shopify/
```
- Run `yarn dev` or `yarn build` to check that the app builds and can execute without error using the updated unstable api

### Checklist

### DIFF
--- a/.changeset/yellow-frogs-help.md
+++ b/.changeset/yellow-frogs-help.md
@@ -1,0 +1,11 @@
+---
+'@shopify/ui-extensions': minor
+'@shopify/ui-extensions-react': minor
+---
+
+# Add checkoutToken Api
+
+Extensions now have access to the checkout token (a stable id used to identify the checkout) directly through the standard api. The checkout token will match:
+
+- **token** field in [WebPixel checkout payload](https://shopify.dev/docs/api/pixels/customer-events#checkout).
+- **checkout_token** in the [Admin REST Api Order resource](https://shopify.dev/docs/api/admin-rest/unstable/resources/order#resource-object)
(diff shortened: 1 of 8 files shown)


.changeset/yellow-frogs-help.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/ui-extensions': minor
+'@shopify/ui-extensions-react': minor
+---
+
+Add checkoutToken api
```
jamesvidler, line 6: Can we make this more descriptive? This will ultimately be part of a public changelog and we want to start putting more effort into these. For example, what problem does this resolve/feature does it unlock?

jamesvidler (approved):
Nicely done 👍 Just a comment regarding the changeset description. Otherwise good to go.

.changeset/yellow-frogs-help.md
dnagoda: Updated.

PR #1319, opened 2023-09-06
Expose secondary action on screen component

bryankou:
### Background
Part of https://github.com/Shopify/pos-next-react-native/issues/27009
As part of stocky work for M2, we need to expose `secondaryAction` for the screen component. The component itself already exists. 

### Solution

Let me know if the typing or documentation is correct.

### 🎩

- ...

### Checklist


vctrchu:
You have to export the new secondary action interface but other than that it looks great! You can search `ActionProps` and see where to update the changes.

Screenshot 2023-09-07 at 9 31 05 AM

vctrchu (approved):
Looks good, please squash the commits before merging. Thanks!

PR #1323, opened 2023-09-07
Publish

vctrchu:
- @shopify/retail-ui-extensions-react@1.5.0-rc.0
 - @shopify/retail-ui-extensions@1.5.0-rc.0
 
 Part of https://github.com/Shopify/ui-extensions/pull/1319

### DIFF
--- a/packages/retail-ui-extensions-react/package.json
+++ b/packages/retail-ui-extensions-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/retail-ui-extensions-react",
-  "version": "1.4.0-rc.3",
+  "version": "1.5.0-rc.0",
   "description": "React bindings for @shopify/retail-ui-extensions",
   "publishConfig": {
     "access": "public",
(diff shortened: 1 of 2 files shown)


heltisace (commented):
We need to take 1.4.0 GA before doing this. Otherwise 1.4.0 GA version will include 1.5.0 changes. Example: https://github.com/Shopify/ui-extensions/commits/main/packages/retail-ui-extensions. 

I would recommend to wait for @NathanJolly to be back to make sure my assumption is right as he was the last person to do this.

js-goupil (approved):

vctrchu:
> We need to take 1.4.0 GA before doing this. Otherwise 1.4.0 GA version will include 1.5.0 changes. Example: https://github.com/Shopify/ui-extensions/commits/main/packages/retail-ui-extensions.
> 
> I would recommend to wait for @NathanJolly to be back to make sure my assumption is right as he was the last person to do this.

@NathanJolly and I confirmed that the tags that are associated with versions will dictate what is included in the release. This means we are safe to release any release candidates before a official release 👍🏻

PR #1327, opened 2023-09-08
Rename CustomerAccountCard to Card

lsit:
### Background

Part of https://github.com/Shopify/core-issues/issues/60197

### Solution

### 🎩

- ...

### Checklist


packages/customer-account-ui-extensions-react/src/components/Card/Card.ts
```diff
@@ -0,0 +1,9 @@
+import {Card as BaseCard} from '@shopify/customer-account-ui-extensions';
```
brianshen1990, line 1: we have already switched to the unstable version, do we still want to update it main?

brianshen1990: cc @lihao ?

lihaokx (approved):
LGTM

packages/customer-account-ui-extensions-react/src/components/Card/Card.ts
lsit: Cool then I think for now it's fine to update both `main` and `unstable`

PR #1330, opened 2023-09-08
Add sample comments for customer account navigation

brianshen1990:
### Background

Part of https://github.com/Shopify/core-issues/issues/60098

Add come sample comments for customer account navigation API. Most of the comments come from https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API directly. 

### Solution

### 🎩

- Navigation doc: https://shopify-dev.doc.brian-shen.us.spin.dev/docs/api/customer-account-ui-extensions/apis/navigation 
- CustomerAccountCard doc: https://shopify-dev.doc.brian-shen.us.spin.dev/docs/api/customer-account-ui-extensions/components/structure/customeraccountcard

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/api/standard-api/standard-api.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/standard-api/standard-api.ts
@@ -121,30 +121,67 @@ export interface Localization {
   extensionLanguage: StatefulRemoteSubscribable<Language>;
 }
 
+/**
+ * An enumerated value representing the type of navigation.
+ */
 export type NavigationType = 'push' | 'replace' | 'traverse';
 
 export interface NavigationOptions {
+  /**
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/src/surfaces/customer-account/api/standard-api/standard-api.ts
```diff
@@ -121,34 +121,75 @@ export interface Localization {
...
   updateCurrentEntry(options: {state: Record<string, any>}): void;
-};
+}
 
 export interface NavigateFunction {
+  /**
+   * Navigates to a specific URL, updating any provided state in the history entries list.
+   * @param url The destination URL to navigate to.
+   */
   (url: string, options?: NavigationOptions): void;
```
brianshen1990, line 193: unfortunately, seems this format ( interface itself is an function ) is not supported yet

lsit (approved):

PR #1333, opened 2023-09-12
Update B2B profile extension targets

brianshen1990:
### Background

Part of https://github.com/Shopify/core-issues/issues/57077 

Update B2B profile extension targets and full page extension target name. 
For location related extension targets, we will provide a `locationId` parameter. 

Figma: 
https://www.figma.com/file/5Op17XqUmqj8dCSI5RZvo6/Extensibility-Dev-Preview?type=design&node-id=1656-12268&mode=design&t=0T2ytBkSPUm64hj4-0 

Screenshot 2023-09-12 at 12 02 32

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/targets.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/targets.ts
@@ -70,6 +70,11 @@ export interface CustomerAccountExtensionTargets {
       FullPageApi,
     AllComponents
   >;
+  'customer-account.page.render': RenderExtension<
+    Omit<StandardApi<'customer-account.page.render'>, 'navigation'> &
+      FullPageApi,
+    AllComponents
+  >;
   'CustomerAccount::Returns::Initiate': RunExtension<
     StandardApi<'CustomerAccount::Returns::Initiate'> & {orderId: string},
     void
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/src/surfaces/customer-account/targets.ts
```diff
@@ -70,6 +70,11 @@ export interface CustomerAccountExtensionTargets {
       FullPageApi,
     AllComponents
   >;
+  'customer-account.page.render': RenderExtension<
+    Omit<StandardApi<'customer-account.page.render'>, 'navigation'> &
+      FullPageApi,
+    AllComponents
+  >;
```
lsit, line 77: Which extension target is this? A dynamic one that renders on every page?

lsit (approved):

lihaokx (approved):

packages/ui-extensions/src/surfaces/customer-account/targets.ts
robin-drexler: iirc that's replacing the full page extension target name

robin-drexler (approved):

packages/ui-extensions/src/surfaces/customer-account/targets.ts
brianshen1990: yep, and @anahita-m  and i will be working on that part

PR #1340, opened 2023-09-13
Add PaymentIcon to Checkout

lsit:
### Background

Resolves https://github.com/Shopify/checkout-web/issues/25226

- Adds PaymentIcon component to Checkout surface

### Solution

Copied the types and docs from private packages

### 🎩

https://shopify-dev.payment-icon.lianne-sit.us.spin.dev/docs/api/checkout-ui-extensions/unstable/components/media/paymenticon

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout/components/PaymentIcon/PaymentIcon.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/components/PaymentIcon/PaymentIcon.ts
@@ -450,6 +450,8 @@ export type PaymentMethod =
 export interface PaymentIconProps {
   /**
    * The name of the payment method.
+   *
+   * Check the list of available payment methods [here](/docs/api/checkout-ui-extensions/components/media/paymenticon#paymentmethod).
    */
   name: PaymentMethod | string;
 
--- a/packages/ui-extensions/src/surfaces/checkout/components/PaymentIcon/PaymentIcon.doc.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/components/PaymentIcon/PaymentIcon.doc.ts
@@ -14,6 +14,11 @@ const data: ReferenceEntityTemplateSchema = {
       description: '',
       type: 'PaymentIconProps',
     },
+    {
+      title: 'PaymentMethod',
+      description: '',
+      type: 'PaymentMethod',
+    },
   ],
   category: 'Components',
   subCategory: 'Media',


packages/ui-extensions/src/surfaces/checkout/components/PaymentIcon/PaymentIcon.ts
```diff
@@ -0,0 +1,481 @@
...
+  | 'rapid-transfer'
+  | 'neteller'
+  | 'cetelem'
+  | 'paid';
+
+export interface PaymentIconProps {
+  /**
+   * The name of the payment method.
+   *
+   * Check the list of available payment methods [here](https://github.com/activemerchant/payment_icons/tree/master/app/assets/images/payment_icons).
```
jffortier, line 454: Just noticed the link. We should point them to the `PaymentMethod` type no?

I don't believe the names of the icons in the the active_merchant repo matches the ones we accept here, so I'd remove this link to avoid any confusion. Developers should only refers to the list of PaymentMethod listed above.

jffortier (approved):

packages/ui-extensions/src/surfaces/checkout/components/PaymentIcon/PaymentIcon.ts
lsit: Removed the link. Do you know why the [docs](https://shopify-dev.payment-icon.lianne-sit.us.spin.dev/docs/api/checkout-ui-extensions/unstable/components/media/paymenticon#paymenticonprops) only show "string" even though the type is `PaymentMethod | string`?

lsit: Yep I asked about it in #core-build-learn and they were able to reproduce the issue and logged an issue on their end for it https://github.com/Shopify/generate-docs/issues/217

lsit: Can this be shipped as is and it'll be fixed eventually or should I wait for their fix? For context, I'm trying to get this component exposed for the 2023-10 release and the cutoff date is next week.

jffortier: Go ahead. The docs can be fixed after, or manually modified if possible.

lsit: Turns out that because `PaymentMethod` encompasses a list of strings, so as far as typescript is concerned, the type is now just "string" since it can be a list of the PaymentMethod's string or a string which ultimately results in a string. The #core-build-learn team's suggestion was to add the definition of PaymentMethod to the page, which I've done now.

PR #1353, opened 2023-09-18
[CustomerAccounts] Remove navigation extensions

oluwatimio:
### Background

Fixes https://github.com/Shopify/core-issues/issues/60546

### Solution

### 🎩

- ...

### Checklist


oluwatimio:
> We detected some changes in `packages/*/package.json` or `packages/*/src`, and there are no updates in the `.changeset` directory. If the changes are user-facing and should cause a version bump, run `yarn changeset` to track your changes and include them in the next release CHANGELOG. If you are making simple updates to repo configuration, examples, or documentation, you do not need to add a changeset.

Looks like we dont bump the versions in unstable?

lsit (approved):

PR #1372, opened 2023-09-21
add apis that were added after 2023-07

robin-drexler:
### Background

It seems like two new apis were added to checkout OSP extensions after `2023-07`.
Which means we need to also support these on the ODP.

This PR adds the typings to ensure that they are available when `2023-10` is being created. 
The actual implementation will happen in different PRs.

I copied types and comments over from checkout

```diff
export type CustomerAccountStandardApi<Target extends keyof ExtensionTargets> =
  Pick<
    StandardApi<Target>,
    | 'analytics'
    | 'appliedGiftCards'
    | 'appMetafields'
    | 'attributes'
    | 'buyerIdentity'
    | 'checkoutSettings'
+   | 'checkoutToken'
    | 'cost'
    | 'discountCodes'
    | 'discountAllocations'
    | 'extension'
    | 'extensionPoint'
    | 'i18n'
    | 'lines'
    | 'localization'
    | 'metafields'
    | 'note'
    | 'query'
    | 'sessionToken'
    | 'settings'
    | 'shippingAddress'
+   | 'billingAddress'
    | 'shop'
    | 'storage'
    | 'ui'
    | 'version'
  >;
```


packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
```diff
@@ -430,6 +430,14 @@ export interface OrderStatusApi<Target extends ExtensionTarget> {
    */
   checkoutSettings: StatefulRemoteSubscribable<CheckoutSettings>;
 
+  /**
+   * A stable id that represents the current checkout.
+   *
+   * Matches the `token` field in the [WebPixel checkout payload](https://shopify.dev/docs/api/pixels/customer-events#checkout)
+   * and the `checkout_token` field in the [Admin REST API Order resource](https://shopify.dev/docs/api/admin-rest/unstable/resources/order#resource-object).
+   */
+  checkoutToken: StatefulRemoteSubscribable<CheckoutToken | undefined>;
```
alxclark, line 439: I'm surprised this is a subscribable and could change during a checkout or an order 😬

robin-drexler: At this point, I'm not asking questions, it's strictly meant to be aligned with what already is in checkout:https://github.com/Shopify/ui-extensions/blob/a9401ccd78e2a05d5d2067084da8dc3ba79ca351/packages/ui-extensions/src/surfaces/checkout/api/standard/standard.ts#L482 

We can look at changing this after.

alxclark (approved):
API wise, the changes make sense, but we should update the comments to reflect that we're after a checkout.

packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
robin-drexler: it's interesting that it is called `stable` id tho. 😅

packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
```diff
@@ -430,6 +430,14 @@ export interface OrderStatusApi<Target extends ExtensionTarget> {
    */
   checkoutSettings: StatefulRemoteSubscribable<CheckoutSettings>;
 
+  /**
+   * id that represents the checkout used to create the order.
```
alxclark, line 434:
```suggestion
   * Identifier that represents the checkout used to create the order.
```

wdyt ?

packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
dnagoda: It's a subscribable due to implementation details. For checkout and TYPs it could have just been a static value (retrieved from Source), but for OSP we need to inspect the negotiated atom: https://github.com/Shopify/checkout-web/pull/25071/files#diff-e21b4cab4f8d453df09511fbef00fa4c2418e5643f5994090bd6ffb440e5e615R1

Id there's a better way to model this, we could make those changes in checkout-web and here. This is still an unstable api, so better to improve it now if there's a better path.

dnagoda (approved):

PR #1382, opened 2023-09-25
remove package-lock.json

robin-drexler:
We're currently using yarn in ci and elsewhere so I think we should only have yarn.lock around and not an additional package-lock.json.


MitchLillie (approved):
LGTM!

jamesvidler (approved):

PR #1388, opened 2023-09-27
Rename order action menu item ui extension

gracerivas:
### Background
Renaming was updated as specified in the following issue: 

https://github.com/Shopify/core-issues/issues/58559

### Checklist


robin-drexler (commented):
I think we should also rename this one: 

https://github.com/Shopify/ui-extensions/blob/rename-order-action/packages/ui-extensions/src/surfaces/customer-account/targets.ts#L144

oluwatimio (approved):

PR #1394, opened 2023-09-28
fix inline code bug

jamesvidler:
### Background

This fixes a mismatch where checkout UI extensions must be generated with version `0.10.13` and admin UI must be generated with `0.10.10` otherwise code within accordions don't show correctly.

This updates the `@shopify/generate-docs` package to a known stable release version of `0.10.13` that checkout UI extensions has been using for some time. This fixes a previous issue where code within an accordion would be inlined as opposed to using a code file. 

### Solution

Use files for code examples as intended and bump the version to `0.10.13` where this was fixed.

### 🎩

- ...

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,6 @@
     "prettier": "^2.8.0",
     "react": ">=18.0.0",
     "typescript": "^4.9.0",
-    "@shopify/generate-docs": "0.10.10"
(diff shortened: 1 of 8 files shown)


jamesvidler:
Closed in favor of https://github.com/Shopify/ui-extensions/pull/1395

PR #1397, opened 2023-09-29
latest from checkout-ui extensions

jamesvidler:
### Background

Brings the latest changes from the private `checkout-web` package.

### Solution
[Publish libraries](https://vault.shopify.io/page/Extension-Libraries~gkKh.md).

### 🎩
[Tophat the package](https://vault.shopify.io/page/Extension-Libraries~gkKh.md#user-content-2-tophat-the-packages)

### Checklist


packages/ui-extensions/src/surfaces/checkout/components/shared.ts
jamesvidler, line 1: @oliverigor should all of these changes be excluded from this file?

oliverigor: The spacing ones are correct, not too sure about the ones related to AutocompleteGroup.

@jffortier or @jun-shop might know something about it.

jffortier: `location` is an invalid group for the autocomplete property. This was removed [here](https://github.com/Shopify/checkout-web/pull/26312) by the team that introduced it.

This will create a breaking change if partners are using it right now. It should be marked as deprecated, at lease for one of the libraries version, and removed in the next one. This should give enough time to developers that uses it to change their code (if any, but we can't take a guess).

cc @RouzbehMajidi

jamesvidler: Ok, i'm going to exclude any changes to this file for this version since we don't want to introduce any breaking changes. If you want to add a deprecated message, that can be done in a separate PR.

packages/ui-extensions/src/surfaces/checkout/components/shared.ts
```diff
@@ -20,7 +20,7 @@ export interface Autocomplete {
   field: AutocompleteField;
 }
 
-export type AutocompleteGroup = 'shipping' | 'billing' | 'location';
```
rcaplanshopify: based on the conversation above, I think `location` should not be deleted here

rcaplanshopify (approved):
LGTM. I noted one change that should be reverted pertaining to the removal of `location`

jamesvidler:
> LGTM. I noted one change that should be reverted pertaining to the removal of `location`

Yep, we aren't changing that file right now.

NabeelAhsen (approved):

PR #1398, opened 2023-10-02
feat: create Card component

belalsj:
### Overview

Adds the new WC Card component as a ui-extension component

### 🎩

Currently unsure of how to tophat 😅

### Checklist


packages/ui-extensions/src/surfaces/admin/components/Card/Card.doc.ts
```diff
@@ -0,0 +1,42 @@
+import {ReferenceEntityTemplateSchema} from '@shopify/generate-docs';
```
Fionoble, line 1: I would rename this file to `Card.doc-skip.ts`. That way it doesn't get added to the next gen-docs command and we can settle on a mechanism to highlight gated components.

packages/ui-extensions/src/surfaces/admin/components/Card/Card.ts
```diff
@@ -0,0 +1,13 @@
+import {createRemoteComponent} from '@remote-ui/core';
+import {BackgroundProps} from '../shared';
+
+export interface CardProps extends BackgroundProps {
+  /**
+   * Adjust the padding of all edges.
+   *
+   * `true` applies a default padding that is appropriate for the component.
+   */
+  padding?: boolean;
```
Fionoble, line 10: Is it true by default?

Fionoble (approved):

packages/ui-extensions/src/surfaces/admin/components/Card/Card.ts
belalsj: It shouldn't. i reworded the comment above in case it was unclear

PR #1399, opened 2023-10-02
Hide PolicyModal from 3P Apps

renerbaffa:
### Background

Contributes to https://github.com/Shopify/core-issues/issues/51494

It is a requirement that Customer Account does not expose `PolicyModal` to any 3P app. This PR prevents the type definition of the `PolicyModal` from being exposed.

### Solution

This PR introduces a new tsconfig that prevents the typescript definition from being included in the bundle when building both the `ui-extensions` and `ui-extensions-react`. In order to achieve that, it is needed to add `/* @internal */` in the line before the export of the type definition.

PS: This approach does not prevent the component itself from being exposed, just the typescript definition of it. That is exactly what we want because Returns [uses it](https://github.com/Shopify/buyer-returns/blob/main/extensions/buyer-returns/src/RequestReturnPage/RequestReturnPage.tsx#L400) and it should be available to 1P apps. They will have to define it on their side.

PPS: `PolicyModal` in the `ui-extensions-react` package was importing the component from the `ui-extensions` packages, which is no longer exposing such component. Because of that, the prop types had to be duplicated in both packages. That is the tradeoff of hiding a whole component from any app.

### 🎩

- https://github.com/Shopify/buyer-returns/pull/292 should be merged
- Documentation with all explanation: https://github.com/Shopify/customer-account-web/pull/3076
- The typescript definition of the `PolicyModal` component is no longer there:

Screenshot 2023-10-02 at 12 10 56

Screenshot 2023-10-02 at 12 11 14

### Checklist

### DIFF
--- a/packages/ui-extensions-react/src/surfaces/customer-account/components/PolicyModal/PolicyModal.ts
+++ b/packages/ui-extensions-react/src/surfaces/customer-account/components/PolicyModal/PolicyModal.ts
@@ -1,11 +1,30 @@
-import {PolicyModal as BasePolicyModal} from '@shopify/ui-extensions/customer-account';
-import {
-  createRemoteReactComponent,
-  ReactPropsFromRemoteComponentType,
-} from '@remote-ui/react';
+import {createRemoteReactComponent} from '@remote-ui/react';
 
-export type PolicyModalProps = ReactPropsFromRemoteComponentType<
-  typeof BasePolicyModal
->;
+type PolicyType = 'refund';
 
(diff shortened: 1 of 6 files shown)


lsit (approved):

robin-drexler:
I know I'm a bit late now, but why are we not deleting the types here completely if no-one is supposed to use them? Am I missing something?

renerbaffa:
Customer Account Web and Returns are using this component specifically. The main advantage is that we get to keep one single source of truth for the component definition. Any 1P apps or projects that wants to use internal components will know what are the supported props and can recreate the types on their side based on that

PR #1414, opened 2023-10-03
[Map]: Publish Map, MapMarker and MapPopover

oliverigor:
### Background

- Publish `Map`, `MapMarker` and `MapPopover` component to unstable.
- [Docs PR](https://github.com/Shopify/shopify-dev/pull/37874)

### Checklist


packages/ui-extensions/docs/surfaces/checkout/staticPages/examples/.eslintrc.js
oliverigor, line 1: @jamesvidler I ran prepare-package-release from checkout-web and I can see some files like this being deleted, and other changes as well that I undone [here](https://github.com/Shopify/ui-extensions/pull/1414/commits/d0fa5fb777dc2b08654145f648852814caa35976) 

Did we forgot to remove/add this in checkout-web? and should I add all this files again ? (I believe the answer is yes since lint is failing all over the place haha)

jamesvidler: I will do a PR to fix what the current state of "checkout-web" should be. Hold tight.

jamesvidler:
Can you review this PR that handles the eslint changes? Then rebase when that ships?
https://github.com/Shopify/ui-extensions/pull/1418

shopi-dori (approved):
✅  Approving based on walkthrough sessions with @oliverigor 

🎩 ❓ My understanding is that there's not really an easy way to test this as a PR reviewer, so I haven't done any visual testing as part of this review, just code scanning. If there is a way to test, let me know.

jamesvidler:
Similarily, here is the PR https://github.com/Shopify/checkout-web/pull/26705 that should ship for `checkout-web` so you don't have these changes in your PR.

HasanArmstrong:
Hi everyone,

The onZoomChange, onPress, onDoublePress, onCenterChange, onBoundsChange - doesn’t seem to fire for me also.

Thanks for this component 👍

shopi-dori:
@patryk-smc Thanks so much for the feedback! Following up on a few more of your comments:

> It would be nice to be able to customize the clustering icon (we would like the color to stand out more and the icon to be slightly larger)

Would you be able to share more details on (or examples of) the kinds of customizations you're wanting for cluster markers? For example, am I getting these right?
- Want to increase/decrease the diameter of the cluster background circle
- Want to change the cluster icon (currently the count of contained markers) to a custom icon
- Want to change the size of the cluster icon (currently the count of contained markers) to fill more/less of the cluster background

> It would be nice to be able to display Popover in the fixed position instead of next to the marker. See this example: https://cln.sh/XyMNqx3q

Noted, thank you for the video! That helps a lot to see what you're looking for.

> Also wanted to ask if passing ReactNode as children of MapPopover is on a short-term roadmap? Without that, we would need to show selection details outside the map component which is not the greatest UX.

Yes, customizing the `MapPopover` content is on the short-term roadmap 😁 👍

ncardeli:
@HasanArmstrong @patryk-smc 
The map callbacks should be working now 👍

PR #1419, opened 2023-10-04
Reexport checkout's missing types

renerbaffa:
### Background

Contributes to https://github.com/Shopify/core-issues/issues/60358

I've compared the components and the types that Customer Account is exporting to what Checkout is exporting. The summary is:
- 2 components missing (`ConsentCheckbox` and `ConsentPhoneField`). Tickets will be created for them.
- Several types were not reexported

### Solution

This PR adds the missing exports (typescript definitions only) from Checkout to Customer Account so that both are in sync in terms of what is provided to extensions.
I also updated the order of the imports to match how Checkout is exporting things on their side so that it is easy to compare them again in the future, if needed.

### 🎩

- Green CI

### Checklist


packages/ui-extensions-react/src/surfaces/customer-account/components/shared-checkout-components.ts
```diff
@@ -1,109 +1,145 @@
...
+  type GridProps,
   GridItem,
+  type GridItemProps,
   Heading,
+  type HeadingProps,
   HeadingGroup,
+  type HeadingGroupProps,
   Icon,
+  type IconProps,
+  // type IconSource,
```
lsit, line 44: Will you uncomment these after https://github.com/Shopify/ui-extensions/pull/1420 is shipped?

lsit (approved):

PR #1424, opened 2023-10-05
Add badge component

MitchLillie:
### Background

⚠️ Do not merge until a new version of app-kit has been consumed in web ⚠️ 

Closes https://github.com/Shopify/app-ui/issues/470

### Solution

### 🎩

- Docs here: https://shopify-dev.docs2.mitch-lillie.us.spin.dev/docs/api/admin-extensions/components/titles-and-text/badge

### Checklist


packages/ui-extensions/src/surfaces/admin/components/Badge/Badge.ts
```diff
@@ -0,0 +1,22 @@
...
+  icon: IconName;
+  iconPosition?: 'start' | 'end';
+}
+
+interface BadgeNoIconProps {
+  icon: never;
+  iconPosition: never;
+}
+
+export type BadgeProps = BadgeBaseProps & (BadgeIconProps | BadgeNoIconProps);
```
MitchLillie, line 20: This doesn't look great in the generated docs. Looking into a better way to achieve the same result without the awkward union showing up in the docs.

packages/ui-extensions/src/surfaces/admin/components/Badge/Badge.doc.ts
```diff
@@ -0,0 +1,42 @@
...
+  type: '',
+  definitions: [
+    {
+      title: 'BadgeProps',
+      description: '',
+      type: 'BadgeProps',
+    },
+  ],
+  category: 'Components',
+  subCategory: 'Titles and text',
```
MitchLillie, line 19: "Titles and text" was the best idea for a subcategory I could come up with. Open to other suggestions.

packages/ui-extensions/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@shopify/ui-extensions",
   "version": "0.0.0-unstable",
   "scripts": {
-    "docs:admin": "sh ./docs/surfaces/admin/build-docs.sh \"admin\"",
```
Fionoble, line 5: Thanks for cleaning this up lol, that was from an experiment I was doing before summer editions

packages/ui-extensions/src/surfaces/admin/components/shared/index.ts
```diff
@@ -122,6 +122,15 @@ export interface MinMaxLengthProps {
...
 }
 
+export interface AccessibilityLabelProps {
+  /**
+   * A label that describes the purpose or contents of the element. When set, it will be announced
+   * to users using assistive technologies and will provide them with more context. When set, any
+   * children or `label` supplied will not be announced to screen readers.
+   */
+  accessibilityLabel?: string;
+}
```
Fionoble, line 132: Good change 👍

Fionoble (approved):

packages/ui-extensions/src/surfaces/admin/components/Badge/Badge.ts
MitchLillie: I don't think there's another way to represent this in typescript unfortunately. We have the same problem with `Image` docs

PR #1427, opened 2023-10-06
Fix type mismatch issue for documentations

brianshen1990:
### Background

Part of https://github.com/Shopify/core-issues/issues/60107 
A fix for https://github.com/Shopify/customer-account-web/pull/3094 

Mainly when we import form checkout parent folder directly, we load almost all the types from `surface/checkout` , which will cause many type conflicts while generating documentations. 

Update some import so that types are correct in documentations. 

### Solution

### 🎩

Green CI

### Checklist


oluwatimio (approved):
Thanks for taking a look into this Brian!

lihaokx (approved):

PR #1430, opened 2023-10-10
Add choicelist component

MitchLillie:
### Background

⚠️ Do not merge until a new version of app-kit has been consumed in web ⚠️ 

Closes https://github.com/Shopify/app-ui/issues/473

### 🎩

- Docs URL: https://shopify-dev.docs2.mitch-lillie.us.spin.dev/docs/api/admin-extensions/components/forms/choicelist

### Checklist

### DIFF
--- a/.changeset/heavy-rules-cheer.md
+++ b/.changeset/heavy-rules-cheer.md
@@ -1,0 +1,6 @@
+---
+'@shopify/ui-extensions': patch
+'@shopify/ui-extensions-react': patch
+---
+
+Add ChoiceList to admin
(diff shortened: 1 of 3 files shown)


packages/ui-extensions/src/surfaces/admin/components/ChoiceList/ChoiceList.ts
```diff
@@ -0,0 +1,12 @@
+import {createRemoteComponent} from '@remote-ui/core';
+import {InputProps} from '../shared';
+
+export interface ChoiceListProps
+  extends Pick<
+    InputProps<string | string[]>,
+    'value' | 'name' | 'onChange' | 'disabled' | 'error'
+  > {}
```
bratsos, line 8: I think we should add `defaultValue` and `readOnly` to `InputProps` and use them here

```typescript
/**
   * A `string` or `string[]` indicating the initial ids of selected choices
   * in an uncontrolled ChoiceList.
   * In case of `multiple: true` and an array use only the first value.
   */
  defaultValue?: T;
  
    /**
   * Whether the field is read-only.
   */
  readOnly?: boolean;
```

MitchLillie: And also `multiple` and `thisInstance`. I will add them

bratsos (approved):

PR #1437, opened 2023-10-12
[Time Picking] Created `unstable` package for time picking pattern docs

shopi-dori:
### Background

Closes https://github.com/Shopify/checkout-web/issues/26982

As part of the Time and Place project that also released the `Map` component, this PR introduces guidance for developers on how to implement time picking using using our existing [Component library](https://shopify.dev/docs/api/checkout-ui-extensions/unstable/components).

#### Sample of Changes
For more details, see https://github.com/Shopify/checkout-web/pull/26944

image

### 🎩

- Package testing was completed according to the instructions in our Extension Libraries guide. I used my [developer preview shop](https://dori-maloney-checkout-developer-preview.myshopify.com/) to verify the packages against a scaffolded app/extension containing the `Map`.
- Anyone wanting to test this PR will need to follow the same steps in the Extension Libraries guide, using this branch.

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout/components/View/View.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/components/View/View.ts
@@ -2,6 +2,7 @@ import {createRemoteComponent} from '@remote-ui/core';
 
 import {MaybeResponsiveConditionalStyle} from '../../style';
 import type {
+  BackgroundProps,
   BlockAlignment,
   BorderProps,
   InlineAlignment,
@@ -101,7 +102,8 @@ export interface Translate {
 }
 
 export interface ViewProps
-  extends BorderProps,
+  extends Pick<BackgroundProps, 'background'>,
+    BorderProps,
     CornerProps,
     SizingProps,
     SpacingProps,


.changeset/big-pets-drop.md
shopi-dori, line 1: I'm jealous of whoever got the `popular-eggs-remember` file name 😂

packages/ui-extensions-react/src/surfaces/checkout/hooks/tests/mount.tsx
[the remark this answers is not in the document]
jamesvidler: Discard this file change

packages/ui-extensions/src/surfaces/checkout/components/View/View.ts
jamesvidler, line 1: You should also discard changes to this file. For context, this is a change that shouldn't be shipped yet. @oliverigor has more information on this.

jamesvidler (changes requested):
A couple requests to discard some changes.

jamesvidler:
Also, we should be adding changesets for each feature this PR is shipping.

.changeset/big-pets-drop.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/ui-extensions': patch
+'@shopify/ui-extensions-react': patch
+---
+
+Added time picking pattern docs
```
jamesvidler, line 6: This will be rolled into the public changelog when this ships as part of a stable version. So, it should be in complete sentences. That being said, I actually don't think you need to add this change log because its just adding docs which doesn't need to be in the changelog. 

For other changes in here though that are adding hooks etc, should have their own change log.

packages/ui-extensions/src/surfaces/checkout/components/View/View.ts
shopi-dori: Thanks @jamesvidler . Will do on reverting 👍 @oliverigor is on vacation for several weeks, so if there's someone else I should talk to let me know; otherwise, I'll check with the Checkout Design System team.

.changeset/big-pets-drop.md
shopi-dori: > it should be in complete sentences

@jamesvidler Could you help me understand what qualifies as a complete sentence for these change logs? I'm looking at the other ones in `.changeset`, and all the ones I'm seeing so far follow a similar sentence fragment pattern. For example, see `.changeset/dirty-beds-hammer.md` which has "Added details about title/summary truncation to docs" for the patch.

packages/ui-extensions/src/surfaces/checkout/components/shared.ts
jamesvidler, line 1: This is the file that should be discarded (shared.ts)

.changeset/big-pets-drop.md
jamesvidler: You would just run `yarn changeset` multiple times, or manually create the markdown files. But I'd recommend using the command to generate the random names.

> I'm not familiar with the other features that are being rolled in here, so how would you recommend determining how many change logs there should be?

This is a tricky part of this process right now, but what we have done in the past is review the other changes and create ones that make sense. Or, reach out to the authors of those original commits in `checkout-web` and ask them.

To me, these changesets should probably look like:
[ui-extensions-react] [patch]
- Adds `usePaymentMethodAttributes`, `usePaymentMethodAttributeValues`, and `useApplyPaymentMethodAttributeChange` react hooks. 

[ui-extensions] [patch]
- Adds the `bankIdNumber` to the `PaymentOptionItemApi`.
- Adds `background` to `View`, `Grid`, `GridItem`, `Pressable`, `ScrollView`, `InlineStack`,  and `BlockStack` components.

packages/ui-extensions/src/surfaces/checkout/components/ScrollView/ScrollView.ts
```diff
@@ -24,7 +30,8 @@ export interface ScrollViewEvent {
 }
 
 export interface ScrollViewProps
-  extends BorderProps,
+  extends Pick<BackgroundProps, 'background'>,
```
jffortier, line 33: We should have the same thing for View as seen here. Its part of the same PR that 

https://github.com/Shopify/checkout-web/blob/main/packages/ui-extensions/src/surfaces/checkout/components/View/View.ts#L105

shopi-dori: Thanks @jffortier ! I'll add it back in 👍  I had removed that at @jamesvidler 's request, but I think he may have meant another file.

jamesvidler (approved):
LGTM!

packages/ui-extensions/src/surfaces/checkout/components/ScrollView/ScrollView.ts
jamesvidler: Ya that was my bad, commented on the wrong file.

shopi-dori: `View.ts` is added back in now ✅ Thanks @jffortier and @jamesvidler !

PR #1438, opened 2023-10-12
New function to add line item discounts in bulk

js-goupil:
### Background

As part of optimizations, we are adding some functions to allow transacting on the cart in bulk. We have seen several apps loop through data and make calls to add line item discounts which comes at a performance cost in redux.

### Solution

This is a new interface to add multiple line item discounts to multiple line items in one go.

### 🎩

- ...

### Checklist


packages/retail-ui-extensions/src/extension-api/cart-api/cart-api.ts
```diff
@@ -101,6 +107,16 @@ export interface CartApiContent {
     amount: string,
   ): Promise<void>;
 
+  /** Add a line item discounts to multiple line items
+   * @param uuid the uuid of the line item that should receive a discount
+   * @param type the type of discount applied (example: 'Percentage')
+   * @param title the title attributed with the discount
+   * @param amount the percentage or fixed monetary amount deducted with the discout
+   */
```
NathanJolly, line 115: Might benefit from a comment about what the `string` key is here

NathanJolly (approved):
Nothing to expose in the -react package, right?

PR #1442, opened 2023-10-17
ImageGroup API for CustomerAccount Surface

oleksandroliynyk:
### Background

### Solution

### 🎩

- ...

### Checklist


packages/ui-extensions/src/surfaces/customer-account/components/ImageGroup/ImageGroup.ts
```diff
@@ -0,0 +1,48 @@
...
+  loading?: boolean;
+
+  /**
+   * Images to be displayed.
+   *
+   */
+  images: RemoteFragment;
+
+  /**
+   * Total count of items.
```
renerbaffa, line 32: Worth to mention that if `totalItems` is less than the amount of images passed we are going to use that instead?

renerbaffa (approved):

packages/ui-extensions/src/surfaces/customer-account/components/ImageGroup/ImageGroup.ts
```diff
@@ -0,0 +1,48 @@
...
+  imagesAccessibilityLabel?: string;
+
+  /**
+   * Loading state of the component.
+   * @default false
+   */
+  loading?: boolean;
+
+  /**
+   * Images to be displayed.
```
renerbaffa, line 26: Do we need to mention that we expect elements of type `Image` and support only the 3 props here (`id`, `accessibilityDescription` and `source`)?

RenataZakauske (approved):
I agree with @renerbaffa comments as well. All the rest looks good 👍

packages/ui-extensions/src/surfaces/customer-account/components/ImageGroup/ImageGroup.ts
```diff
@@ -0,0 +1,48 @@
...
+  totalItems?: number;
+
+  /**
+   * Label for the total count indicator.
+   @default { singular: "item", plural: "items" }
+  */
+  itemName?: {
+    singular: string;
+    plural: string;
+  };
```
lsit, line 43: As per [this discussion](https://github.com/Shopify/core-issues/issues/60848#issuecomment-1779109291), we're going to remove this from the API

packages/ui-extensions/src/surfaces/customer-account/components/ImageGroup/ImageGroup.ts
```diff
@@ -0,0 +1,48 @@
...
+   * Loading state of the component.
+   * @default false
+   */
+  loading?: boolean;
+
+  /**
+   * Images to be displayed.
+   *
+   */
+  images: RemoteFragment;
```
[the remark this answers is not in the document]
jkeohan: It's been replaced with passing children

packages/ui-extensions/src/surfaces/customer-account/components/ImageGroup/ImageGroup.ts
jkeohan: Think so as well and have updated the description

robin-drexler (approved):

renerbaffa (approved):

PR #1448, opened 2023-10-18
set unstable to latest customer-account internal state

robin-drexler:
### Background

This PR moves back our `internal` apis to `unstable` in preparation for release.

It mainly ports over new components (like Card) and APIs that we have added, but also includes some type renamings caused by clashes in OrderStatus and StandardApi.

The PR also ensures that hooks that only work on the ODP properly throw errors if used elsewhere.

### 🎩

Here's what I did on the spinstance to make this 🎩 able.

inside ui-extensions

```
yarn build
```
To build the packages and make them consumable locally.

Then inside the customer account and extension repos

```
rm -rf node_modules/@shopify/{ui-extensions,ui-extensions-react} && cp -r ../ui-extensions/packages/{ui-extensions,ui-extensions-react} node_modules/@shopify/
```

So both are now using the locally built version of `ui-extension` packages.

I ran `type-check` inside in both repos, which succeeded.
It'll run again properly in CI after we merge this PR and update the version properly in package.json

You can see the still working extension here.

- Go to https://customer-accounts-ui-extension-dev.unstable-one.robin-drexler.us.spin.dev/extensions/dev-console
- Click on `customer-account.order.action.render`
- Login with `dev@shopify.com` https://identity.unstable-one.robin-drexler.us.spin.dev/services/mail
- Go to https://development-store-1.account.shopify.unstable-one.robin-drexler.us.spin.dev/orders/3
- You should see an extension running in the summary

### Checklist


packages/ui-extensions-react/src/surfaces/checkout.ts
```diff
@@ -50,10 +50,12 @@ export type {
   MapLocation,
   MaybeConditionalStyle,
   MaybeResponsiveConditionalStyle,
+  IconSource,
```
robin-drexler, line 53: Those seemed to be missing. Other non-prop exports of components , like `MapBounds` are exported here as well.

packages/ui-extensions-react/src/surfaces/checkout/errors.ts
```diff
@@ -18,6 +18,16 @@ export class ExtensionHasNoMethodError extends Error {
   }
 }
 
+export class ExtensionHasNoFieldError extends Error {
```
robin-drexler, line 21: A bunch of our hooks only work on the ODP. We're throwing this error when they are used outside of the ODP.

This is done so that we can keep the typings that we have. Alternatively, we'd have to mark all of them as optional which would be a breaking change for existing OSP/ODP extensions.

packages/ui-extensions-react/src/surfaces/customer-account/errors.ts
```diff
@@ -17,3 +17,13 @@ export class ExtensionHasNoMethodError extends Error {
     );
   }
 }
+
+export class ExtensionHasNoFieldError extends Error {
```
robin-drexler, line 21: A bunch of our hooks only work on the ODP. We're throwing this error when they are used outside of the ODP.

This is done so that we can keep the typings that we have. Alternatively, we'd have to mark all of them as optional which would be a breaking change for existing OSP/ODP extensions.

packages/ui-extensions-react/src/surfaces/customer-account.ts
```diff
@@ -5,8 +5,11 @@ export {
   isConditionalStyleWithDefault,
   Style,
 } from '@shopify/ui-extensions/customer-account';
+export {render, reactExtension} from './customer-account/render';
+
 export type {
   AccessibilityRole,
+  ActionExtensionApi,
```
robin-drexler, line 12: Added new APIs

lsit (approved):
Components and related types look good

packages/ui-extensions-react/src/surfaces/customer-account/hooks/attributes.ts
```diff
@@ -5,14 +5,21 @@ import type {
 
 import {useApi} from './api';
 import {useSubscription} from './subscription';
+import {ExtensionHasNoFieldError} from '../errors';
 
 /**
  * Returns the proposed `attributes` applied to the checkout.
  */
 export function useAttributes<
   Target extends RenderExtensionTarget = RenderExtensionTarget,
```
brianshen1990, line 14: hm... we limited it to only `RenderOrderStatusExtensionTarget` to strongly suggest that it should only be used on order status page, why do we change it back to RenderExtensionTarget as it will not put any type check error for targets outside order status page ?   https://github.com/Shopify/ui-extensions/blob/internal/packages/ui-extensions-react/src/surfaces/customer-account/hooks/attributes.ts#L24

robin-drexler: good catch!

PR #1454, opened 2023-10-19
Order status api docs

brianshen1990:
### Background

Part of https://github.com/Shopify/core-issues/issues/60109

added order status api related definition. Related Checkout side code example:  https://github.com/Shopify/checkout-web/blob/main/packages/ui-extensions/src/surfaces/checkout/api/docs.ts 

### Solution

### 🎩

- ...

### Checklist


robin-drexler (approved):
are we going to get a lot of these? 😅 

I'm currently in the process of moving `internal` back to `unstable` and I just wanna make sure I have an overview of what's coming so I can incorporate these changes

brianshen1990:
> are we going to get a lot of these? 😅
> 
> I'm currently in the process of moving `internal` back to `unstable` and I just wanna make sure I have an overview of what's coming so I can incorporate these changes

Just one file, exporting types for docs, so it won't break anything ( without this, customer account web ci won't break haha )

lihaokx (approved):

PR #1458, opened 2023-10-20
fix policy modal

robin-drexler:
### Background

Not stripping the internal fields caused clashes in an app. 

I missed moving this over. We already had this setting in internal: https://github.com/Shopify/ui-extensions/pull/1399/files#diff-256b5258d7b08eedfba647de7d28bae26f13f08d985b93f5c1ddb456644857d2R7

I also changed how the type is constructed. I have honestly no idea how/why this worked before 🤔 
We never combined the Component with its types. We're now passing the props so that the type is complete. 

🎩 will be done in said app, after merge.

### Checklist

### DIFF
--- a/packages/ui-extensions-react/src/surfaces/customer-account/components/PolicyModal/PolicyModal.ts
+++ b/packages/ui-extensions-react/src/surfaces/customer-account/components/PolicyModal/PolicyModal.ts
@@ -27,4 +27,7 @@ export interface PolicyModalProps {
   type: PolicyType;
 }
 
-export const PolicyModal = createRemoteReactComponent('PolicyModal');
+export const PolicyModal = createRemoteReactComponent<
+  'PolicyModal',
+  PolicyModalProps
+>('PolicyModal');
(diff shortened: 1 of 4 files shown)


packages/ui-extensions-react/src/surfaces/customer-account/components/PolicyModal/PolicyModal.ts
```diff
@@ -27,4 +27,7 @@ export interface PolicyModalProps {
   type: PolicyType;
 }
 
-export const PolicyModal = createRemoteReactComponent('PolicyModal');
+export const PolicyModal = createRemoteReactComponent<
```
robin-drexler, line 30: I have honestly no idea how/why this worked before 🤔
We never combined the Component with its types. We're now passing the props so that the type is complete.

packages/ui-extensions-react/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "outDir": "build/ts",
     "rootDir": "src",
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "stripInternal": true
```
robin-drexler, line 7: Missed this when moving back to unstable. We had in internal already.

But not for the non-react package. I've also added it here. I don't think it makes sense to only strip internals for react.

oluwatimio (approved):

brianshen1990 (approved):

PR #1463, opened 2023-10-24
Update Background type values

lsit:
### Background

This update is related to this https://github.com/Shopify/checkout-web/pull/26499 and got missed in https://github.com/Shopify/ui-extensions/pull/1437. Need these values for exposing the background prop in the customer account surface.

It should have been part of [this changeset](https://github.com/Shopify/ui-extensions/blob/lsit/fix-background-type/.changeset/big-pets-drop.md) so I don't think I need to add another. Let me know if I do.

### Solution

### 🎩

https://shopify-dev.checkout-web-api-docs-6bhk.lianne-sit.us.spin.dev/docs/api/checkout-ui-extensions/unstable/components/structure/view#viewprops-propertydetail-background

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout/components/shared.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/components/shared.ts
@@ -162,8 +162,11 @@ export type CornerRadiusDeprecated = 'tight' | 'loose';
 export interface BackgroundProps {
   /**
    * Adjust the background.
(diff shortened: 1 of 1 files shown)


shopi-dori (commented):
Thanks for catching this @lsit ! I did indeed revert the entire `shared.ts` on the previous PR.

PR #1465, opened 2023-10-24
Support storefront api for customer account api

brianshen1990:
### Background

Part of Shopify/core-issues#61073

Please also have a review of customer account web side PR https://github.com/Shopify/customer-account-web/pull/3246 

Support storefront api for standard customer account ui extension api 

### Solution

### 🎩

Please see the tophat here https://github.com/Shopify/customer-account-web/pull/3246 

### Checklist


alxclark (approved):
Looks good 👍🏼  Should we make the OrderStatus targets extend the standard api interface on unstable ? It would avoid having to duplicate the `query` typings here (although it will be a much bigger lift)

lihaokx (approved):
🎩  LGTM

PR #1467, opened 2023-10-24
fix(security): add CORS policy section to Admin UI Extensions docs

charlesdobson:
### Background

We're updating the domain for the `load.html` file in Web to the new `https://extensions.shopifycdn.com/`. If apps have a custom CORS policy in their server, they need to add this new domain to the allow list.

This change will be done in https://github.com/Shopify/web/pull/107284. However, we need to send out messaging to Partners first in case they have a different CORS policy already set.

### Solution

Add documentation for this for extension developers. This is consumed in `shopify-dev` in https://github.com/Shopify/shopify-dev/pull/38519.

### 🎩

https://shopify-dev.shopify-dev-9ncp.charlie-dobson.us.spin.dev/docs/api/admin-extensions#security

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/admin/staticPages/admin-extensions.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/admin/staticPages/admin-extensions.doc.ts
@@ -226,6 +226,13 @@ const data: LandingTemplateSchema = {
         ],
       },
     },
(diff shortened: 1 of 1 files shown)


AaronSpringut (approved):
LGTM

PR #1469, opened 2023-10-25
add transaction as an exported field

h74zhou:
### Background

Adding the `Transaction` as an exported type!

### Solution

### 🎩

NA

### Checklist


sorta-shopify (approved):
👍

PR #1472, opened 2023-10-26
Sirgalleto/add location modal for order routing extensions

sirgalleto:
### Background

### Solution

### 🎩

- ...

### Checklist


jbalsas (approved):
Didn't 🎩  (is this something we could do through `yarn link` or `yalc`, perhaps?), but code makes sense to me!

packages/ui-extensions/src/surfaces/admin/components/LocationList/LocationList.doc.ts
```diff
@@ -0,0 +1,41 @@
+import {ReferenceEntityTemplateSchema} from '@shopify/generate-docs';
```
sirgalleto, line 1: As we're not intended to show this components to developers we should remove the docs 

@olavoasantos do we have a way to test that the docs will be correctly built if we don't add any files? 

Or maybe preserve the `doc` but use any config to hide it?

packages/ui-extensions/src/surfaces/admin/api/order-routing-rule/order-routing-rule.doc.ts
```diff
@@ -0,0 +1,20 @@
+import {ReferenceEntityTemplateSchema} from '@shopify/generate-docs';
```
MitchLillie, line 1: Are these docs ready to ship? Once this change gets merged in to the `unstable` branch, someone might generate the docs and cause this API to be published.

jbalsas: Not at all 😓

MitchLillie: We should just remove these docfiles until we're ready to publish

packages/ui-extensions/src/surfaces/admin/api/order-routing-rule/order-routing-rule.ts
```diff
@@ -0,0 +1,28 @@
...
+  id: string;
+  priority?: number;
+  metafields: Metafield[];
+}
+
+export interface OrderRoutingRuleApi<ExtensionTarget extends AnyExtensionTarget>
+  extends StandardApi<ExtensionTarget> {
+  /* Utilities for translating content according to the current `localization` of the admin. */
+  i18n: I18n;
+  data: {
```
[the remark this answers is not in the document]
MitchLillie: I think you just need to add whatever methods/properties that plugin provides developers to the typedef here.

packages/ui-extensions/src/surfaces/admin/api/order-routing-rule/order-routing-rule.doc.ts
jbalsas: Will do! I think we weren't super sure what was the process here so this was a bit of a shot in the dark. If I understand correctly, we should:
- Commit the types so they get to unstable
- Work on the docs separately and merge them when ready

Would that be the expected workflow?

Is there a timeline as to when unstable becomes stable? Is there any sort of deadline we'd want to hit?

packages/ui-extensions/src/surfaces/admin/api/order-routing-rule/order-routing-rule.ts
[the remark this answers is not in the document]
jbalsas: Yeah, makes sense, I'll just prep over here but wait for it to land before pushing this!!

packages/ui-extensions/src/surfaces/admin/components/LocationList/LocationList.doc.ts
jbalsas: Done, @MitchLillie! 🙏

packages/ui-extensions/src/surfaces/admin/api/order-routing-rule/metafields.ts
```diff
@@ -0,0 +1,54 @@
...
+  'list.product_reference',
+  'list.rating',
+  'list.single_line_text_field',
+  'list.url',
+  'list.variant_reference',
+  'list.volume',
+  'list.weight',
+] as const;
+
+interface MetafieldChange {
```
jbalsas, line 46: @olavoasantos, @MitchLillie, there's still no easy way to expose this to the extensions (nor to web, for that matter).

I managed to resolve the type through some TS gymnastics like:

```ts
type ExtractElementType<T> = T extends readonly (infer ElementType)[]
  ? ElementType
  : never;

type ApiForRenderExtension = ReturnType<
  typeof useApi<'admin.settings.order-routing-rule.render'>
>;

type ExtractMethod<T extends keyof ApiForRenderExtension> =
  ApiForRenderExtension[T];

type SaveMetafields = ExtractMethod<'saveMetafields'>;

type MetafieldChange = ExtractElementType<Parameters<SaveMetafields>[0]>;
```

I like how the types are clean and properly derived from `useApi(TARGET)`, but wondering if we'd want to expose some of the helpers that already exist in here as well in order to make this slightly simpler...

Normally one wouldn't need this, but as things grow, types might be needed in files that do not call `useApi` making some of this necessary 🤔

MitchLillie: > I like how the types are clean and properly derived from useApi(TARGET)

Couldn't you just export and import these types directly? I know this would make them not-perfectly derived from useApi, but it would make the extension code simpler.

MitchLillie: Yep, makes sense. This would be a good enhancement. I created https://github.com/Shopify/ui-extensions/issues/1525

MitchLillie (approved):
Didn't 🎩, but code looks good. It would be good to have someone 🎩 before merge, I can take a look a bit later today.

**Note**: Please run `yarn changeset` as suggested in [this comment](https://github.com/Shopify/ui-extensions/pull/1472#issuecomment-1780588229). It's required to generate our changelog properly.

PR #1479, opened 2023-10-27
Remove duplicated fields from order status api

brianshen1990:
### Background

Remove duplicated fields for customer account standard api and order status api

Part of this ticket https://github.com/Shopify/core-issues/issues/61073 

### Solution

### 🎩

So with this branch,
- Both customer account web and customer account dev ui extension's type check still pass
- Extension still works 
  - open this in an incognito window https://customer-accounts-ui-extension-dev.fetch.brian-shen.us.spin.dev/extensions/dev-console 
  - click any of the extension targets
  - login as `dev@customer.com` and [this identity mailer](https://identity.fetch.brian-shen.us.spin.dev/services/mail/)
  - go to order details page 
  - order status page still has query in api even if it is removed from standard api. 

extension code
```
function BlockExtension() {
  const api = useApi<"customer-account.order-status.block.render">();
  function queryMe(){
    api.query(
      `query ($first: Int!) {
        products(first: $first) {
          nodes {
            id
            title
          }
        }
      }`,
      {
        variables: {first: 5},
      },
    )
      .then(console.log)
      .catch(console.error);
    }

  return <Button onPress={queryMe}>Query</Button>;
}
```

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
@@ -18,10 +18,7 @@ export interface Docs_OrderStatus_CostApi
   extends Pick<OrderStatusApi<any>, 'cost'> {}
 
 export interface Docs_OrderStatus_LocalizationApi
-  extends Pick<OrderStatusApi<any>, 'i18n' | 'localization'> {}
-
-export interface Docs_OrderStatus_DiscountsApi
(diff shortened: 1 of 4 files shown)


lihaokx (approved):
🎩  LGTM

renerbaffa (approved):

oluwatimio (approved):

PR #1485, opened 2023-10-31
Sunset price override discount

depl0y:
### Background

In POS we sunset the 'Price override' functionality, in [this PR](https://github.com/Shopify/pos-next-react-native/pull/24495). This discount type is no longer supported and needs to be removed from the UI extensions as well.

### Solution

Removing the discount type from the Cart API.

### 🎩

- n/a

### Checklist

- [ ] I have updated relevant documentation _(currently working on this)._


pjn82:
LGTM.
This is for `unstable` API version, when is the next stable version release? Do we need to do anything about that?

PR #1489, opened 2023-11-01
patch: deprecate DiscountType

NathanJolly:
### Background

PriceOverride is no longer supported. Code discounts are only supported at the Cart level.

### Solution

Deprecate existing DiscountType and create new, correct types

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/extension-api/cart-api/cart-api.ts
+++ b/packages/retail-ui-extensions/src/extension-api/cart-api/cart-api.ts
@@ -15,11 +15,20 @@ export interface CartApi {
   cart: CartApiContent;
 }
 
+/**
+ * @deprecated
+ * PriceOverride is no longer supported. Code discounts are only supported at the Cart level.
+ * Please migrate to using CartDiscountType and LineItemDiscountType as soon as possible.
+ */
 export type DiscountType =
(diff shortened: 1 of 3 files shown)


packages/retail-ui-extensions/src/extension-api/cart-api/cart-api.ts
```diff
@@ -35,7 +44,7 @@ export interface CartApiContent {
    * @param amount the percentage or fixed monetary amount deducted with the discount. Pass in `undefined` if using discount codes.
    */
   applyCartDiscount(
-    type: DiscountType,
+    type: DiscountType | CartDiscountType,
```
js-goupil, line 47: Should we just remove `DiscountType` from here? What's the risk if we do that? AFAIK their IDE is just gonna show a linter error, even if they keep using the old type it will still run. But I feel like an error showing up might prompt them to migrate to the new type.

NathanJolly: Ooh, yeah, I think that makes sense. Depends on your definition of "breaking", I think.

js-goupil (approved):
Looks great!!

PR #1502, opened 2023-11-06
Add shipping method list api for Checkout::ShippingMethods::RenderBefore/After

jonathanhamel4:
### Background

For https://github.com/Shopify/checkout-web/issues/27000

Adds a ShippingMethodListApi to the `Checkout::ShippingMethod::RenderBefore/After` extension points with the `deliveryGroupId` extended property. 

### Solution

Adding a new api to the ShippingMethod render before/after. A feature is coming up that will duplicate extensions in each groups when used in the 2024-01 or unstable Api. 

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/checkout-ui-extensions/src/extension-points.ts
+++ b/packages/checkout-ui-extensions/src/extension-points.ts
@@ -9,6 +9,7 @@ import type {
   StandardApi,
   CheckoutApi,
   OrderStatusApi,
+  ShippingMethodListApi,
 } from './api';
 
 type Components = typeof import('./components');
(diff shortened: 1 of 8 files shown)


packages/checkout-ui-extensions/src/extension-points.ts
```diff
@@ -222,15 +223,19 @@ export interface ExtensionPoints {
    * header and shipping method options.
    */
   'Checkout::ShippingMethods::RenderBefore': RenderExtension<
-    CheckoutApi & StandardApi<'Checkout::ShippingMethods::RenderBefore'>,
+    ShippingMethodListApi &
```
loic-d, line 226: Any reason not to stick to `ShippingMethodsApi` per [tech doc](https://docs.google.com/document/d/1NktKnffnHYk6EezZ5SRwZdt04mQy7I-swd4QssU0qbU/edit#heading=h.xxpf69ehyiln)?

PR #1508, opened 2023-11-07
feat: DatePicker

belalsj:
### Background

https://github.com/Shopify/app-ui/issues/562

### Checklist


packages/ui-extensions/src/surfaces/admin/components/DatePicker/DatePicker.doc.ts
```diff
@@ -0,0 +1,43 @@
...
+  subCategory: 'Forms',
+  defaultExample: {
+    image: 'datepicker-default.png',
+    codeblock: {
+      title:
+        'Use when merchants need to select a single day, multiple dates, or a range of dates',
+      tabs: [
+        {
+          title: 'React',
+          code: '../../../../../../ui-extensions-react/src/surfaces/admin/components/DatePicker/examples/basic-datepicker.example.tsx',
```
MitchLillie, line 28: I think you're missing this example file, and the rest of the React stuff

packages/ui-extensions/src/surfaces/admin/components/DatePicker/DatePicker.doc.ts
```diff
@@ -0,0 +1,43 @@
...
+    {
+      title: 'DatePickerProps',
+      description: '',
+      type: 'DatePickerProps',
+    },
+  ],
+  category: 'Components',
+  subCategory: 'Forms',
+  defaultExample: {
+    image: 'datepicker-default.png',
```
MitchLillie, line 21: I don't really know if we need this image in _this_ repo, but the other components have it

packages/ui-extensions/src/surfaces/admin/components/DatePicker/DatePicker.doc.ts
```diff
@@ -0,0 +1,43 @@
...
+    codeblock: {
+      title:
+        'Use when merchants need to select a single day, multiple dates, or a range of dates',
```
MitchLillie, line 24: I think this needs to be shorter to look good. For example look at Badge:

Screenshot 2023-11-07 at 9 41 37 AM

packages/ui-extensions/src/surfaces/admin/components/DatePicker/DatePicker.ts
```diff
@@ -0,0 +1,100 @@
...
+   * When year/month navigation is controlled you must store these values in state and
+   * reflect it back in the `yearMonth` prop.
+   */
+  onYearMonthChange?(yearMonth: {year: Year; month: Month}): void;
+}
+
+/**
+ * A date string using the simplified ISO 8601 format (`YYYY-MM-DD`)
+ */
+export type DateString = string;
```
MitchLillie, line 59: Is it worth restricting these further? For example https://blog.logrocket.com/handling-date-strings-typescript/

packages/ui-extensions/src/surfaces/admin/components/DatePicker/DatePicker.doc.ts
tiffanytse: We should be consistent about providing images for all components that have UI. This is not optional IMO.

tiffanytse (changes requested):
Is there a spin link for this; or a preview of how things would render on shopify.dev?

packages/ui-extensions/src/surfaces/admin/components/DatePicker/DatePicker.ts
belalsj: That's great! I like this .. I wonder if we should just implement it or present it in an RFC. Granted, it is not a breaking change

Fionoble (approved):

PR #1514, opened 2023-11-08
add docs for special buttons

robin-drexler:
### Background
See https://github.com/Shopify/customer-account-web/pull/3362 for context and 🎩  instructions

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
@@ -2,6 +2,7 @@ import {OrderStatusApi} from './order-status/order-status';
 import {StandardApi} from './standard-api/standard-api';
 import {CartLineItemApi} from './cart-line/cart-line-item';
 import {FullPageApi} from '../targets';
+import {ButtonProps} from '../components';
 
 export interface Docs_OrderStatus_MetafieldsApi
   extends Pick<OrderStatusApi<any>, 'appMetafields' | 'metafields'> {}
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
```diff
@@ -77,3 +78,48 @@ export interface Docs_Standard_QueryApi
...
+  > {}
+export interface Docs_Page_Button_SecondaryAction
+  extends Pick<
+    ButtonProps,
+    'onPress' | 'to' | 'loading' | 'loadingLabel' | 'disabled'
+  > {
+  /**
+   * A label used for buyers using assistive technologies. Needed because `children` passed to this component will be discarded.
+   */
+  accessibilityLabel: ButtonProps['accessibilityLabel'];
```
robin-drexler, line 100: One downside is that this messes with the order of props in the doc. the one you override always come last.

We could do something like:

```ts
export interface Docs_Page_Button_SecondaryAction {
  to: ButtonProps['to']
  onPress: ButtonProps['onPress']
  // etc..
}
```

But then we'd need to copy paste the comment for each prop from the original component over and over again.

packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
```diff
@@ -77,3 +78,48 @@ export interface Docs_Standard_QueryApi
...
+    | 'loading'
+    | 'loadingLabel'
+    | 'disabled'
+    | 'accessibilityLabel'
+  > {}
+
+export interface Docs_OrderActionMenu_Button
+  extends Pick<
+    ButtonProps,
+    'onPress' | 'loading' | 'loadingLabel' | 'disabled' | 'accessibilityLabel'
```
brianshen1990, line 117: Do we accept all of this? I think we only picked the
`text` `to` 
https://github.com/Shopify/customer-account-web/blob/main/app/foundation/extensibility/extensions/menu-items/useMenuItems.tsx#L59-L67

robin-drexler: I think we will, but I could remove it until we actually do and we'd add it once it's there for real.

packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
robin-drexler: I'll have a look at adding this to the host so we don't need to bump this package over and over again.

packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
robin-drexler: I'll check how much effort it is to add it to the host. I'm ok, if this PR here is open for a day or so

robin-drexler: It's not a lie anymore :D

packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
```diff
@@ -77,3 +78,49 @@ export interface Docs_Standard_QueryApi
...
+    | 'to'
+    | 'loading'
+    | 'loadingLabel'
+    | 'disabled'
+    | 'accessibilityLabel'
+  > {}
+export interface Docs_Page_Button_SecondaryAction
+  extends Pick<
+    ButtonProps,
+    'onPress' | 'to' | 'loading' | 'loadingLabel' | 'disabled'
```
lsit, line 95: I don't think we're adding `loading`, `loadingLabel`, `disabled` to the secondaryAction (i.e. back button).

kirakik (approved):

packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
robin-drexler: good catch! fixed.

lsit (approved):

PR #1522, opened 2023-11-10
Keep generated docs

jamesvidler:
### Background

Related to https://github.com/Shopify/ui-extensions-private/issues/2088. This commits the generated docs to source control so that this associated chron job to open a PR on shopify-dev runs to update the docs.

### Solution

Don't ignore the files, and generate the latest one.

### 🎩

We'll need to tophat with the shopify-dev PR that gets created to ensure this is working as we expect.

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,6 @@ packages/*/*.js
 packages/*/bin
 !packages/*/.eslintrc.js
 *.log
-packages/ui-extensions/docs/**/generated
 
 # IntelliJ IDE ignore
 .idea
--- a/packages/ui-extensions/docs/surfaces/checkout/generated/generated_static_pages.json
+++ b/packages/ui-extensions/docs/surfaces/checkout/generated/generated_static_pages.json
@@ -1,0 +1,912 @@
+[
+  {
+    "title": "Configuration",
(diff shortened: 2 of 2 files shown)


oliverigor (approved):
LGTM

lemonmade (changes requested):
Doesn't it seem bad to have a 140,000 line, automatically-generated file in source control? Is there really no way to generate this file on-the-fly, when attempting to sync the changes between repos?

jamesvidler:
> Doesn't it seem bad to have a 140,000 line, automatically-generated file in source control? Is there really no way to generate this file on-the-fly, when attempting to sync the changes between repos?

It's not as bad as I first thought. The file itself is ~10mb, but I like the idea of guaranteeing a snapshot of the generated docs that can be tied to "this committed source code".

One way or another that file needs to live in source control. Up to this point, that has only lived in shopify-dev. The reason this file is necessary to commit into source control is so that the shopify-dev nightly process can find it and open a new PR for the changes. 

Alternatively, we could opt-out of the ui-framework doc gen automation and build our own pipeline. This would be something where we would open our own shopify-dev PR with the output of the build docs command without committing anything to source control. A bit more work involved in this approach though.

jamesvidler:
> @jamesvidler I think we should take the slightly longer way to achieve this. We can't let it become some common practice for teams to leave huge, uneditable messes of files in their repos. If one team is already doing it, us doing it will be the thing that makes this a normalized behavior. Let's choose a cleaner behavior to model, one that pushes the docs changes as code merges into this repo.

Fair enough. I'll consider an alternative approach when at a later time.

PR #1537, opened 2023-11-17
adds shipping option list api

jonathanhamel4:
### Background

Adds a new api `ShippingOptionListApi` to the `purchase.checkout.shipping-option-list.render-before/after` extension points.

Adds `id?: string` to the deliveryGroups api. 

Both ids have a potentially undefined type as this is the graphql type we are receiving in checkout-web. 

It seems like all the examples are auto-generated. Should we be adding new examples for this? 

### Checklist

### DIFF
--- a/.changeset/breezy-guests-protect.md
+++ b/.changeset/breezy-guests-protect.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Adding the ShippingOptionListApi and a DeliveryGroup.id in the StandardApi.


.changeset/breezy-guests-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Adding the ShippingOptionListApi and a new id to the DeliveryGroup in the StandardApi.
```
loic-d, line 5: nit: should we write `DeliveryGroup.id` instead of "a new id to..."?

packages/ui-extensions/src/surfaces/checkout/api/shipping/shipping-option-list.ts
```diff
@@ -0,0 +1,6 @@
+export interface ShippingOptionListApi {
+  /**
+   * The delivery group id the item is attached to
```
loic-d, line 3: Do we usually use `item`? It doesn't feel super description. Maybe: "The unique identifier of the delivery group the extension is referencing"?

loic-d (approved):
API changes are aligned with what we discussed 👍

jonathanhamel4:
> It looks like these changes have not shipped in `checkout-web` yet. You should open a PR to change the [private package](https://github.com/Shopify/checkout-web/tree/main/packages/ui-extensions/src/surfaces/checkout) there prior to including it in a public release.

Yes, there's a PR open for that already. https://github.com/Shopify/checkout-web/pull/28099

jonathanhamel4:
> Since this is adding a new API and changes the APIs for multiple targets there are a few doc related changes required.
> 
> 1. Define a new Doc API specification (eg.`SHIPPING_OPTION_LIST_API`), similar to [CART_LINE_ITEM_API](https://github.com/Shopify/checkout-web/blob/2d79b2ad0fa8687c4324c9471328473035ff26ba/packages/ui-extensions/docs/surfaces/checkout/reference/helper.docs.ts#L640)
> 2. Modify the existing, relevant [target](https://github.com/Shopify/checkout-web/tree/main/packages/ui-extensions/docs/surfaces/checkout/reference/targets) pages to use this `SHIPPING_OPTION_LIST_API`
> 
> If you have any questions about this, let me know and i'd be happy to walk you through it.

I'll look at adding those to my existing PR. Thanks for the info.

PR #1569, opened 2023-11-28
Remove location api

oluwatimio:
### Background

Fixes https://github.com/Shopify/core-issues/issues/62603

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/targets.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/targets.ts
@@ -1,4 +1,3 @@
-import {StatefulRemoteSubscribable} from '@remote-ui/async-subscription';
 import {AnyComponent} from '../checkout/shared';
 
 import {CartLineItemApi} from './api/cart-line/cart-line-item';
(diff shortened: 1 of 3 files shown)


robin-drexler (approved):
type checks are complaining, but it's only an unused import

packages/ui-extensions/src/surfaces/customer-account/targets.ts
```diff
@@ -257,13 +257,6 @@ export type AllowedComponentsForRenderExtension<
 > = ExtractedAllowedComponentsFromRenderExtension<RenderExtensions[Target]>;
 
 export interface FullPageApi {
-  /**
-   * @deprecated This is deprecated. DO NOT USE. This will be removed soon without warning. use `api.navigation.currentEntry` and `api.navigation.addEventListener` instead.
-   */
-  location: StatefulRemoteSubscribable<{
-    pathname: string;
-    search: string;
-  }>;
```
brianshen1990, line 266: also here?  https://github.com/Shopify/ui-extensions/blob/6376b6f175e889bcca59c53ce17ab31321a185d1/packages/ui-extensions/src/surfaces/customer-account/api/docs.ts#L80

oluwatimio: Actually no wouldnt need to do that since we omit location

PR #1581, opened 2023-12-01
Add POS UI Extension react package

js-goupil:
### Background

This PR brings in the react bindings for the POS UI extension components, and the convenience react hooks to the unified package.

### Solution

It's pretty much a straight port, but I also changed the file structure to use what admin and checkout are using. 

It's likely we may have missed a few exports from the vanilla package, but we'll go back and do a final run through to make sure everything is included in a clean up PR.

### 🎩

- ...

### Checklist


packages/ui-extensions-react/src/surfaces/point-of-sale.ts
```diff
@@ -0,0 +1,5 @@
+export {extend, extension} from '@shopify/ui-extensions/point-of-sale';
```
agorovyi, line 1: Looking at [checkout example](https://github.com/Shopify/ui-extensions/blob/18efa584c8a3bd6f1b56620152fe74aa6ea99ac5/packages/ui-extensions-react/src/surfaces/checkout.ts#L65) I think we should also export all individual types from here.

packages/ui-extensions/src/surfaces/point-of-sale/components.ts
```diff
@@ -1,29 +1,34 @@
 export {Badge} from './components/Badge/Badge';
-export type {BadgeProps} from './components/Badge/Badge';
+export type {BadgeProps, BadgeVariant} from './components/Badge/Badge';
```
js-goupil, line 2: @agorovyi That was a solid suggestion. Exporting the types from the point-of-sale.ts file exposed all the types that were not being exported from the ui-extensions package, so we caught a lot of potential errors haha.

PR #1599, opened 2023-12-12
Expose banner overlay props in camera scanner

bryankou:
### Background

Part of https://github.com/Shopify/pos-next-react-native/issues/28204

Exposing the props necessary to allow banner overlay for camera scanners, based on @vctrchu's suggestions in the ticket. 

### Checklist

### DIFF
--- a/packages/retail-ui-extensions-react/package.json
+++ b/packages/retail-ui-extensions-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/retail-ui-extensions-react",
-  "version": "1.5.1",
+  "version": "1.6.0-rc.0",
   "description": "React bindings for @shopify/retail-ui-extensions",
   "publishConfig": {
     "access": "public",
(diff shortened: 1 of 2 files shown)


packages/retail-ui-extensions/src/components/CameraScanner/CameraScanner.ts
```diff
@@ -1,6 +1,22 @@
...
+
+export interface Information {
+  message: string;
+  bannerType?: BannerType;
+}
+
+export interface CameraScannerProps {
+  barcodeStatus?: Information;
+  showBanner?: boolean;
+}
```
rtymchyk, line 19: I took a look at POS again, there's some other props like this used on the scnaner.
```
   onBannerDismiss?: () => void; 
   overlayClickableText?: string; 
   hideOverlayClickableText?: boolean; 
```
   
Is `overlayClickableText` the "Dismiss" action? Does it make sense to expose some/all of these while we're at it?

rtymchyk (approved):
I had a suggestion to expand the API, but not a blocker - looks fine to me!  I'd wait for @vctrchu to give us his blessing 😄

PR #1626, opened 2024-01-02
Update camera scanner props names

bryankou:
### Background

Part of https://github.com/Shopify/pos-next-react-native/issues/28204
Updated the names to more closely match we currently do with Polaris, based on @NathanJolly and @vctrchu's suggestion.

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/components/CameraScanner/CameraScanner.ts
+++ b/packages/retail-ui-extensions/src/components/CameraScanner/CameraScanner.ts
@@ -8,14 +8,14 @@ export type BannerType =
   | 'Critical'
   | 'Default';
 
-export interface Information {
-  message: string;
-  bannerType?: BannerType;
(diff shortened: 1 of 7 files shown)


vctrchu (approved):

packages/retail-ui-extensions/src/components/CameraScanner/CameraScanner.ts
```diff
@@ -8,14 +8,14 @@ export type BannerType =
   | 'Critical'
   | 'Default';
 
-export interface Information {
-  message: string;
-  bannerType?: BannerType;
+export interface CameraScannerBannerProps {
+  title: string;
+  variant?: BannerType;
```
NathanJolly, line 13: I believe this should be of type `BannerVariant` and then you can delete `BannerType`

PR #1632, opened 2024-01-04
validation settings extension target

cpeddecord:
### Background

Pending approval of https://github.com/Shopify/ui-api-design/pull/197, adds APIs for the validation settings UI extension target.

### Checklist


packages/ui-extensions/src/surfaces/admin/api/validation-settings/metafields.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+type MetafieldsChange =
+  | MetafieldUpdateChange
+  | MetafieldRemoveChange
+  | MetafieldUpdateChange[]
+  | MetafieldRemoveChange[];
+
+export type SupportedDefinitionType = typeof supportedDefinitionTypes[number];
+export type ApplyMetafieldsChange = (changes: MetafieldsChange[]) => void;
```
jamesvidler, line 67: Wouldn't this be exporting the same type as the existing order routing metafield?

cpeddecord: yeah, I don't know where best to locate this. `/src/surfaces/admin/api/shared.ts` would be my guess so it doesn't get slurped up into the StandardApi as yet but still not sure. Thought it best to remain duplicate for the moment.

jamesvidler:
Just an FYI, this should ship into `unstable` before it is considered as a patch for `2024-01`.

cpeddecord:
closed in favor https://github.com/Shopify/ui-extensions/pull/1681

PR #1654, opened 2024-01-15
[UI components]: Update style helper docs for unstable

oliverigor:
### What problem are you trying to solve? 💡

complete https://github.com/Shopify/checkout-web/issues/23677 fix, that wasn't included in all versions

### How are you solving it? 🛠️

- For `NonPresentationalAccessibilityRole` I took JF's suggestion and just duplicated a bit of code since generate-docs cannot handle `Exclude` property right, and to avoid future errors I opted to not use Pick property as well.
- For conditional styles I end up simplifying some of the types along with @ncardeli, and created a simplified version for docs only, that way we were able to avoid the infinite loop when trying to access the property on the docs.

#### Spin Instance 🌀

https://shopify-dev.conditional-styles-doc-fixes.igor-deoliveiramartins.us.spin.dev/docs/api/checkout-ui-extensions

Before
Screenshot 2023-12-06 at 11 16 01 AM
Screenshot 2023-12-06 at 11 18 32 AM

After
Screenshot 2024-01-15 at 12 00 15 PM

Screenshot 2023-12-06 at 11 19 04 AM
-8ad4-d909d41d3da3">

### Checklist

### DIFF
--- a/.changeset/sharp-houses-rest.md
+++ b/.changeset/sharp-houses-rest.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+Add docs fixes for the conditional style Style type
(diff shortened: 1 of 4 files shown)


jamesvidler (approved):
Looks good 👍. You'll just need to add a changeset for this as a 'patch'.

ncardeli (approved):
:shipit:

PR #1681, opened 2024-01-26
Validation settings Admin UI Extension target

cpeddecord:
### Background

Resolves https://github.com/Shopify/checkout-web/issues/18126

spec'ed here https://github.com/Shopify/ui-api-design/pull/197

_tl;dr: a task-specific extension target that configures only validation functions._

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/api/checkout-rules/launch-options.ts
+++ b/packages/ui-extensions/src/surfaces/admin/api/checkout-rules/launch-options.ts
@@ -1,0 +1,14 @@
+export interface ValidationData {
+  validation?: {
+    /**
+     * the validation's gid when active in a shop
+     */
+    id: string;
+  };
+  shopifyFunction: {
+    /**
+     * the validation's unique identifier
+     */
+    uuid: string;
+  };
+}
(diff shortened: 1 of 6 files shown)


packages/ui-extensions/src/surfaces/admin/api/checkout-rules/launch-options.ts
jamesvidler, line 1: Just a general question, why is this in a `launch-options.ts` file separate form `validation-settings.ts`, when it appears to be specific to the validation api?

jamesvidler (approved):
Looks good to me. Don't forget to [add a changeset](https://github.com/Shopify/ui-extensions/blob/unstable/documentation/versions-and-deploys.md#adding-code) (minor change), otherwise nothing will get published to the unstable package.

Has anyone from the Admin team reviewed this PR? @MitchLillie might be able to help with that.

packages/ui-extensions/src/surfaces/admin/api/checkout-rules/launch-options.ts
cpeddecord: There are a number of different checkout rules coming soon, order submission may fall into this general class of admin ui extensions, for example. Validation settings is one of this class of target. Happy to do single target/single directory here.

PR #1685, opened 2024-01-29
expose authenticated account related hooks

brianshen1990:
### Background

expose authenticated account related hooks

### 🎩

With the local package, we can get the hook correctly. 

Screenshot 2024-01-29 at 12 29 03

### Checklist


lihaokx (approved):
LGTM

PR #1686, opened 2024-01-29
Fix Admin Extensions docgen

igor10k:
### Background

There are some issues with generating docs, mainly inconsistencies with the casing which causes the script to fail on case sensitive file systems.

### Solution

Fix the casing.

While at it I also copied from the checkout surface script that copies the generated files to their rightful place.

### Checklist


packages/ui-extensions/docs/surfaces/admin/build-docs.sh
```diff
@@ -1,25 +1,71 @@
...
+fail_and_exit() {
+  echo "** Failed to generate docs"
+  echo "See https://vault.shopify.io/page/Extension-Docs~SkgE.md"
+  exit $1
+}
+
+if [ -z $API_VERSION ]
+then
+  API_VERSION="unstable"
+  echo "Building docs for 'unstable' checkout UI extensions API. You can add a calver version argument (e.g. 'yarn docs:checkout 2023-07') to generate the docs for a stable version."
```
MitchLillie, line 16:
```suggestion
  echo "Building docs for 'unstable' admin UI extensions API. You can add a calver version argument (e.g. 'yarn docs:admin 2023-07') to generate the docs for a stable version."
```

packages/ui-extensions/docs/surfaces/admin/build-docs.sh
```diff
@@ -1,25 +1,71 @@
...
+else
+  echo "Building docs for '$API_VERSION' checkout UI extensions API."
```
MitchLillie, line 18:
```suggestion
  echo "Building docs for '$API_VERSION' admin UI extensions API."
```

MitchLillie (commented):
Only issues are some small text changes and waiting for this PR to land before generating: https://github.com/Shopify/ui-extensions/pull/1687

Otherwise all the docs changes look great, thanks!

packages/ui-extensions/docs/surfaces/admin/build-docs.sh
```diff
@@ -1,25 +1,71 @@
...
+      ~/src/github.com/Shopify/shopify-dev/db/data/docs/templated_apis/admin_extensions/$API_VERSION/generated_docs_data.json
+    sed_exit=$?
+    if [ $sed_exit -ne 0 ]; then
+      fail_and_exit $sed_exit
+    fi
+    rsync -a --delete ./$DOCS_PATH/screenshots/ ~/src/github.com/Shopify/shopify-dev/app/assets/images/templated-apis-screenshots/admin-extensions/$API_VERSION
+
+    cd ~/src/github.com/Shopify/shopify-dev
+
+    echo "Docs: https://$SPIN_SHOPIFY_DEV_SERVICE_FQDN/docs/api/admin-ui-extensions"
```
MitchLillie, line 65:
```suggestion
    echo "Docs: https://$SPIN_SHOPIFY_DEV_SERVICE_FQDN/docs/api/admin-extensions"
```

packages/ui-extensions/docs/surfaces/admin/build-docs.sh
```diff
@@ -1,25 +1,71 @@
...
+fi
+
+if [ -n "$SPIN" ]; then
+  if [ -n "$SPIN_SHOPIFY_DEV_SERVICE_FQDN" ]; then
+
+    mkdir -p ~/src/github.com/Shopify/shopify-dev/db/data/docs/templated_apis/admin_extensions/$API_VERSION
+    cp ./$DOCS_PATH/generated/* ~/src/github.com/Shopify/shopify-dev/db/data/docs/templated_apis/admin_extensions/$API_VERSION
+    # Replace 'unstable' with the exact API version in relative doc links
+    sed -i \
+      "s/\/docs\/api\/admin-ui-extensions\/unstable/\/docs\/api\/admin-ui-extensions\/$API_VERSION/gi" \
```
MitchLillie, line 55: I'm not sure on this piece, but I think this should match the change on L65

igor10k: This one is actually not required in admin as that kind of links is not getting generated, I kept it only to stay in sync with the checkout script that ideally should be extracted somewhere and used both for admin and checkout docs generation. You are right though that the path should be updated.

PR #1696, opened 2024-01-31
Publish `@shopify/web-pixels-extension@0.5.0`

sorta-shopify:
### Background
Closes https://github.com/Shopify/ce-customer-behaviour/issues/3379
Bumps the version and publishes the changes done in https://github.com/Shopify/ui-extensions/pull/1691

### Solution
Went through the normal version bump process outlined [here](https://github.com/Shopify/web-pixels-manager/blob/main/help-docs/updating-events.md#updating-ui-extensions)

A minor version felt appropriate, but happy to do a patch instead if people would prefer.

### 🎩

Tag is at: https://github.com/Shopify/ui-extensions/releases/tag/%40shopify%2Fweb-pixels-extension%400.5.0

### Checklist

### DIFF
--- a/packages/web-pixels-extension/package.json
+++ b/packages/web-pixels-extension/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@shopify/web-pixels-extension",
   "description": "Provides tools to author Web Pixels extension",
-  "version": "0.4.22",
+  "version": "0.5.0",
   "publishConfig": {
     "access": "public",
     "@shopify:registry": "https://registry.npmjs.org"


packages/web-pixels-extension/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@shopify/web-pixels-extension",
   "description": "Provides tools to author Web Pixels extension",
-  "version": "0.4.22",
+  "version": "0.5.0",
```
h74zhou, line 4: Looks good! Seems like a true first major update in a while.

BTW, so we're good with [exposing the meta type](https://github.com/Shopify/ui-extensions/pull/1691/files#diff-fddd73774ed30e13140b787531539f00abec07bf666e9b89d08a1a135d8c8fc1R457) right?

h74zhou: Also, I thought we were going to [remove this line from the PR](https://github.com/Shopify/ui-extensions/pull/1691/files#diff-250cae8251833df357522e5b9350bf587e1f1833f971fd51ab0e66df0e6b3e6aR7)? 

Maybe Github is bugging out, but it seems like it was merged?

antoinegrant: We're ok with shipping this as is for now, the types in question are not used. We have an issue to omit certain types in the future.

antoinegrant (approved):

h74zhou (approved):

PR #1712, opened 2024-02-02
Add a constellation for building dev docs

MitchLillie:
### Background

After [this PR](https://github.com/Shopify/ui-extensions/pull/1686), we can build our ui-extensions docs in spin. But first we need a constellation to use :)

### Solution

### 🎩

- ...

### Checklist


elanalynn (approved):

MitchLillie:
Wrong target branch

PR #1721, opened 2024-02-08
Deprecating SmartGridAPI and adding ActionAPI

BrittaEvansFenton:
Resolves part of https://github.com/Shopify/pos-next-react-native/issues/31855

### Background
We are updating the APIs and Targets to be more generic for future use, and this PR replaces the existing SmartGridAPI with a generic ActionAPI by deprecating SmartGridAPI and adding an identical ActionAPI

### 🎩
-

### Checklist

- ~[ ] I have :tophat:'d these changes~
- ~[ ] I have updated relevant documentation~

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/action-api/action-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/action-api/action-api.ts
@@ -11,5 +11,5 @@ export interface ActionApiContent {
  * Access the Action API to present your app in a full screen modal.
  */
 export interface ActionApi {
-  Action: ActionApiContent;
+  action: ActionApiContent;
 }


packages/ui-extensions/src/surfaces/point-of-sale/api/smartgrid-api/smartgrid-api.ts
```diff
@@ -1,3 +1,7 @@
+
+/** * 
+ * @deprecated Use `ActionApiContent` instead.
```
js-goupil, line 3:
```suggestion
 * @deprecated Use `ActionAPI` instead.
```

Just changing the wording here, since from the interface perspective, partners are going to see "actionAPI". ActionAPIContent is just the type 😄

BrittaEvansFenton:
missed the export added here: https://github.com/Shopify/ui-extensions/pull/1749

PR #1725, opened 2024-02-09
Remove misleading types

sorta-shopify:
### Background
Closes https://github.com/Shopify/ce-customer-behaviour/issues/4213

Removing some types that may be misleading.

### Solution

Removed the relevant types

### Checklist

### DIFF
--- a/packages/web-pixels-extension/src/types/index.ts
+++ b/packages/web-pixels-extension/src/types/index.ts
@@ -12,6 +12,7 @@ export type {
   DiscountApplication,
   GenericElement,
   InputElement,
+  FormElement,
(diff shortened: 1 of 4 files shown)


packages/web-pixels-extension/src/types/index.ts
```diff
@@ -12,6 +12,7 @@ export type {
   DiscountApplication,
   GenericElement,
   InputElement,
+  FormElement,
```
sorta-shopify, line 15: This export was added after DOM Events launch, so is expected.

emileber (approved):

PR #1746, opened 2024-02-15
[UI components]: Sheet release

oliverigor:
### Background

Release `Sheet` component.

Docs PR https://github.com/Shopify/shopify-dev/pull/41409

### Solution

Running `prepare-package-release`

### 🎩

Docs generated from this repo: https://shopify-dev.sheet-release.igor-deoliveiramartins.us.spin.dev/docs/api/checkout-ui-extensions/unstable/components/overlays/sheet

### Checklist


packages/ui-extensions/src/surfaces/customer-account.ts
```diff
@@ -5,4 +5,5 @@ export * from './customer-account/api';
 export * from './customer-account/components';
 
 export * from './checkout/shared';
-export * from './checkout/style';
+export * from './checkout/style/types';
+export * from './checkout/style/style';
```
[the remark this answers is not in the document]
oliverigor: Here's the PR for Docs 

https://github.com/Shopify/shopify-dev/pull/41409

packages/ui-extensions/src/surfaces/checkout/targets.ts
jamesvidler, line 1: There might be a problem with this, I don't think we should be removing the `OrderConfirmationApi`. Can you track down the author of this commit and to confirm?

packages/ui-extensions/src/surfaces/customer-account.ts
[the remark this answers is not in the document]
BPScott: Yep I think this is right. the style/index.ts file was deleted as part of https://github.com/Shopify/checkout-web/pull/30153/files#diff-e41076ae90717876251cef2ec8caa01124f1a24ad819de8c97031fd32b988892


You could go for 

```
export {
  Style,
  isConditionalStyle,
  isConditionalStyleWithDefault,
} from './checkout/style/style';
export type {
  AtLeastOne,
  Conditions,
  ConditionalStyle,
  ConditionalValue,
  InteractiveConditions,
  MaybeConditionalStyle,
  MaybeResponsiveConditionalStyle,
  ResolutionCondition,
  ViewportSizeCondition,
} from './checkout/style/types';
```

to keep the list of exports the same as before - I've not checked to see if there's any that you want to not expose publicly.

packages/ui-extensions/src/surfaces/checkout/targets.ts
oliverigor: @thitemple I believe this came from your [PR](https://github.com/Shopify/checkout-web/pull/29866), should this be published to unstable ?

packages/ui-extensions/src/surfaces/customer-account.ts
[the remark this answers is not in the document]
oliverigor: Since we were exporting all of them before, there were no exceptions so I'll leave as is, to not introduce any errors.

Thanks Ben

packages/ui-extensions/src/surfaces/checkout/targets.ts
alxclark: I have not followed the latest developments with the API, but to my knowledge we indeed want to keep the order confirmation on the TYP and customer account targets that Thiago added. Any idea what happened during the generation of the types ?

oliverigor: I think I found the error... 
There's a mismatch between ui-extensions and checkout-web 

check this two PR's [ui-extensions](https://github.com/Shopify/ui-extensions/commit/02ca0b9a56206d14ec742c062ca6b66c46f67c30) and [checkout web](https://github.com/Shopify/checkout-web/pull/29866/files#diff-2ad58c886715676fb5069ea6b79d5758618b4f9c9c49acfaab04b09cd59bbfe7) the additions seems the same but they are different. 

I'll add the orderConfirmation to customer accounts again but we should flag that to be matched in both repos

oliverigor: Thanks for catching this @alxclark.

@thitemple once you're back can you make sure that both checkout-web and this repo match regarding to this code.

AlexVasVas:
The **Sheet** component is missing from the package - @shopify/ui-extensions-react.
In order to test the **Sheet** component I created a new blank JavaScript React checkout extension. After that, I updated the **@shopify**, **@shopify/ui-extensions-react**, and **@shopify/checkout-ui-extensions-react** . These are the package versions i tested with: 
- @shopify/app@3.56.3
- @shopify/cli@3.56.3
- @shopify/ui-extensions-react@2024.1.1
- @shopify/ui-extensions@2024.1.1
- @types/react@18.2.60
- react-reconciler@0.29.0
- react@18.2.0

After that, I copied the example provided in the Sheet component's [documentation](https://shopify.dev/docs/api/checkout-ui-extensions/unstable/components/overlays/sheet) and execute the **npm run dev command**. Running that command resulted in the following error message:

 extensions/test-checkout-extension/src/Checkout.jsx:8:2: ERROR: No matching export in
    "node_modules/@shopify/ui-extensions-react/build/esm/surfaces/checkout.mjs" for import "Sheet"

oliverigor:
Hey @AlexVasVas.

That's because you're trying to use a stable version, and the Sheet component is on `unstable` still, you can look for the unstable versions on [NPM](https://www.npmjs.com/package/@shopify/ui-extensions-react?activeTab=versions).

PR #1754, opened 2024-02-19
Added new functionality for combined cart discounts

js-goupil:
### Background

POS is going to be supporting combined cart discounts. This PR brings in some functions, as well as a new property called `cartDiscounts: Discounts[]` to support this feature. 

### Solution

Partners are now going to be able to take advantage of this feature. Since regular manual discounts are still limited to only one discount per cart, but code discounts are not, in order to simplify the logic we are going to be separating the two functions. 

We are also providing new functions to remove discounts in a batch call, or remove a specific discount code.

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/retail-ui-extensions/src/extension-api/cart-api/cart-api.ts
+++ b/packages/retail-ui-extensions/src/extension-api/cart-api/cart-api.ts
@@ -49,8 +49,18 @@ export interface CartApiContent {
     amount?: string,
   ): Promise<void>;
 
+  /** Add a code discount to the cart
+   * @param code the code for the discount to add to the cart
+   */
+  addCartCodeDiscount(code: string): Promise<void>;
+
   /** Remove the cart discount */
   removeCartDiscount(): Promise<void>;
+
+  /** Remove all cart and line item discounts
+   * @param disableAutomaticDiscounts Whether or not automatic discounts should be enabled after removing the discounts.
+   */
+  removeAllDiscounts(disableAutomaticDiscounts: boolean): Promise<void>;
 
   /** Clear the cart */
   clearCart(): Promise<void>;
--- a/packages/retail-ui-extensions/src/extension-api/types/cart.ts
+++ b/packages/retail-ui-extensions/src/extension-api/types/cart.ts
@@ -6,6 +6,7 @@ export interface Cart {
   grandTotal: string;
   note?: string;
   cartDiscount?: Discount;
+  cartDiscounts: Discount[];
   customer?: Customer;
   lineItems: LineItem[];
   properties: Record<string, string>;


jiricrispeyn (commented):
Just quickly wanted to sanity check that we're covering all things after r[eading the impact doc](https://docs.google.com/document/d/1feei2qhGZmPaOX64-c0Xn35KZXFH_fnXrGicvF1_Jd8/edit) again.

> extension of Cart data model to allow multiple cart discounts to be returned

✅ Solved by introducing `cartDiscounts: Discount[]`

> Introduction of a new method to remove cart custom discount separately from other discounts, e.g. removeCartCustomDiscount

~❓ Would we still have to implement this?~
@daniszewskik already clarified we don't need this as POS can not do this either at the moment.

> Introduction of a new method to remove a given single discount by code, e.g. removeDiscount(discountCode)

~❓ Would we still have to implement this?~
@daniszewskik already clarified we don't need this as POS can not do this either at the moment.

> Optionally:
    - Allowing 3P apps to remove all discounts from the cart 
    - Allowing 3P apps to toggle automatic discounts

✅ Solved by `removeAllDiscounts(disableAutomaticDiscounts: boolean): Promise<void>;`

Additionally, we're introducing `applyCartCodeDiscount(code: string): Promise<void>;`, which allows a code discount to be applied next next to other discounts, which is not possible using `applyCartDiscount` at the moment.
💡 As this was not included in the initial doc, is this an issue that we only just discovered?

js-goupil:
> Additionally, we're introducing `applyCartCodeDiscount(code: string): Promise<void>;`, which allows a code discount to be applied next next to other discounts, which is not possible using `applyCartDiscount` at the moment. 💡 As this was not included in the initial doc, is this an issue that we only just discovered?

@jiricrispeyn 
I think we want to be clear on the different functionality of manual discounts and discount codes, especially when it comes to 3P devs since they don't have internal Shopify context. By having two separate functions (and we've been wanting to do this for a while anyways), it makes it clear that there's a difference between the two and that they have different behaviour. ie discount codes being able to be multiple in the cart, vs manual discounts which are only one per cart.

NathanJolly (approved):

packages/retail-ui-extensions/src/extension-api/cart-api/cart-api.ts
```diff
@@ -49,9 +49,19 @@ export interface CartApiContent {
     amount?: string,
   ): Promise<void>;
 
+  /** Add a code discount to the cart
+   * @param code the code for the discount to add to the cart
+   */
+  addCartCodeDiscount(code: string): Promise<void>;
```
vctrchu, line 55: Will this remain a string? Will we be moving to `CartDiscountType` in unified?

https://github.com/Shopify/ui-extensions/blob/46ce096430f5b42730d49c6ef2d9d2bbd439528b/packages/ui-extensions/src/surfaces/point-of-sale/api/cart-api/cart-api.ts#L29

PR #1762, opened 2024-02-20
Fix broken hyperlink for `selectedPaymentOptions`

Steve-D-Powell:
- update changeset
- Version Packages
- adjust changelog, package.json
- fix dev dependancy
- fix sentence
- Update CHANGELOG.md
- Hide line-components-render from 2023-10
- order status
- thank you
- Adding updated figma link (#1553)
- [UI components]: Update style helper docs for 2023-10 (#1657)
- Version Packages
- Update build-docs.sh
- fix broken hyperlink
- add changeset

### Background

### Solution

### 🎩

- ...

### Checklist


.changeset/chatty-papayas-sip.md
jamesvidler, line 1: For doc changes, you don't need to include a changeset. If you do, it will patch the version which is not necessary. Otherwise, looks good to me 👍.

jamesvidler (approved):

PR #1767, opened 2024-02-22
Add OrderAPI

BrittaEvansFenton:
### Background

### Solution

### 🎩

- ...

### Checklist


packages/ui-extensions/src/surfaces/point-of-sale/api/postpurchase-api/postpurchase-api.tsx
```diff
@@ -0,0 +1,26 @@
+/**
+ * Interface to access the order after checkout
+ */
+export interface PostPurchaseAPI {
```
js-goupil, line 4: I think we should name this `OrderAPIContent`, which will be generic enough to be used in OrderDetails down the road. What do you think? The Content part is strictly for the interface naming as this is how we have all our other interfaces defined (you can look up CartAPIContent for example).

js-goupil (commented):
Looks good! I think we should rename it to be more generic. Also, we're going to need to export this type and API from the package. The other APIs are all set up so you can use them as a reference :)

PR #1781, opened 2024-02-26
[Unstable] add buyer journey api to unstable

jonathanhamel4:
### Background

This PR syncs the latest content for the buyer journey and some small changes pending in checkout-web. 

Original issue with PRs linked https://github.com/Shopify/checkout-web/issues/29336

### Solution

Ran `yarn prepare-package-release`

### 🎩

Tophatted with local extension by syncing the files over to my local app. 

Both activeStep and steps are defined from using the hook
Screenshot 2024-02-26 at 5 01 16 PM

### Checklist

### DIFF
--- a/.changeset/nine-laws-switch.md
+++ b/.changeset/nine-laws-switch.md
@@ -1,0 +1,6 @@
+---
+'@shopify/ui-extensions': minor
+'@shopify/ui-extensions-react': minor
+---
+
+Adds `buyerJourney.steps`, `buyerJourney.activeStep`, and their associated react hooks to the checkout surface API.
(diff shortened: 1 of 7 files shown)


packages/ui-extensions/src/surfaces/checkout/style/types.ts
jamesvidler, line 1: @oliverigor I'm assuming the `shared.ts` file should be ignored, but what about these other changes to `style/types.ts`?

jonathanhamel4: I can also only add my stuff in this PR if that is a concern :)

oliverigor: That would be the wisest, yes :D 
Thanks @jonathanhamel4

jonathanhamel4: Removed all changes. Should be only the buyer journey api changes now 👍 . Will let the people responsible for those changes to manage the addition of the new styles/types when it is a good time.

PR #1803, opened 2024-03-04
Add SectionHeader to point-of-sale

arigny:
Part of https://github.com/Shopify/polaris-for-retail/issues/438

### Background

Add new `SectionHeader` component so that the PFR component can be exposed in POS UI extensions.

### Solution

Add `SectionHeader.ts` based on a subset of the [component interface in PFR](https://github.com/Shopify/polaris-for-retail/blob/main/src/components/List/components/SectionHeader.tsx#L29) (i.e. without list-specific props).

### 🎩

- ...

### Checklist

~- [ ] I have updated relevant documentation~


rtymchyk (approved):
✨

PR #1807, opened 2024-03-05
[Prototype] Add New Extension Target for Reorder Flow

nataliejeong:
### Background

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/extension-targets.ts
+++ b/packages/ui-extensions/src/surfaces/admin/extension-targets.ts
@@ -218,6 +218,16 @@ export interface ExtensionTargets {
    *
    * See the [tutorial](/docs/apps/selling-strategies/bundles/product-config) for more information
    */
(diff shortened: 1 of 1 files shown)


nataliejeong:
Requesting another set of reviews as I noticed the `extension-targets` file moved since I last put up this PR, so I'd like to make sure the files (including the `.changeset` is still valid before merging. Thanks!

packages/ui-extensions/src/surfaces/admin/extension-targets.ts
```diff
@@ -140,6 +140,16 @@ export interface ExtensionTargets {
     AllComponents
   >;
 
+  /**
+   * Renders an admin block in the order details page.
```
gdube, line 144: Should this be:
```suggestion
   * Renders an admin block in the product details page.
```

gdube (approved):

elanalynn (approved):

PR #1814, opened 2024-03-07
Fix missing post-purchase API types

lemonmade:
### Background

In trying to upgrade a project to use `type: "module"`, I noticed that we had never deployed a version of `@shopify/post-purchase-ui-extensions(-react)` that correctly use the `exports.types` field to set TypeScript types in an ESM project. I also noticed that we do not export some key types from the package, which makes it hard to use if you can't reach deeply into the package to grab those types.

This PR will create a new version, and in that version it applies a fix to export all the API types from the post purchase packages.


kumar303 (approved):
🚢

PR #1832, opened 2024-03-18
Adds region to Customer Privacy API

rcaplanshopify:
### Background

- https://github.com/Shopify/checkout-web/issues/31151

Adds region property to Customer Privacy API to allow partners to determine if/when to render more granular consent controls.

Monosnap Customer Privacy 2024-03-18 13-55-37

### Solution

The update uses the `getRegion` method of the `consent-tracking-api` package to surface the `region` value from the `_tracking_consent` cookie.

The Customer Privacy API parses the string that contains country and province data (when available) and translates it into country and province code values for the new `region` object.

### 🎩

[Docs](https://shopify-dev.checkout-web-api-docs-mxmh.rick-caplan.us.spin.dev/docs/api/checkout-ui-extensions/unstable/apis/customer-privacy)
- ...

### Checklist

### DIFF
--- a/.changeset/shiny-buttons-join.md
+++ b/.changeset/shiny-buttons-join.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Adds `region` to Customer Privacy API.


packages/ui-extensions/src/surfaces/checkout/api/standard/standard.ts
```diff
@@ -1774,6 +1795,14 @@ export interface CustomerPrivacy {
    * Whether the visitor is in a region requiring data sale opt-outs.
    */
   saleOfDataRegion: boolean;
+  /**
+   * Details about the visitor's current location for use in evaluating if more granular consent controls should render.
+   *
+   * @example `{countryCode: 'CA', provinceCode: 'ON'}` for a visitor in Ontario, Canada; `{countryCode: 'US', provinceCode: undefined}` for a visitor in the United States if geolocation fails to detect the state; or `undefined` if neither country nor province is detected or geolocation fails.
+   *
+   * {% include /apps/checkout/privacy-icon.md %} Requires level 1 access to [protected customer data](/docs/apps/store/data-protection/protected-customer-data).
+   */
```
rcaplanshopify, line 1804: It seems strange that these don't render in the docs. Are the examples mean for IDE type signatures only?

I see the protected customer data information render for country and province code, but no examples for them, either.

jamesvidler (approved):

PR #1836, opened 2024-03-19
add confirmation number to checkout order type

robin-drexler:
### Background

Same change as in https://github.com/Shopify/checkout-web/pull/31736
Please read this PR for context.

### Checklist


jamesvidler (approved):
Since this API doesn't actually get used and is only here for docs purposes, you can skip adding a changeset.

PR #1840, opened 2024-03-20
Redefine types for validations in a way that works for doc generation

aledalgrande:
### Background

Types were not getting generated which hindered use of the new Cart and Checkout Validation settings in Admin UI extensions.

### Solution

Refactored types so that they can be included in the generated docs.

### 🎩

[ask for internal link]

### Checklist


jamesvidler:
Can you include a spin instance to preview the doc changes?

cpeddecord (approved):

PR #1855, opened 2024-03-26
Update build-docs.sh

jamesvidler:
### Background

Add a check when regenerating this stable version to exclude customer-account types.


rcaplanshopify (approved):
LGTM

PR #1868, opened 2024-04-01
[Ui components]: Remove sheet from release

oliverigor:
### Background

### Solution

### 🎩

- ...

### Checklist


jarthorn:
> remove the code examples and the link to the code examples that were done for the Customer Privacy APIs

Isn't this ok because the Sheet component is still available for developers in the `unstable` release? We expect to re-release this in stable within 1-2 weeks so we should minimize any work to remove and re-add it.

PR #1871, opened 2024-04-01
removes redundant Capability type and uses checkout interface in customer account

rcaplanshopify:
### Background

This PR replaces the redundant `Capability` type in customer account with the type maintained in the checkout surface.

### Solution

We have been maintaining two `Capability` types, causing the instance in customer account to fall out of sync with checkout. This change allows us to maintain the capabilities in one type.

### 🎩

- ...

### Checklist


jamesvidler (commented):
I think I would prefer to have the actual type for `Capability` set in the `src/shared` rather than exported from the `checkout` surface. This will require a change in to the private package in checkout though since you'd be moving where these are defined.

PR #1939, opened 2024-05-08
Prototype shared extension context

jamesvidler:
### Background

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions-react/src/context.ts
+++ b/packages/ui-extensions-react/src/context.ts
@@ -1,0 +1,16 @@
+import {createContext} from 'react';
+import type {
+  ApiForRenderExtension as ApiForRenderExtensionCheckout,
+  RenderExtensionTarget as ApiForExtensionCheckout,
+} from '@shopify/ui-extensions/checkout';
+
+import type {
+  ApiForRenderExtension as ApiForRenderExtensionCustomerAccount,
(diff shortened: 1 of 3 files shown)


packages/ui-extensions-react/src/context.ts
```diff
@@ -0,0 +1,16 @@
...
+import type {
+  ApiForRenderExtension as ApiForRenderExtensionCustomerAccount,
+  RenderExtensionTarget as ApiForExtensionCustomerAccount,
+} from '@shopify/ui-extensions/customer-account';
+
+export const ExtensionApiContext = createContext<
+  | ApiForRenderExtensionCheckout<ApiForExtensionCheckout>
+  | ApiForRenderExtensionCustomerAccount<ApiForExtensionCustomerAccount>
+  | null
+>(null);
```
alxclark, line 16: I think this would make things simpler and less prone for consumers to import our internal union type.

```suggestion
import type {
  ExtensionApiContext as CheckoutExtensionApiContext,
} from '@shopify/ui-extensions/checkout';

import type {
  ExtensionApiContext as CustomerAccountExtensionApiContext,
} from '@shopify/ui-extensions/customer-account';

export const ExtensionApiContext = createContext<
  | CheckoutExtensionApiContext
  | CustomerAccountExtensionApiContext>
  | null
>(null);
```

PR #1954, opened 2024-05-10
Add POS UI extension getting started static page.

NathanJolly:
### Background

This is the first pass at adding a static page. This one covers https://shopify.dev/docs/apps/pos/ui-extensions/getting-started and should have the same content, just in the new format.

### Solution

I went with 'Generic' sections, which support markdown and code blocks on the side. It seems to look and feel correct for the type of page it is. I also updated the links on the placeholder overview page.

### 🎩

Compare it here:
https://shopify.dev/docs/apps/pos/ui-extensions/getting-started
https://shopify-dev.ui-extensions-70k7.nathan-oliveira.us.spin.dev/docs/api/pos-ui-extensions/unstable/getting-started

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/getting-started.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/getting-started.doc.ts
@@ -105,12 +105,9 @@ You can install your app and preview your extension in Shopify POS from the deve
 
 1. With your server running, press \`p\` to open the developer console.
 
-2. In the developer console page, complete the following tasks:
+2. To preview your extension, select the **View mobile** button. This generates a deep link or QR code.
 
-  - To install your app, click on the preview link for the POS UI extension.
-  - To preview your extension, select the **View mobile** button. This generates a deep link or QR code.
-
-3. Using a mobile device with the Shopify POS app installed, scan the QR code that was generated from your developer console. This opens Shopify POS on your device, and installs the extension. If your extension's URL changes, then you need to re-add the smart grid extension tile.
+3. Using a mobile device with the Shopify POS app installed, scan the QR code. This opens Shopify POS on your device and installs the extension in preview mode. If your extension's URL changes, you will need to re-add the smart grid extension tile.
 
 > Tip:
 > If you're using Android and your extension isn't loading, then refer to the [troubleshooting guide](/docs/apps/pos/ui-extensions/troubleshooting).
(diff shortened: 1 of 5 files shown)


packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/getting-started.doc.ts
```diff
@@ -0,0 +1,140 @@
+import type {LandingTemplateSchema} from '@shopify/generate-docs';
+
+const examplePath = './examples/getting-started';
+
+const data: LandingTemplateSchema = {
+  title: 'Getting started with POS UI extensions',
+  description:
+    'Learn how to prepare your development environment to start building POS UI extensions.',
+  id: 'getting-started',
```
vctrchu, line 9: I think we can use this `id` to specify the order of the side bar. Let me look around too see how we can do this

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/getting-started.doc.ts
```diff
@@ -0,0 +1,140 @@
...
+      anchorLink: 'step-3-install-your-app-and-preview-your-extension',
+      title: 'Step 3: Install your app and preview your extension',
+      sectionContent: `
+You can install your app and preview your extension in Shopify POS from the developer console.
+
+1. With your server running, press \`p\` to open the developer console.
+
+2. In the developer console page, complete the following tasks:
+
+  - To install your app, click on the preview link for the POS UI extension.
```
js-goupil, line 110: I think we should distinguish between installing an app (which is already installed on the test store) and running an extension
```suggestion
  - To run your extension on POS, click on the preview link for the POS UI extension.
```

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/getting-started.doc.ts
```diff
@@ -0,0 +1,140 @@
...
+  - To preview your extension, select the **View mobile** button. This generates a deep link or QR code.
```
js-goupil, line 111: Maybe we can just get rid of the previous step and say here, "To preview your extension in POS..."

js-goupil (approved):
Couple of minor suggestions. Aside from that loooks good!

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/getting-started.doc.ts
NathanJolly: Gonna merge this, and then we can mess with the side bar later! mostly focusing on getting things moved to the new format for now. lmk, i can always change this for my next PR.

PR #1958, opened 2024-05-12
Renamed EventBus to WebPixelsManager

sandor-shopify:
### Background

Updating wpm types

### Solution

Result of updates from: https://github.com/Shopify/web-pixels-manager/pull/729

### 🎩

- ...

### Checklist


emileber (approved):
We should probably update the PR title to something like "Renamed EventBus to WebPixelsManager"

PR #1970, opened 2024-05-15
Add discount extension, move some files so helpers work

NathanJolly:
### Background

Resolves https://github.com/Shopify/pos-next-react-native/issues/36272

### Solution

This adds the example discount extension static page. It also moves the pages into staticPages/pages so that the generateCodeBlock helper works for both pages in reference as well as staticPages.

I also fixed a typo in the cart api and found I had missed a section in the getting-started page. I also updated a couple of links.

### 🎩
Old: https://shopify.dev/docs/apps/pos/ui-extensions/discount-extension
New: https://shopify-dev.ui-extensions-70k7.nathan-oliveira.us.spin.dev/docs/api/pos-ui-extensions/unstable/example-discount-extension

### Checklist


packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/examples/discount-example/tile.ts
vctrchu, line 1: A follow up in the future would be to have a matching screenshot to reflect this toml file. Most people should get the general idea though

PR #1975, opened 2024-05-16
Fix naming on order and scanner api

NathanJolly:
### Background

Docs for order and scanner API weren't generating due to a typo in the file names. Also there was `subsribe` and the name of OrderAPIContent was not right.

### Solution

Fix!

### 🎩

Lookie here: https://shopify-dev.ui-extensions-70k7.nathan-oliveira.us.spin.dev/docs/api/pos-ui-extensions/unstable/apis
See? Order and Scanner API are here now. Good. Goooood.

### Checklist


vctrchu:
Thank you 🙏🏻

PR #1980, opened 2024-05-16
Add selling_plan_allocation to Web Pixels checkout.line_item object

aashna-shopify:
### Background

Part of https://github.com/Shopify/ce-customer-behaviour/issues/3991. Depends on https://github.com/Shopify/web-pixels-manager/pull/739

### Solution

- Discussed the potential schema heavily in Slack. Essentially, we want to limit the amount of fields we add to avoid adding maintenance and avoid having to load unnecessary fields onto every single checkout. This schema includes the bare minimum merchants are requesting/need. 
- A handful of fields that are in the liquid schema could not be populated given the fields that are already being requested in the API calls. We don't want to add to these api calls because that would decrease performance / increase latency on checkout. This is summarized [here](https://github.com/Shopify/ce-customer-behaviour/issues/3991#issuecomment-2105247309).

### 🎩
https://shop1.shopify.selling-plan3.aashna-narang.us.spin.dev/checkouts/cn/Z2NwLXVzLWNlbnRyYWwxOjAxSFlLNzdZTTZOTVE0WlRaWTRENlg5NDdZ/
Screenshot 2024-05-23 at 2 29 43 PM

### Checklist

### DIFF
--- a/packages/web-pixels-extension/src/schemas/pixel-events.ts
+++ b/packages/web-pixels-extension/src/schemas/pixel-events.ts
@@ -1488,6 +1488,40 @@ export const pixelEvents = {
         },
       },
     },
+    MouseEventData: {
+      metadata: {
+        description: 'An object that contains data about a mouse event',
+      },
+      properties: {
+        element: {
+          ref: 'GenericElement',
+        },
+        screenX: {
+          type: 'int32',
+        },
+        screenY: {
+          type: 'int32',
+        },
+        pageX: {
+          type: 'int32',
+        },
+        pageY: {
+          type: 'int32',
+        },
+        offsetX: {
+          type: 'int32',
+        },
+        offsetY: {
+          type: 'int32',
+        },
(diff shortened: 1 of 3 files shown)


sandor-shopify (approved):
lgtm

richpoirier (approved):

james-a-c (approved):

packages/web-pixels-extension/src/schemas/pixel-events.ts
```diff
@@ -1444,6 +1488,40 @@ export const pixelEvents = {
...
+          type: 'int32',
+        },
+        movementX: {
+          type: 'int32',
+        },
+        movementY: {
+          type: 'int32',
+        },
+      },
+    },
```
emileber, line 1524: Recently merged this, it should have no impact being available here. 👍

emileber (approved):

PR #1999, opened 2024-05-22
Njo/pos docs/stepper text

NathanJolly:
### Background

Resolves https://github.com/Shopify/pos-next-react-native/issues/36283

### Solution

Added definitions and examples

### 🎩

https://shopify-dev.ui-extensions-70k7.nathan-oliveira.us.spin.dev/docs/api/pos-ui-extensions/unstable/components/text
https://shopify-dev.ui-extensions-70k7.nathan-oliveira.us.spin.dev/docs/api/pos-ui-extensions/unstable/components/stepper

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/reference/components/Stepper.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/reference/components/Stepper.doc.ts
@@ -1,4 +1,5 @@
 import {ReferenceEntityTemplateSchema} from '@shopify/generate-docs';
+import {generateCodeBlock} from '../helpers/generateCodeBlock';
 
 const data: ReferenceEntityTemplateSchema = {
   name: 'Stepper',
@@ -14,6 +15,9 @@ const data: ReferenceEntityTemplateSchema = {
   ],
   category: 'Components',
   related: [],
+  defaultExample: {
+    codeblock: generateCodeBlock('Stepper', 'stepper', 'default.example'),
+  },
 };
 
 export default data;
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/stepper/default.example.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/stepper/default.example.ts
@@ -1,0 +1,38 @@
+import {
+  extension,
+  Screen,
+  ScrollView,
+  Stepper,
+  Text,
+} from '@shopify/ui-extensions/point-of-sale';
+
+export default extension(
+  'pos.home.modal.render',
+  (root) => {
+    const mainScreen = root.createComponent(
+      Screen,
+      {name: 'stepper', title: 'Stepper'},
+    );
+    const scrollView =
+      root.createComponent(ScrollView);
+    const text = root.createComponent(
+      Text,
+      null,
+      '4',
+    );
+    const stepper = root.createComponent(
+      Stepper,
+      {
+        initialValue: 4,
+        onValueChanged: (value) => {
+          text.replaceChildren(value.toString());
+        },
+      },
+    );
+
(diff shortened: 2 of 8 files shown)


packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/stepper/default.example.ts
```diff
@@ -0,0 +1,38 @@
...
+    const text = root.createComponent(
+      Text,
+      null,
+      '4',
+    );
+    const stepper = root.createComponent(
+      Stepper,
+      {
+        initialValue: 4,
+        onValueChanged: (value) => {
```
js-goupil, line 27: Do we need to update the value prop of the stepper itself as well? If so @merkoyep and I did something similar for the TextField docs!

js-goupil (approved):
Small comment!

packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/stepper/default.example.ts
NathanJolly: https://shopify-dev.ui-extensions-70k7.nathan-oliveira.us.spin.dev/docs/api/pos-ui-extensions/unstable/components/stepper#stepper-propertydetail-value

No, in fact, the docs explain not to do this!

vctrchu: I'm curious whats the difference between maintaining a external value state versus updating the internal state? Why is this not recommended @NathanJolly

NathanJolly: https://github.com/Shopify/pos-next-react-native/blob/main/src/components/Stepper/Stepper.tsx It's because this component is like... an uncontrolled component that also allows updates to state directly?

vctrchu: Ah i see its explicitly stated in the component, thanks!

packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/stepper/default.example.ts
```diff
@@ -0,0 +1,38 @@
...
+      Screen,
+      {name: 'stepper', title: 'Stepper'},
+    );
+    const scrollView =
+      root.createComponent(ScrollView);
+    const text = root.createComponent(
+      Text,
+      null,
+      '4',
+    );
```
vctrchu, line 22:
```suggestion
    const text = root.createComponent(
      Text,
      '4',
    );
```

packages/ui-extensions/docs/surfaces/point-of-sale/reference/components/Stepper.doc.ts
```diff
@@ -14,6 +15,9 @@ const data: ReferenceEntityTemplateSchema = {
   ],
   category: 'Components',
   related: [],
+  defaultExample: {
+    codeblock: generateCodeBlock('Stepper', 'stepper', 'default.example'),
```
vctrchu, line 19: Obviously theres no general rule but I started following JS's helper function pattern. What's your opinion? I know it may be unnecessary here since theres only one example but it might be good for consistency sake if the team uses the same approach

```
const generateCodeBlockForStepper = (title: string, fileName: string) =>
  generateCodeBlock(title, 'stepper', fileName);
```

NathanJolly: Yeah, I think as long as it's only being called once, it's unnecessary. Once it's called two or more times, it's nice.

PR #2027, opened 2024-05-28
Added docs and examples for FormattedTextField

js-goupil:
### Background

Adds docs and an example for `FormattedTextField`. Something odd that I noticed was a subtitle property on the BaseTextFieldProps that doesn't do anything. Looking on the POS side, it looks like it was a part of the old text field implementation, but even then we should not be using that at all anymore. On unstable we should be opting to only be using PFR and I'm pretty sure this has been the case anyways.

### Solution

### 🎩

https://shopify-dev.ui-extensions-7t4u.jeansebastien-goupil.us.spin.dev/docs/api/pos-ui-extensions/unstable/components/formattedtextfield

### Checklist


packages/ui-extensions/src/surfaces/point-of-sale/components/shared/BaseTextField.ts
```diff
@@ -1,9 +1,34 @@
...
+ * @property `isValid` set whether the current value in the `TextField` is valid.
+ * @property `errorMessage` sets an error message to present to the user.
+ * @property `onChangeText` a callback that is executed every time the `TextField` value changes.
+ */
 export interface BaseTextFieldProps {
+  /**
+   * The title of the `TextField`.
+   */
   title?: string;
-  subtitle?: string;
```
NathanJolly, line 3: I think there are two paths to choose from here:
1. Keep this, document it as doing nothing
2. Remove this, document it in a new section in the `versions` file for 2024.08(?)

NathanJolly: Yep, in that case, I think you can add the new section to the versions file in this PR so that we don't lose track of it.

packages/ui-extensions/docs/surfaces/point-of-sale/reference/components/FormattedTextField.doc.ts
```diff
@@ -12,9 +16,25 @@ const data: ReferenceEntityTemplateSchema = {
...
+      type: 'AutoCapitalizationType',
+    },
   ],
   category: 'Components',
   related: [],
+  defaultExample: {
+    codeblock: generateCodeBlockForComponent(
+      'Render a FormattedTextField that validates email addresses',
+      'default.example',
+    ),
```
vctrchu, line 36:
```suggestion
  defaultExample: {
    image: 'formattedTextField-default.png',
    codeblock: generateCodeBlockForComponent(
      'Render a FormattedTextField that validates email addresses',
      'default.example',
    ),
```

vctrchu (approved):

PR #2033, opened 2024-05-29
Add optional alt prop to POS image

BrittaEvansFenton:
### Background
Adds alt text as an optional prop
Follows what is expected: https://github.com/Shopify/ui-api-design/tree/main/components/Image

### Questions
have I missed anything?

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/components/Image/Image.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/components/Image/Image.ts
@@ -2,6 +2,7 @@ import {createRemoteComponent} from '@remote-ui/core';
 
 export interface ImageProps {
   src?: string;
+  alt?: string;
 }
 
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/components/Image/Image.ts
```diff
@@ -2,6 +2,7 @@ import {createRemoteComponent} from '@remote-ui/core';
 
 export interface ImageProps {
   src?: string;
+  alt?: string;
```
NathanJolly, line 5:
```suggestion
/**
 * The url of the image.
 **/
  src?: string;
/**
 * Alternative text for the image.
 **/
  alt?: string;
```

NathanJolly (approved):
I think these just nee

js-goupil:
@BrittaEvansFenton should we merge this?

PR #2043, opened 2024-05-31
Added navigation and scanner APIs to PP action

js-goupil:
Resolves https://github.com/Shopify/pos-next-react-native/issues/37280

### Background

The ScannerAPI and the NavigationAPI do work for the post purchase action target, however we were not adding it to the type in the package. This fixes the problem, and will allow developers to see that these APIs do exist on these targets. I also updated our docs. 

### 🎩

- ...

### Checklist


vctrchu (approved):
Lets go! We need to be careful with committing the json file on shopify-dev with these changes until the version with these changes are released. Oo we have versioned docs so i guess we generate it for that version

PR #2070, opened 2024-06-11
customize components category

sylvhama:
### Background

### Solution

### 🎩

- ...

### Checklist


packages/ui-extensions/docs/surfaces/customer-account/categories/components.doc.ts
```diff
@@ -0,0 +1,24 @@
...
+const data: CategoryTemplateSchema = {
+  category: 'components',
+  sections: [
+    {
+      type: 'Generic',
+      anchorLink: 'additional-components',
+      title: 'Additional components',
+      sectionContent:
+        'In addition to the components below, you can also use checkout components to build customer account UI extensions.',
+      sectionCard: [
```
brianshen1990, line 12: 👍

packages/ui-extensions/src/surfaces/customer-account/components/Avatar/Avatar.doc.ts
```diff
@@ -16,7 +16,7 @@ const data: ReferenceEntityTemplateSchema = {
       filePath: 'components/Avatar/Avatar.ts',
     },
   ],
-  category: 'Components',
+  category: 'components',
```
brianshen1990, line 19: why do we have to do this?

brianshen1990 (approved):

packages/ui-extensions/src/surfaces/customer-account/components/Avatar/Avatar.doc.ts
sylvhama: It's because the feature to customize categories was case sensitive, but they have pushed a fix since.
I still decided to keep it as lowercase, since it does not affect the rendering.

PR #2089, opened 2024-06-17
Updates to target overview docs

PSalmers:
### Background
Generated docs for https://github.com/Shopify/shopify-dev/pull/44902

Tophat here https://shopify-dev.rename-placement-big-const.patrick-salmers.us.spin.dev/docs/api/customer-account-ui-extensions/unstable/extension-targets-overview#block-extension-targets
and here
https://shopify-dev.rename-placement-big-const.patrick-salmers.us.spin.dev/docs/apps/build/customer-accounts/extension-placement#placement-references

### Checklist


deannatron (changes requested):
Updated Targets Overview content looks great on both the Checkout and Customer accounts pages.

On the Decide extension placement page for customer accounts, the "Placement references" line should be on its own line, rather than a continuation of the Block target line.

image

Johnkinging (approved):
it's too bad we can't just merge all this to one surface for the sake of docs like this.

PR #2094, opened 2024-06-18
prepare customer account 2024-07 release

robin-drexler:
**🚨DO NOT MERGE THIS INTO UNSTABLE🚨**
We'll change the base once the 2024-07 prep branch exists

### Background

Same procedure as every ~year~ quarter
https://github.com/Shopify/ui-extensions/pull/1586
https://github.com/Shopify/ui-extensions/pull/1857

Put everything in customer-accounts back to 2024-07 with the exception of updating the shared components to what is in unstable today. So e.g. `Progress` will become available.

This time around there are more changed files than the other times because we're removing the doc files which were recently added for unstable. This should be ok since the docs for the stable version still live in checkout for the time being.

### 🎩

**Only the stable targets are available**
image
If you want you can check this yourself by installing 

```
"@shopify/ui-extensions": "0.0.0-snapshot-20240618143904",
"@shopify/ui-extensions-react": "0.0.0-snapshot-20240618143904"
```

in an extension that you have yourself.

**Proof that extensions using 2024-07 still render**
- Go to https://shopify.version-bump.robin-drexler.us.spin.dev/3/account/orders/5
- Login with `dev@shopify.com` https://identity.version-bump.robin-drexler.us.spin.dev/services/mail
- You should see a rendered extension with a `2024-07` header (which reflects the version)
- There's a dropdown which uses a bunch of our apis, click through them and see that it still works. Please note that some return undefined because not everything has data

### Checklist


lihaokx (approved):
🎩  it looks good, though I do not have context of all the files.

jamesvidler (approved):

PR #2097, opened 2024-06-18
Change required sections to markdown in POS API docs

NathanJolly:
### Background

Resolves https://github.com/Shopify/pos-next-react-native/issues/38101

The `required` sections weren't doing what we expected. They were only printing the first string, which makes sense when you think of it logically as code. The first string is truthy, so the rest are ignored.

### Solution

Instead, create a subheader in the description that enumerates the supported targets, with links.

image

### 🎩

https://shopify-dev.ui-extensions-uga5.nathan-oliveira.us.spin.dev/docs/api/pos-ui-extensions/unstable/apis/cart-api

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/reference/apis/cart-api.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/reference/apis/cart-api.doc.ts
@@ -1,14 +1,27 @@
 import {ReferenceEntityTemplateSchema} from '@shopify/generate-docs';
 import {generateCodeBlock} from '../helpers/generateCodeBlock';
-import {ExtensionTargetType} from '../types/ExtensionTargetType';
+import {ExtensionTargetType, TargetLink} from '../types/ExtensionTargetType';
 
 const generateCodeBlockForCartApi = (title: string, fileName: string) =>
   generateCodeBlock(title, 'cart-api', fileName);
 
 const data: ReferenceEntityTemplateSchema = {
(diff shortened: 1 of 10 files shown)


packages/ui-extensions/docs/surfaces/point-of-sale/reference/apis/cart-api.doc.ts
```diff
@@ -1,14 +1,25 @@
...
+
+### Supporting targets
+- ${TargetLink.PosHomeTileRender}
+- ${TargetLink.PosHomeModalRender}
+- ${TargetLink.PosProductDetailsActionMenuItemRender}
+- ${TargetLink.PosProductDetailsActionRender}
+- ${TargetLink.PosCustomerDetailsActionMenuItemRender}
+- ${TargetLink.PosCustomerDetailsActionRender}
+- ${TargetLink.PosDraftOrderDetailsActionMenuItemRender}
+- ${TargetLink.PosDraftOrderDetailsActionRender}
```
vctrchu, line 21: Missing order targets?

vctrchu (approved):
Nice catch on the union type defaulting to true

packages/ui-extensions/docs/surfaces/point-of-sale/reference/apis/cart-api.doc.ts
NathanJolly: Yep, thank you!

PR #2098, opened 2024-06-18
[UI components]: `scrollTo` docs changes

oliverigor:
### Background

### Solution

### 🎩

- ...

### Checklist


shopi-dori (approved):
✅ Thanks for adding these clarifications to the docs! Confirming that we can ignore the [warning](https://github.com/Shopify/ui-extensions/pull/2098#issuecomment-2176800500) since the behavior of the `ScrollView` component remains the same.

PR #2109, opened 2024-06-19
[UI components]: 2024-01 Docs patch

oliverigor:
### Background

- Bringing some checkout docs changes to previous versions

### Solution

- Adding `StyleHelper` docs changes.
- Adding `scrollTo` clarification on docs.
- Adding `Image` source docs changes.

### 🎩

- https://shopify-dev.checkout-web-api-docs-z7ub.igor-deoliveiramartins.us.spin.dev/

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout/style/types.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/style/types.ts
@@ -109,7 +109,8 @@ export interface ConditionalStyle<
 
 /**
  * A type that represents a value that can be a conditional style.
- * We highly recommend using the Style helper which simplifies the creation of conditional styles.
+ * We highly recommend using the `Style` helper which simplifies the creation of conditional styles.
+ *
  * To learn more check out the [conditional styles](/api/checkout-ui-extensions/components/utilities/stylehelper) documentation.
  */
 export type MaybeConditionalStyle<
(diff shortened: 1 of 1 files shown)


ncardeli (approved):
:shipit:

packages/ui-extensions-react/src/surfaces/checkout/components/Image/examples/basic-image.example.tsx
```diff
@@ -10,6 +10,6 @@ export default reactExtension(
 
 function Extension() {
   return (
-    <Image source="https://shopify.dev/assets/api/checkout-extensions/checkout/components/image-example-code.png" />
+    <Image source="https://cdn.shopify.com/YOUR_IMAGE_HERE" />
```
shopi-dori, line 13: This is definitely a good clean-up 👍 

🤔 This may be out of scope (definitely tell me if it is 😉 ), but I'm wondering if there's a way to reference guidance on how to look up image links on the Shopify CDN--or if maybe using `cdn.shopify.com` is a bit misleading for third-party devs. For example as a merchant, I know I can go to Admin > Content > Files and copy the link for any of my images (see [example image link](https://cdn.shopify.com/s/files/1/0774/9802/4241/files/transformers-qr-code.jpg?v=1717525488), which has a lot of extra stuff in the URL), but my understanding is that partners/third-party developers can't upload to the Shopify CDN. Would it be better to use a generic `source` like `https://your.image.here`?

PR #2132, opened 2024-06-21
[UI Components]: Fix View position property documentation

ncardeli:
### Background

The `position` property documentation is difficult to read.

### Solution

Fix `View` component's `position` property documentation.

### 🎩

Docs: https://shopify-dev.checkout-web-api-docs-q8qm.nicolas-cardelino.us.spin.dev/docs/api/checkout-ui-extensions/unstable/components/structure/view#viewprops-propertydetail-position

### Checklist

### DIFF
--- a/.changeset/warm-tigers-sin.md
+++ b/.changeset/warm-tigers-sin.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+Improve TypeScript definition for the Position type used by the position property
--- a/packages/ui-extensions/src/surfaces/checkout/components/View/View.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/components/View/View.ts
@@ -176,16 +176,22 @@ export interface ViewProps
    */
   accessibilityRole?: ViewLikeAccessibilityRole;
   /**
-   * Changes how the `View` is positioned. For setting the position, only properties
-   * that set the offset on a different axis can be set simultaneously. This means
-   * that these combinations are not allowed:
+   * Changes how the `View` is positioned. When setting `position`, set each axis only once.
    *
    * ```
-   * { blockStart, blockEnd } // Not allowed
+   * <View position={{blockStart, inlineEnd}} /> // Allowed; sets the `block` and `inline` axes once each
    * ```
    *
    * ```
-   * { inlineStart, inlineEnd } // Not allowed
+   * <View position={{inlineStart, blockEnd}} /> // Allowed; sets the `inline` and `block` axes once each
+   * ```
+   *
+   * ```
+   * <View position={{blockStart, blockEnd}} /> // Not allowed; sets the `block` axis twice
+   * ```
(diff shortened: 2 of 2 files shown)


shopi-dori (commented):
From an internal code review perspective, this is so much easier to follow--it's super-clear which combinations are allowed now 👍 

From an external docs perspective, I'm still finding the docs pretty hard to follow and parse into something actionable for coding. I don't know if we even have control over this with the way the docs are generated, but for example, why does it give me all options for `PositionBlockEndInlineEnd` but shows `undefined` for the not-allowed ones? How would I know that means not allowed?

image

---

Possible suggestions to help: Focus on highlighting actionable content in the docs we have control over (e.g., what the developer _can_ do in code). We could adjust wording to something like:

---

Changes how the `View` is positioned. When setting `position`, use properties that affect each axis only once.

```ts
<View position={blockStart, inlineEnd} /> // Allowed; affects the `block` and `inline` axes once each
<View position={inlineStart, blockEnd} /> // Allowed; affects the `inline` and `block` axes once each
<View position={blockStart, blockEnd} /> // Not allowed; affects the `block` axis twice
<View position={inlineStart, inlineEnd} /> // Not allowed; affects the `inline` axis twice
```
---

Or if this is too much for the prop section, we could consider adding a dedicated example with description in the unused Best Practices or Examples sections. What do you think? 🤔

oliverigor:
I believe that theres no better way to show the docs due to the way they are extracted from the types. 

I agree with Dori, that in a next step we should leverage the examples section and give examples to make it clear to developers when/how to use it.

jamesvidler (approved):
Definitely agree that more code examples would be useful to augment the docs here.

oliverigor (approved):

ncardeli:
> From an external docs perspective, I'm still finding the docs pretty hard to follow and parse into something actionable for coding. I don't know if we even have control over this with the way the docs are generated, but for example, why does it give me all options for `PositionBlockEndInlineEnd` but shows `undefined` for the not-allowed ones? How would I know that means not allowed?

Agree here, not sure if we can get something better than this.


> Changes how the `View` is positioned. When setting `position`, use properties that affect each axis only once.
> 
> ```ts
> <View position={blockStart, inlineEnd} /> // Allowed; affects the `block` and `inline` axes once each
> <View position={inlineStart, blockEnd} /> // Allowed; affects the `inline` and `block` axes once each
> <View position={blockStart, blockEnd} /> // Not allowed; affects the `block` axis twice
> <View position={inlineStart, inlineEnd} /> // Not allowed; affects the `inline` axis twice
> ```

Nice suggestion, thank you.

> Or if this is too much for the prop section, we could consider adding a dedicated example with description in the unused Best Practices or Examples sections. It could show allowed `position` settings/combinations with a visual example for clarity. What do you think? 🤔

Makes sense, for now we need to improve these reference docs, but 100% agree that we need more examples.

PR #2143, opened 2024-06-28
Add UI Extensions StorageAPI

Alex-Palad:
### Background

[Hackdays project](https://hackdays.shopify.io/projects/18478)
[Partners Issue](https://github.com/Shopify/ui-extensions/issues/975)

Adding interfaces for new Storage API for POS UI Extensions

It will let UI extensions to store up to 100_000 symbols of data in POSSecureStorage and fetch it from there instead of fetching it from external API

### Solution

Add interfaces so that we could add that API to the[ pos-next-react-native](https://github.com/Shopify/pos-next-react-native/pull/38424)

### 🎩

- ...

### Checklist


Alex-Palad:
No need anymore

PR #2184, opened 2024-07-11
Remove thumbnail, and resource list from admin

Fionoble:
### Background

Some components in the Admin packages are not consumed/provided for any extensions. 

### Solution

Remove the unused components

### Checklist


vividviolet (commented):
Do we have to re-generate the docs? I'm surprised that these components didn't end up on the Dev docs initially 🤔

Fionoble:
@vividviolet No, these components were never documented.

PR #2194, opened 2024-07-17
add Menu

sylvhama:
⚠️ Should be merged after https://github.com/Shopify/customer-account-web/pull/4940

Add Menu API and its doc based on https://docs.google.com/document/d/19wN9eCwE3WZ927-C0V1HWpbgzy0u24fWGTa8OH0s1oI/edit#heading=h.y5kmpm4t7h9z

Part of https://github.com/Shopify/core-issues/issues/59120

### 🎩

- Component list: https://shopify-dev.customer-account-web-shopify-dev-4542.sylvain-hamann.us.spin.dev/docs/api/customer-account-ui-extensions/unstable/components
- Menu doc: https://shopify-dev.customer-account-web-shopify-dev-4542.sylvain-hamann.us.spin.dev/docs/api/customer-account-ui-extensions/unstable/components/menu

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/components/Menu/Menu.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/components/Menu/Menu.ts
@@ -1,5 +1,5 @@
 import {createRemoteComponent} from '@remote-ui/core';
-import {IdProps} from '../../../checkout';
+import type {IdProps} from '../shared';
 
 export interface MenuProps extends IdProps {
   /**
--- a/packages/ui-extensions/src/surfaces/customer-account/components/Avatar/Avatar.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/components/Avatar/Avatar.ts
@@ -1,5 +1,6 @@
 import {createRemoteComponent} from '@remote-ui/core';
-import {IdProps, Size} from '../../../checkout';
+import {Size} from '../../../checkout';
+import type {IdProps} from '../shared';
 
 export interface AvatarProps extends IdProps {
   /**
--- a/packages/ui-extensions/src/surfaces/customer-account/components/shared.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/components/shared.ts
@@ -1,0 +1,6 @@
+export interface IdProps {
+  /**
+   * A unique identifier for the component.
+   */
+  id?: string;
+}


packages/ui-extensions/src/surfaces/customer-account/components/Menu/Menu.ts
```diff
@@ -0,0 +1,11 @@
+import {createRemoteComponent} from '@remote-ui/core';
+import {IdProps} from '../../../checkout';
+
+export interface MenuProps extends IdProps {
```
sylvhama, line 4: @oleksandroliynyk the doc won't show `id` as a prop despite extending `IdProps`.Avatar also has this problem, progress does not 🤔 Wanna pair on it someday? 

I don't think that's a blocker for this PR.

packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
```diff
@@ -117,6 +117,12 @@ export interface Docs_ResourceItem_Button_Action
     | 'kind'
   > {}
 
+export interface Docs_Menu_Button_Action
+  extends Omit<
+    ButtonProps,
+    'kind' | 'textDecoration' | 'inlineAlignment' | 'inlineSize' | 'size'
+  > {}
```
sylvhama, line 124: I omitted props overriden in https://github.com/Shopify/checkout-web/blob/main/packages/checkout-web-ui/src/components/Button/withContextualStyle.tsx#L23

packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
```diff
@@ -117,6 +117,12 @@ export interface Docs_ResourceItem_Button_Action
     | 'kind'
   > {}
 
+export interface Docs_Menu_Button_Action
+  extends Omit<
```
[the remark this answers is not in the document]
sylvhama: Using Omit reflects how [our HOC works](https://github.com/Shopify/checkout-web/blob/main/packages/checkout-web-ui/src/components/Button/withContextualStyle.tsx#L23). We don't pick props, we force the values of a very specific set of props (those are omitted cause you can set another value, except for apparence which is not omitted). Every other props will be passed to Button. 

The day Checkout removes `submit` prop from Button, the doc will update and won't show that prop in Menu as well since we rely on the same type, so I believe it's okay to keep it for now.

oleksandroliynyk (approved):

PR #2197, opened 2024-07-17
Add `metafields` to `PickupPointOption`

klenotiw:
### Background

This simply adds the `metafields` field on to `PickupPointOption`.

### Solution

[Here is the checkout-web pr ](https://github.com/Shopify/checkout-web/pull/35842)with the implementation

### Checklist


mleandres (approved):
🚀 

good call on making the local pickup metafields not optional too!

PR #2207, opened 2024-07-23
[POS] Add POSBlock and POSBlockRow components

NathanJolly:
### Background

This adds the POSBlock and POSBlockRow components, as well as documentation.

resolves https://github.com/Shopify/pos-next-react-native/issues/39089

### Solution

Add the new components

### 🎩

No quick way to top hat since these are brand new. Should release in unstable and then start working on implementation via POS.

Docs:
https://shopify-dev.ui-extensions-vu2a.nathan-oliveira.us.spin.dev/docs/api/pos-ui-extensions/unstable/components

### Checklist


packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/pos-block/default.example.ts
```diff
@@ -0,0 +1,23 @@
...
+        action: {
+          title: 'A toast message',
+          onPress: () => {
+            api.Toast.show('A toast message');
+          },
+        },
+      },
+    );
+
+    root.append(posBlock);
```
js-goupil, line 21: We probably want an example with the actual POSBlockRow being appended to this eh? Then we can use the single example for both components' docs?

packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/pos-block/default.example.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  return (
+    <POSBlock
+      action={{
+        title: 'A toast message',
+        onPress: () => {
+          api.Toast.show('A toast message');
+        },
+      }}
+    />
```
js-goupil, line 20: Same thing here, should we include some POSBlockRow components?

packages/ui-extensions/src/surfaces/point-of-sale/components/POSBlock/POSBlockRow.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Renders a `POSBlockRow` in a `POSBlock`.
+ */
+export interface POSBlockRowProps {
+  /**
+   * Whether to remove the default padding from the `POSBlockRow`.
+   * @defaultValue false
+   */
+  removeInsets?: boolean;
```
js-goupil, line 11: Just remembered that we're also applying vertical padding by default as well, and this prop will only remove the horizontal ones (vertical ones should not be modifiable imo). Should we name this `removeHorizontalInsets` to be more clear?

packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/pos-block/default.example.ts
NathanJolly: I don't like writing examples without verifying they work anyway, so I'd rather revisit when we have screenshots and working code so I can guarantee the quality of these. That's why I started barebones. Agreed that it can be the same example for both pages tho.

js-goupil (approved):

vctrchu (approved):

PR #2220, opened 2024-07-29
[QRCode] Publish QRCode component

LucasLacerdaUX:
### Background
- Publishing QRCode component.

### Solution
- Added component structure and docs.

### 🎩 Tophat
https://shopify-dev.qr-code-component.lucas-lacerda.us.spin.dev/docs/api/checkout-ui-extensions/unstable/components/other/qrcode

### Checklist


packages/ui-extensions/docs/surfaces/checkout/reference/examples/ui-components/qrcode-fill-size.example.ts
```diff
@@ -0,0 +1,21 @@
...
+} from '@shopify/ui-extensions/checkout';
+
+export default extension(
+  'purchase.checkout.block.render',
+  (root) => {
+    const view = root.createComponent(View, {
+      maxInlineSize: 300,
+    });
+
+    const qrCode = root.createComponent(QRCode, {
```
oliverigor, line 14: I think we're missing the size: 'fill' here

packages/ui-extensions/docs/surfaces/checkout/screenshots/qrcode-example.png
oliverigor, line 1: can we keep the standard of naming the default example images as 'default'?
We can revisit it later for all the images, but for now keeping the pattern `component-default.png` makes it easier to know which one is which.

packages/ui-extensions/src/surfaces/checkout/components/QRCode/QRCode.doc.ts
```diff
@@ -0,0 +1,66 @@
...
+        },
+      ],
+    },
+  },
+  subSections: [
+    {
+      type: 'Generic',
+      anchorLink: 'best-practices',
+      title: 'Best Practices',
+      sectionContent: `
```
oliverigor, line 44: I believe we can write this as markdown right now, [see](https://github.com/Shopify/ui-extensions/pull/2014/files#diff-c068355d6a213f4ba727caca947dc925244c28050b713840f5b365cfb70df0a7R45-R64)

so no need for all of the \n\n

We can write and issue to go back and fix all other places that still utilizes this syntax.

packages/ui-extensions/src/surfaces/checkout/components/QRCode/QRCode.ts
```diff
@@ -0,0 +1,54 @@
+import {createRemoteComponent} from '@remote-ui/core';
+
+import type {IdProps, BorderStyle} from '../shared';
+
+export interface QRCodeProps extends IdProps {
+  /**
+   * The content to be encoded in the QR code, which can be any string such as a URL, email address, plain text, etc.
+   * Specific string formatting can trigger actions on the user's device when scanned, like opening geolocation
```
oliverigor, line 8: nit

```suggestion
   * Specific string formatting can trigger actions on the user’s device when scanned, like opening geolocation
```

packages/ui-extensions/src/surfaces/checkout/components/QRCode/QRCode.ts
```diff
@@ -0,0 +1,54 @@
...
+   * @default 'base'
+   */
+  border?: Extract<BorderStyle, 'none' | 'base'>;
+
+  /**
+   * A label that describes the purpose or contents of the QR code. When set,
+   * it will be announced to users using assistive technologies and will
+   * provide more context about what the QR code may do when scanned.
+   *
+   * @default 'QR code' (translated to the user's locale)
```
oliverigor, line 43: same as above

oliverigor (commented):
I'll do another scan in a bit :D 

but code looks good, let's not forget also to merge the changes applied here back to checkout web after.

packages/ui-extensions/src/surfaces/checkout/components/QRCode/QRCode.doc.ts
```diff
@@ -0,0 +1,68 @@
...
+    ],
+  },
+  related: [
+    {
+      name: 'Image',
+      subtitle: 'Component',
+      url: 'image',
+      type: 'Component',
+    },
+  ],
```
oliverigor, line 65: Why is the image component related to QRCode, is it because folks might've looked for the Image component to host the QRcode before or just because of the logo prop ? if thats the case we can add the related section also in the Image component, something worth exploring.

oliverigor (approved):
left a small question but otherwise LGTM

packages/ui-extensions/src/surfaces/checkout/components/QRCode/QRCode.doc.ts
ryan-ludwig: Oh good point. We could remove `Image` as a related component now.

.changeset/violet-lamps-relate.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/ui-extensions-react': patch
```
ryan-ludwig, line 2: this should be a minor, not a patch as it adds new features
https://github.com/Shopify/ui-extensions/blob/unstable/documentation/versions-and-deploys.md#adding-code

ryan-ludwig (approved):
Please update the .changeset to a minor, but otherwise looks great. Nice work!

.changeset/violet-lamps-relate.md
LucasLacerdaUX: Makes sense. I've seen component releases such [as Switch](https://github.com/Shopify/ui-extensions/pull/2014/files) being a patch, though.

@oliverigor Just double checking with you if we should set this as a `minor` release instead.

oliverigor: Yeah, it has been a back and fourth with this, but I believe Ryan is correct, should be a minor as this is a release of a new component. 

Sorry for steering you in the wrong direction initially :( 
I should create clear guidelines for us on when to do what regarding the changesets.

LucasLacerdaUX: Parfait! Updated the patch to `minor` :D

PR #2223, opened 2024-07-31
[QRCode] [Docs] Fix broken bulleted list formatting

LucasLacerdaUX:
### Background
While Markdown formatting works in these, list rendering on shopify.dev doesn't seem to work as expected.
This PR reverts a last-minute change introduced in #2220 where we got rid of `\n\n`.

image
https://shopify.dev/docs/api/checkout-ui-extensions/unstable/components/other/qrcode#best-practices

### Solution
Reintroduce line separation with `\n\n`.

### 🎩
https://shopify-dev.qr-code-component.lucas-lacerda.us.spin.dev/docs/api/checkout-ui-extensions/unstable/components/other/qrcode#best-practices

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout/components/QRCode/QRCode.doc.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/components/QRCode/QRCode.doc.ts
@@ -41,11 +41,7 @@ const data: ReferenceEntityTemplateSchema = {
       type: 'Generic',
       anchorLink: 'best-practices',
       title: 'Best Practices',
-      sectionContent: `
-        - Always test that the QR code is scannable from a smartphone.
-        - Include a square logo if that’s what your customers are familiar with.
-        - Increase usability by adding a text description of what the QR code does and provide alternate methods to get to the same destination, like a text input containing the value.
-      `,
+      sectionContent: `- Always test that the QR code is scannable from a smartphone.\n\n- Include a square logo if that’s what your customers are familiar with.\n\n- Increase usability by adding a text description of what the QR code does and provide alternate methods to get to the same destination, like a text input containing the value.`,
     },
   ],
   examples: {


packages/ui-extensions/src/surfaces/checkout/components/QRCode/QRCode.doc.ts
```diff
@@ -41,11 +41,7 @@ const data: ReferenceEntityTemplateSchema = {
       type: 'Generic',
       anchorLink: 'best-practices',
       title: 'Best Practices',
-      sectionContent: `
-        - Always test that the QR code is scannable from a smartphone.
-        - Include a square logo if that’s what your customers are familiar with.
-        - Increase usability by adding a text description of what the QR code does and provide alternate methods to get to the same destination, like a text input containing the value.
-      `,
```
oliverigor, line 48: I still think it does, I use it on [Switch](https://shopify.dev/docs/api/checkout-ui-extensions/unstable/components/forms/switch), can we check the formatting and try again? like [this](https://github.com/Shopify/ui-extensions/pull/2014/files#diff-c068355d6a213f4ba727caca947dc925244c28050b713840f5b365cfb70df0a7R45-R62). 

We can pair if you want

LucasLacerdaUX: Oh, you're right. The white space was the culprit. Updated so that it's multiline now without extra whitespacing.

packages/ui-extensions/src/surfaces/checkout/components/QRCode/QRCode.doc.ts
```diff
@@ -42,10 +42,9 @@ const data: ReferenceEntityTemplateSchema = {
       anchorLink: 'best-practices',
       title: 'Best Practices',
       sectionContent: `
-        - Always test that the QR code is scannable from a smartphone.
-        - Include a square logo if that’s what your customers are familiar with.
-        - Increase usability by adding a text description of what the QR code does and provide alternate methods to get to the same destination, like a text input containing the value.
-      `,
+- Always test that the QR code is scannable from a smartphone.
+- Include a square logo if that’s what your customers are familiar with.
+- Increase usability by adding a text description of what the QR code does and provide alternate methods to get to the same destination, like a text input containing the value.      `,
```
oliverigor, line 47:
```suggestion
- Increase usability by adding a text description of what the QR code does and provide alternate methods to get to the same destination, like a text input containing the value.`,
```
? not sure tbh haha but felt a lot of space there

oliverigor (approved):
approving here, but I would check with james regarding the changeset.

Also can you update the PR description and the spin link with the latest changes? :D

LucasLacerdaUX:
@oliverigor Thanks for the review! I've removed the changeset, as you suggested and confirmed by James, and updated the PR description and preview docs to reflect the latest solution!

We should be good now 🎉

PR #2244, opened 2024-08-07
UseMemo instead of useEffect when registering interceptor

dnagoda:
### Background

### Solution

### 🎩

- ...

### Checklist


dnagoda:
Hi @patryk-smc 👋 

This was just an exploration of a purposed change. Unfortunately, the change didn't pan out. What issues are you running into?

dnagoda:
What you're describing doesn't sound like it's related to what we were looking into. I'm going to close this PR because we're not going to merge the change, so if you haven't already, please file an issue for what you're seeing.

PR #2287, opened 2024-08-21
Deprecate custom validator from formatted text field

vctrchu:
Resolves https://github.com/Shopify/pos-next-react-native/issues/27757

### Background

It seems that callbacks for extensions cannot be defined with a return value other than void. I'm not exactly sure why extensions can't return values from callbacks but to resolve this issue lets deprecate and remove this for 2025-01 since it was never working in the first place.

You can find my investigations in the ticket issue description. Open to solutions.

If the team is fine with this, I will close the original issue and open a new one to track removal of this callback after 2024-10 gets released.


vctrchu (commented):
Create follow up issue to remove the callback  if team approves

NathanJolly (approved):
Might be worth it to warn that it doesn't work, too. It's not a critical feature so I am fine with removing it.

vctrchu:
Issue to track removal https://github.com/Shopify/ui-extensions/issues/2295

PR #2289, opened 2024-08-21
Make build-docs work with local `dev`

jplhomer:
### Background

You can't build the docs currently on an Apple Macintosh computer. We should change that 😄 

### Solution

- Fixed some stuff with `sed` (did you know that `-i ''` is required for Mac? idk why)
- Updated some conditional checks to make things work in both local and Spin dev
- Printed out the correct docs link based on current environment

### 🎩

- Clone it down
- Run `yarn docs:admin` for example

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/admin/build-docs.sh
+++ b/packages/ui-extensions/docs/surfaces/admin/build-docs.sh
@@ -10,6 +10,15 @@ fail_and_exit() {
   exit $1
 }
 
+run_sed() {
+  if [ -z $SPIN ]
+  then
+    sed -i '' "$1" $2
+  else
+    sed -i "$1" $2
+  fi
+}
+
 if [ -z $API_VERSION ]
 then
   API_VERSION="unstable"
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/docs/surfaces/admin/build-docs.sh
```diff
@@ -10,6 +10,15 @@ fail_and_exit() {
   exit $1
 }
 
+run_sed() {
+  if [ -z $SPIN ]
+  then
+    sed -i '' "$1" $2
```
elanalynn, line 16: According to chatGPT: The '' after -i is used for compatibility with macOS's version of sed, which requires an argument for -i (the empty string indicates no backup should be made).

Is this correct and can you explain this a bit more? Why does Spin need the MacOS version and why no back up on spin? Is there a backup when it runs locally?

elanalynn (approved):
Looks good! Just one question for my own understanding.

packages/ui-extensions/docs/surfaces/admin/build-docs.sh
elanalynn: Oh I just read your comment in the PR description ("idk why"). 👌

PR #2313, opened 2024-08-29
Generate docs update

oliverigor:
### Background

We forgot to update the general package when doing the last unstable release

The cause of shopify-dev workflow to fail was due to [this PR](https://github.com/Shopify/ui-extensions/pull/2289/files#diff-2d8e23187e91310e596a2241803ca7eb4d2411d06b9c630b2540344c48a901fb), check the errors [here](https://github.com/Shopify/shopify-dev/actions/workflows/checkout_ui_extensions_apis_regen_automation.yml)

But a fix was created [here](https://github.com/Shopify/ui-extensions/pull/2319)

### Solution

- Updating package

### Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -23,7 +23,7 @@
   resolved "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.25.2.tgz"
   integrity sha512-bYcppcpKBvX4znYaPEeFau03bp89ShqNMLs+rmdptMw+heSZh9+z84d2YG+K7cYLbWwzdjtDoW/uqZmPjulClQ==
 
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -23,7 +23,7 @@
   resolved "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.25.2.tgz"
   integrity sha512-bYcppcpKBvX4znYaPEeFau03bp89ShqNMLs+rmdptMw+heSZh9+z84d2YG+K7cYLbWwzdjtDoW/uqZmPjulClQ==
 
-"@babel/core@^7.0.0", "@babel/core@^7.0.0-0", "@babel/core@^7.0.0-0 || ^8.0.0-0 <8.0.0", "@babel/core@^7.1.0", "@babel/core@^7.11.0", "@babel/core@^7.12.0", "@babel/core@^7.12.3", "@babel/core@^7.13.0", "@babel/core@^7.14.6", "@babel/core@^7.16.0", "@babel/core@^7.4.0 || ^8.0.0-0 <8.0.0", "@babel/core@^7.7.2", "@babel/core@^7.8.0":
```
oliverigor, line 26: all of this other stuff needed to be resolved I in order to install the new docs package, does that sound right @kumar303 ?

hheyhhay: IMO it seems weird that non-`generate-doc` dependencies would have changed in the `yarn.lock` file. Would that be because `yarn -i` needed to have been run in a previous pr and/or `package.json` change?

oliverigor: Maybe? 

in this case I actually ran 

`yarn add @shopify/generate-docs@0.16.4`

To avoid updating any other dependencies, but that was the result, maybe I'm doing something wrong there ? wdyt?

jamesvidler (approved):

PR #2314, opened 2024-08-29
deprecate ActionItem and update doc examples

merkoyep:
## Background
- To align with Checkout UI Extensions, we have decided to deprecate the ActionItem component in favour of using the Button interface.
- On [POS-Next](https://github.com/Shopify/pos-next-react-native/pull/40569), the `Button` component has been mapped to the `ActionItem` component, allowing for developers to use either a `Button` or `ActionItem` for menu-item targets.

### ActionItem Documentation
- On the ActionItem page, a message to indicate the deprecated component has been added directly to the description.
   - We decided on this implementation because there are no `deprecated` tags available on the `ReferenceEntityTemplate` used for components.
- For additional visibility, the `deprecated` tag has been added to props belonging to the `ActionItem` component.
- See images and recordings below for resulting changes.
### Menu-item target code examples
- For each menu-item target, code examples in both react and typescript have been updated to utilize `Button` instead of `ActionItem`.
### Versioning Documentation
- A point has been added to versioning documentation to summarize ActionItem deprecation, telling developers to use the Button component instead.
### ActionComponents update
- Button has been added to the ActionComponents type, allowing for developers to use either an ActionItem or Button in menu-item targets.

### Components Menu
image
### ActionItem page

https://github.com/user-attachments/assets/0ed1fe38-cdca-4633-bdc7-5b0db9129015

## Example extension using ActionItem(Deprecated) VS Button

https://github.com/user-attachments/assets/119b940b-1a8d-45ca-82ea-fe204f62c3b2

## 🎩

- Visit [this Spinstance](https://shopify-dev.ui-extensions-myb4.marco-yip.us.spin.dev/docs/api/pos-ui-extensions).
- Review clarity in deprecation messaging for the following pages:
   - ActionItem
   - Components menu page
- Review code examples in both Typescript and React for the following menu-item targets:
   - pos.customer-details.action.menu-item.render
   - pos.draft-order-details.action.menu-item.render
   - pos.order-details.action.menu-item.render
   - pos.product-details.action.menu-item.render
   - pos.purchase.post.action.menu-item.render

## Checklist


packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/targets/pos-draft-order-details-action-menu-item.ts
```diff
@@ -1,13 +1,16 @@
...
+import {Button, extension} from '@shopify/ui-extensions/point-of-sale';
 
 export default extension(
   'pos.draft-order-details.action.menu-item.render',
   (root, api) => {
-    const actionItem = root.createComponent(ActionItem, {
-      onPress: () => api.action.presentModal(),
-      enabled: true,
+    const button = root.createComponent(Button, {
+      title: 'Press me!',
```
js-goupil, line 7: We should remove title from all the examples in the menu-item cases, since we don't actually render it. We instead force the `description` field to be rendered here. Can we also update the Button docs for `title` to mention this?

merkoyep: For sure! Does this mean that the `title` prop should be changed to optional?

image

NathanJolly: Hah, that's a good question. Title is required but ignored by POS in this instance. Yes, I would say making it optional is a good call.

packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/targets/customer-details-menu-item.ts
```diff
@@ -1,15 +1,16 @@
...
   (root, api) => {
-    const actionItem = root.createComponent(ActionItem, {
-      onPress: () => api.action.presentModal(),
-      enabled: true,
+    const button = root.createComponent(Button, {
+      title: 'Press me!',
+      onPress: () => {
+        api.action.presentModal();
+      },
+      isDisabled: false,
```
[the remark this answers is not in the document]
merkoyep: Good call Nate, there was something flawed in my logic on the pos-next side. Here is [the PR](https://github.com/Shopify/pos-next-react-native/pull/41546) to address the issue.

packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/targets/pos-draft-order-details-action-menu-item.ts
merkoyep: Thanks @nate, I've updated the Button's title prop to Optional and added a note for clarity.
image

js-goupil (approved):

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
```diff
@@ -30,6 +30,7 @@ const data: LandingTemplateSchema = {
 - Added support for the ${TargetLink.PosPurchasePostBlockRender} target.
 - Added support for the ${TargetLink.PosOrderDetailsBlockRender} target.
 - Added support for the ${TargetLink.PosCustomerDetailsBlockRender} target.
+- Deprecated the [ActionItem component](/docs/api/pos-ui-extensions/components/ActionItem). Please use the [Button component](/docs/api/pos-ui-extensions/components/Button) instead.
```
vctrchu, line 33: Did we decide when we are actually removing the `ActionItem`? Is having one version enough of a buffer for partners to migrate?

packages/ui-extensions/src/surfaces/point-of-sale/components/Button/Button.ts
```diff
@@ -12,8 +12,10 @@ export type ButtonType = 'primary' | 'basic' | 'destructive' | 'plain';
 export interface ButtonProps {
   /**
    * The text set on the `Button`.
+   *
+   * Note: In the case of using a button menu-item targets, the title will be ignored.
```
NathanJolly, line 16:
```suggestion
   * Note: When using a Button for menu-item targets, the title will be ignored. The text on the menu-item will be the extension's description.
```

PR #2329, opened 2024-09-09
add forceDataRefresh api

robin-drexler:
### Background
Part of Shopify/core-issues#74679

Docs and comments will follow later.

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
@@ -403,6 +403,7 @@ export interface Ui {
   toast: {
     show(content: string): void;
   };
+  forceDataRefresh(content: string): Promise<void>;
 }
 
 export interface SessionToken {


packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
```diff
@@ -403,6 +403,7 @@ export interface Ui {
   toast: {
     show(content: string): void;
   };
+  forceDataRefresh(content: string): Promise<void>;
```
robin-drexler, line 406: I also called it `content` instead of `message` because that's what we're calling it in the `toast` api.

brianshen1990: more like `forceRefresh` from the behaviour here https://github.com/Shopify/customer-account-web/pull/5361/files#diff-af2699ea46fb1d1c8a9a2352599b0bf214c648d060b7ef5613312cc322e21829R63-R64
?

robin-drexler: The goal is to refresh data on host and extensions. That we're currently implementing this by a full page reload is an implementation detail. We might change this in the future and e.g. refetch data / remount extensions without refreshing the entire page.

brianshen1990 (approved):

packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
brianshen1990: thanks for the context

lihaokx (approved):

PR #2343, opened 2024-09-12
Checkout: rename `overlay` target to `chat`

jffortier:
It was decided to rename the previously generic `purchase.checkout.overlay.render` and `purchase.thank-you.overlay.render` to a feature specific name. In our case, its for the Chat feature.

New:
- `purchase.checkout.chat.render`
- `purchase.thank-you.chat.render`

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout/targets.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/targets.ts
@@ -710,6 +710,14 @@ export interface RenderExtensionTargets {
     AnyComponent
   >;
   /**
+   * A static extension target that is rendered on top of the Checkout page as an overlay.
+   * It is positioned in the bottom right corner of the screen.
+   */
+  'purchase.checkout.chat.render': RenderExtension<
+    CheckoutApi & StandardApi<'purchase.checkout.chat.render'>,
+    AllowedComponents<'Chat'>
+  >;
+  /**
    * A static extension target that is rendered below the header on the **Thank you** page.
    */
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/src/surfaces/checkout/targets.ts
```diff
@@ -726,10 +733,10 @@ export interface RenderExtensionTargets {
     AnyComponent
   >;
   /**
-   * A static extension target that is rendered on top of the Checkout page as an overlay.
+   * A static extension target that is rendered on top of the **Thank you** page as an overlay.
    */
-  'purchase.checkout.overlay.render': RenderExtension<
-    CheckoutApi & StandardApi<'purchase.checkout.overlay.render'>,
+  'purchase.thank-you.chat.render': RenderExtension<
```
jffortier, line 738: The thank-you target was not added here, only in our internal version. This PR corrects that.

cpeddecord (approved):

packages/ui-extensions/src/surfaces/checkout/targets.ts
```diff
@@ -709,6 +709,13 @@ export interface RenderExtensionTargets {
     CheckoutApi & StandardApi<'purchase.checkout.footer.render-after'>,
     AnyComponent
   >;
+  /**
+   * A static extension target that is rendered on top of the Checkout page as an overlay.
```
igor10k, line 713: Should we update the copy considering that the target will be restricted to the bottom right corner?

igor10k (approved):

PR #2339, opened 2024-09-10
Add barcodeScannerNext props and deprecate old props

coderBackPacker:
### Background

[project vault page](https://vault.shopify.io/gsd/projects/40512)
[Figma designs](https://www.figma.com/design/yIHAojc0lbKByl3Y9Sxrpp/Search-%26-Scan-Improvements?node-id=5749-63250&node-type=CANVAS&t=omRytKGiYanXe1Xu-0)

It is part of changes that replaces the legacy cameraScanner with a new scanner that incorporates success and error animation.
I attached a video showing how the new scanner works in one of the implemented flow:
 

https://github.com/user-attachments/assets/405776d6-34d7-4925-a07b-18d154d8bc43

### Solution
1- Since the new scanner doesnt include a banner, the bannerProps will be deprecated
2- the new barcodeScannerNext takes a child prop which will be a Banner component
3- Another prop that will be introduced is the scanStatus: `success` | `error`. Based on the state, the corresponding animation will be triggred
4- the banner with the new cameraScanner is only used to show error messages. 
5- upon success and after animation, the same current logic is used.

### 🎩

- ...

### Checklist

- [x] I have :tophat:'d these changes
- [x] I have updated relevant documentation

PR #2345, opened 2024-09-13
Revert "Add barcodeScannerNext props and deprecate old props"

coderBackPacker:
Reverts Shopify/ui-extensions#2339


coderBackPacker:
Yesterday, I spent the day adding he new scanner into the sanbox extension to test it.
After trying and pairing with 
[@js.goupil](https://shopify.slack.com/team/W018PTETX9S)
, we realized that even though dispatching the actions onScucess, onError and rest works, the aminations were triggered. This is probably because the actions are not updating the redux store but only local scannerState.
That will require bigger refactor and potentially looking at better approches.
I chatted this morning with 
[@Prashant](https://shopify.slack.com/team/U02K7LNM4SF)
 and we agreed that it is not a good solution to release the UX but not being able to see the animations. We can still can use the banner as an error state as it is being used to day. But it will be half baked solution
Steps to move forward:
1- I will revert the PR making the changes to scanner.
2- there will be more investigation on how to better implement the new solution in ui-extensions.

js-goupil (commented):
Had a comment about the deprecation message. I think we can still support the banner being a child, that way it makes state handling way simpler for the UI Extension.

PR #2347, opened 2024-09-16
add customer-account.order.page.render target

brianshen1990:
### Background

Part of https://github.com/Shopify/core-issues/issues/76861
Add customer-account.order.page.render target

### 🎩

Green CI

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/targets.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/targets.ts
@@ -90,6 +90,11 @@ export interface OrderStatusExtensionTargets {
       StandardApi<'customer-account.order-status.cart-line-list.render-after'>,
     AnyComponent
   >;
+  'customer-account.order.page.render': RenderExtension<
+    OrderStatusApi<'customer-account.order.page.render'> &
+      StandardApi<'customer-account.order.page.render'>,
+    AllComponents
+  >;
 }
 export type OrderStatusExtensionTarget = keyof OrderStatusExtensionTargets;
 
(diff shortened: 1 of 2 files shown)


packages/ui-extensions/src/surfaces/customer-account/targets.ts
```diff
@@ -90,6 +90,11 @@ export interface OrderStatusExtensionTargets {
       StandardApi<'customer-account.order-status.cart-line-list.render-after'>,
     AnyComponent
   >;
+  'customer-account.order.page.render': RenderExtension<
+    OrderStatusApi<'customer-account.order-status.cart-line-list.render-after'> &
+      StandardApi<'customer-account.order-status.cart-line-list.render-after'>,
```
mathieulag, line 95: For both `OrderStatusApi<T>` and `StandardApi<T>`, shouldn't T = `customer-account.order.page.render` instead of `customer-account.order-status.cart-line-list.render-after`?

lihaokx:
What is the difference between this target and full page target?

packages/ui-extensions/src/surfaces/customer-account/targets.ts
brianshen1990: damn... I updated locally, never pushed 🤦

brianshen1990:
> What is the difference between this target and full page target?

from the target itself, this order specific full page extension will have order status api;

But we added this target so that order specific full page extension can have buyer context of an order ( order action + order specific full page extension ), if we use existing full page extension, even if the extension is designed for an order, it doesn't know

sylvhama (approved):

mathieulag (approved):

lihaokx (approved):

PR #2353, opened 2024-09-17
Update POS render function to use createRoot

js-goupil:
### Background

Currently all of our teardown functionality does not work. For example, the return method from a useEffect does not work in UI Extensions for POS (and I think on the other surfaces it's also broken?).

### Solution

This replaces the deprecated render call with the `createRoot` function, which returns a root which has a `render` function, and an `unmount` function. We are passing the unmount method into the `Promise.resolve`, which means that it can be used on the host side to properly `unmount` everything when POS unmounts the `UIExtensionContainer`.

### 🎩

You can tophat this on POS with the following POS Branch:  https://github.com/Shopify/pos-next-react-native/pull/42289

1. Pull this branch to your machine.
2. Run yarn build, and go to the build directory for ui-extensions-react, and find the `esm/surfaces/point-of-sale/render.mjs` file that gets generated. Copy the contents.
3. In a POS UI Extension running locally, paste the contents of that into the `node_modules/@shopify/ui-extensions-react/build/esm/surfaces/point-of-sale/render.mjs`.
4. Run your UI Extension and implement a useEffect with a tear down function.
5. Inject your UI Extension into POS on the branch I specified at the top of this tophat section.
6. Using a webview debugging console, make sure your tear down is getting called!

### Checklist

### DIFF
--- a/packages/ui-extensions-react/src/surfaces/point-of-sale/render.tsx
+++ b/packages/ui-extensions-react/src/surfaces/point-of-sale/render.tsx
@@ -1,6 +1,5 @@
 import type {ReactElement, PropsWithChildren} from 'react';
 import {Component} from 'react';
-import {createRoot} from '@remote-ui/react';
 import {extension} from '@shopify/ui-extensions/point-of-sale';
 import type {
   ExtensionTargets,
@@ -9,6 +8,7 @@ import type {
 } from '@shopify/ui-extensions/point-of-sale';
 
 import {ExtensionApiContext} from './context';
+import {remoteRootRender} from '../../utilities/remoteRootRender';
 
 /**
  * Registers your React-based UI Extension to run for the selected extension target.
@@ -38,24 +38,12 @@ export function reactExtension<Target extends RenderExtensionTarget>(
   // shape (`RenderExtension`).
   return extension<'pos.home.tile.render'>(target as any, async (root, api) => {
     const element = await render(api as ApiForRenderExtension<Target>);
-
-    return new Promise<() => void>((resolve, reject) => {
-      try {
-        const remoteRoot = createRoot(root);
-        remoteRoot.render(
-          <ExtensionApiContext.Provider value={api}>
-            <ErrorBoundary>{element}</ErrorBoundary>
-          </ExtensionApiContext.Provider>,
-        );
-
-        resolve(remoteRoot.unmount);
-      } catch (error) {
-        // Workaround for https://github.com/Shopify/ui-extensions/issues/325
-        // eslint-disable-next-line no-console
-        console.error(error);
-        reject(error);
-      }
-    });
+    return remoteRootRender(
+      <ExtensionApiContext.Provider value={api}>
+        <ErrorBoundary>{element}</ErrorBoundary>
+      </ExtensionApiContext.Provider>,
+      root,
+    );
   }) as any;
 }
 
--- a/packages/ui-extensions-react/src/surfaces/point-of-sale/render.tsx
+++ b/packages/ui-extensions-react/src/surfaces/point-of-sale/render.tsx
@@ -1,6 +1,6 @@
 import type {ReactElement, PropsWithChildren} from 'react';
 import {Component} from 'react';
-import {render as remoteRender} from '@remote-ui/react';
+import {createRoot} from '@remote-ui/react';
 import {extension} from '@shopify/ui-extensions/point-of-sale';
 import type {
   ExtensionTargets,
@@ -39,17 +39,16 @@ export function reactExtension<Target extends RenderExtensionTarget>(
   return extension<'pos.home.tile.render'>(target as any, async (root, api) => {
     const element = await render(api as ApiForRenderExtension<Target>);
 
-    await new Promise<void>((resolve, reject) => {
+    return new Promise<() => void>((resolve, reject) => {
       try {
-        remoteRender(
+        const remoteRoot = createRoot(root);
+        remoteRoot.render(
           <ExtensionApiContext.Provider value={api}>
             <ErrorBoundary>{element}</ErrorBoundary>
           </ExtensionApiContext.Provider>,
-          root,
-          () => {
-            resolve();
-          },
         );
+
+        resolve(remoteRoot.unmount);
       } catch (error) {
         // Workaround for https://github.com/Shopify/ui-extensions/issues/325
         // eslint-disable-next-line no-console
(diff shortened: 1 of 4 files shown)


packages/ui-extensions-react/src/surfaces/point-of-sale/render.tsx
```diff
@@ -39,22 +39,21 @@ export function reactExtension<Target extends RenderExtensionTarget>(
...
-          },
         );
+
+        return resolve(remoteRoot.unmount);
       } catch (error) {
         // Workaround for https://github.com/Shopify/ui-extensions/issues/325
         // eslint-disable-next-line no-console
         console.error(error);
-        reject(error);
+        return reject(error);
```
js-goupil, line 56: Not sure what the benefits of this are currently. It would be great to understand the type of error that can be caught here so that we can test to see what benefits this unlocks

packages/ui-extensions/src/extension.ts
```diff
@@ -37,10 +37,10 @@ export interface RenderExtensionWithRemoteRoot<
     any
   > = RemoteComponentType<any, any, any>,
 > {
-  (
-    root: RemoteRoot<AllowedComponents, AllowedComponents>,
-    api: Api,
-  ): void | Promise<void>;
+  (root: RemoteRoot<AllowedComponents, AllowedComponents>, api: Api):
+    | void
+    | Promise<() => void>
```
js-goupil, line 42: Added this return type so that `return new Promise<() => void>` in render.tsx does not cause any type complaints cc @vividviolet

packages/ui-extensions-react/src/surfaces/point-of-sale/render.tsx
```diff
@@ -39,22 +39,21 @@ export function reactExtension<Target extends RenderExtensionTarget>(
   return extension<'pos.home.tile.render'>(target as any, async (root, api) => {
     const element = await render(api as ApiForRenderExtension<Target>);
 
-    await new Promise<void>((resolve, reject) => {
+    return new Promise<() => void>((resolve, reject) => {
       try {
-        remoteRender(
+        const remoteRoot = createRoot(root);
```
js-goupil, line 44: I'm not exactly sure what else would be required in this file, but this seems to work properly. Also it would be nice to know if this change is desired elsewhere. @vividviolet you mentioned we could just share this function implementation instead of having it be duplicated everywhere.

js-goupil:
@NathanJolly Testing this on main of POS without any changes as seen in my other draft PR seems to work just fine. However, I don't feel the need to squeeze this in for 2024-10. If anything breaks on POS because of something we didn't test, well it just doesn't feel worth the risk right now. But perhaps while I'm on intermission you guys can ship this once the 2024-10 cut as been made, and then we can implement the appropriate changes on POS.

packages/ui-extensions-react/src/surfaces/point-of-sale/render.tsx
```diff
@@ -39,22 +39,21 @@ export function reactExtension<Target extends RenderExtensionTarget>(
...
           <ExtensionApiContext.Provider value={api}>
             <ErrorBoundary>{element}</ErrorBoundary>
           </ExtensionApiContext.Provider>,
-          root,
-          () => {
-            resolve();
-          },
         );
+
+        return resolve(remoteRoot.unmount);
```
vividviolet, line 51: Do we need the return? Resolving the promise should be good enough?

packages/ui-extensions-react/src/surfaces/point-of-sale/render.tsx
vividviolet: Not sure that we need a return at all

packages/ui-extensions/src/extension.ts
vividviolet: Don't we need `void | Promise<void> | Promise<() => void>`? The vanilla version of this don't have a clean up callback

packages/ui-extensions-react/src/surfaces/point-of-sale/render.tsx
```diff
@@ -39,22 +39,21 @@ export function reactExtension<Target extends RenderExtensionTarget>(
   return extension<'pos.home.tile.render'>(target as any, async (root, api) => {
     const element = await render(api as ApiForRenderExtension<Target>);
 
-    await new Promise<void>((resolve, reject) => {
+    return new Promise<() => void>((resolve, reject) => {
```
vividviolet, line 42: I think we can move most of this into a shared util: 
```ts
async function remoteRootRender(node: ReactNode) {
   return new Promise<() => void>((resolve, reject) => {
      try {
        const remoteRoot = createRoot(root);
        remoteRoot.render(node);
        return resolve(remoteRoot.unmount);
      } catch (error) {
        // Workaround for https://github.com/Shopify/ui-extensions/issues/325
        // eslint-disable-next-line no-console
        console.error(error);
        reject(error);
      }
    });
}
```

```tsx
return extension<'pos.home.tile.render'>(target as any, async (root, api) => {
  const element = await render(api as ApiForRenderExtension<Target>);
  return remoteRootRender(
    <ExtensionApiContext.Provider value={api}>
      <ErrorBoundary>{element}</ErrorBoundary>
    </ExtensionApiContext.Provider>
  );
}
```

vividviolet (commented):
I do think we should share this fix with other surface areas to be consistent

packages/ui-extensions-react/src/surfaces/point-of-sale/render.tsx
js-goupil: Got it, just wanted to check in with you first before making that larger change. I'll try to change this today if I have any downtime, but likely will be tomorrow.

packages/ui-extensions/src/extension.ts
js-goupil: Yeah sorry this diff is confusing, the Promise<void> is right under on line 43

vividviolet:
> @vividviolet Updated the PR to move the logic into a shared util function. I can update the admin UI Extension with another commit in this PR if you want, or we can wait until you guys are ready down the line?

Can you please add it for admin as well? I think we do want it for the upcoming Oct release. I'll test the PR for Admin.

js-goupil:
> Can you please add it for admin as well? I think we do want it for the upcoming Oct release. I'll test the PR for Admin.

done

@vividviolet

NathanJolly (approved):

js-goupil:
@olavoasantos @vividviolet I've been gone for a month, did we make any decisions on the useEffect cleanup? Have we shipped anything yet? If not, we might have to settle with this approach for now, at least for POS, since we have some merchant/partners who have been nagging us about this for quite some time.

vividviolet:
@js-goupil No decision has been made yet unfortunately. I'm leaning toward the simplicity of the clean up function instead of adding an event listener. @jakearchibald any opinions? We discussed this recently.

PR #2368, opened 2024-09-23
move settings to standard api

robin-drexler:
### Background
Every extension target has access to it, not just order status targets.

### 🎩

- Go to https://shopify-dev.checkout-web-api-docs-4hcy.robin-drexler.us.spin.dev/docs/api/customer-account-ui-extensions/unstable/apis
- `Settings` should not be in the `OrderStatus` section anymore, but in `Uncategorized`
- Click on `Settings`
- The page should still work and you should still be able to see the examples 

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
@@ -31,12 +31,6 @@ export interface Docs_OrderStatus_GiftCardsApi
 export interface Docs_OrderStatus_NoteApi
   extends Pick<OrderStatusApi<any>, 'note'> {}
 
-export interface Docs_OrderStatus_SettingsApi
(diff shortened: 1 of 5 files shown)


packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
```diff
@@ -31,12 +31,6 @@ export interface Docs_OrderStatus_GiftCardsApi
 export interface Docs_OrderStatus_NoteApi
   extends Pick<OrderStatusApi<any>, 'note'> {}
 
-export interface Docs_OrderStatus_SettingsApi
-  extends Pick<OrderStatusApi<any>, 'settings'> {}
-
-export interface Docs_OrderStatus_SettingsApi
```
brianshen1990, line 37: 🥶

brianshen1990 (approved):

PR #2372, opened 2024-09-25
add comment to forceDataRefresh api

robin-drexler:
Part of https://github.com/Shopify/core-issues/issues/74679
Adds explanation on how to request access to forceDataRefresh api.

### 🎩
See https://github.com/Shopify/shopify-dev/pull/48908 for instructions

### Checklist


jrrafols (approved):
Instructions look good to me.

One question: Should we indicate anywhere that the `content` string will appear in the toast following refresh?

deannatron:
>One question: Should we indicate anywhere that the `content` string will appear in the toast following refresh?

@jrrafols I think that's a good idea. What about adding this sentence before the steps, so all in all the description would be:  

> Refresh data so the surrounding information on the page is updated. The `content` string will appear in a toast message after refresh, to confirm the action was successful.
> 
> To request access to this API:
> 
> 1. Go to your partner dashboard and click **Apps**.
> 2. Select the app you need to request access for.
> 3. Click **API access**.
> 4. Under **Access force data refresh**, click **Request access**.

Does that work?

@robin-drexler I also just realized that the word "on" was missing from the first sentence (information **on** the page).

PR #2382, opened 2024-09-27
Automatically set calver version in changelog and package.json

MitchLillie:
### Background

Just an idea I had: https://shopify.slack.com/archives/C0683EBQ55M/p1727287021016169?thread_ts=1727285984.795059&cid=C0683EBQ55M

### Solution

We add a script to be run after `changeset version` which should simplify step 5 of [Creating new stable versions](https://github.com/Shopify/ui-extensions/blob/unstable/documentation/versions-and-deploys.md#creating-new-stable-versions)

### 🎩

1. [Create a github personal access token](https://github.com/settings/tokens/new) with `read:user` and `repo:status` permissions. Copy its value.
2. In the [list of PATs](https://github.com/settings/tokens), select Configure SSO -> Authorize next to Shopify.
3. Checkout this branch, and then checkout a new branch from here with a valid name (YYYY-MM, so 2027-03, or whatever value you like).
4. In your terminal run `GITHUB_TOKEN=<your token here> yarn calver` - this simulates a new stable release created from unstable.
5. Check the package.json and CHANGELOG.md files in each of the packages/ subdirectories (the top-level package.json doesn't change). They should all have a version corresponding to your branch name. For example, if you chose 2027-03 as your branch you should see 2027.3.0.
6. Run `GITHUB_TOKEN=<your token here> yarn calver` - This simulates a patch release on a pre-existing stable release.
7. Check the package.json and CHANGELOG.md files again. You should see the patch number incremented by 1. For example, if you chose 2027-03 as your branch, and you saw 2027.3.0 in the previous step, it should now read 2027.3.1.

### Checklist

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -30,6 +30,7 @@ jobs:
         with:
           title: Version Packages (${{ github.ref_name }})
           publish: yarn run deploy --tag ${{ github.ref_name }}
+          version: yarn calver
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.SHOPIFY_GH_ACCESS_TOKEN }}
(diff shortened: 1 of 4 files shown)


belalsj (approved):
🎩 'ed. LGTM

Issue #2295, opened 2024-08-22
(2025-01) Remove customValidator from FormattedTextField


vctrchu:
Follow up task from this [PR](https://github.com/Shopify/ui-extensions/pull/2287), the decision is to remove the `customValidator` from the `FormattedTextField`. This issue sets out to track the removal for 2025-01.

PR #2410, opened 2024-10-09
[Docs] Add End of Support plan to POS UI Extensions

NathanJolly:
### Background

Part of https://github.com/Shopify/pos-next-react-native/issues/43667
Resolves https://github.com/Shopify/ui-extensions/issues/2295

### Solution

Adds a section to the Versions page describing the End of Support Plan for POS UI Extensions.
Tack-on: Removes customValidator from FormattedTextField

### 🎩

https://shopify-dev.ui-extensions-k324.nathan-oliveira.us.spin.dev/docs/api/pos-ui-extensions/unstable/versions

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
@@ -39,6 +39,20 @@ Refer to the [migration guide](/docs/api/pos-ui-extensions/migrating) for more i
     },
     {
       type: 'Generic',
+      anchorLink: '202501',
+      title: '2025.01',
+      sectionContent: `
+- Added in POS version: N/A
+- Removed in POS version: N/A
+- Release day: 1/1/2025
+
+### Features
+
+- Removed \`customValidator\` prop from the [FormattedTextField](/docs/api/pos-ui-extensions/apis/formatted-text-field) component.
+      `,
+    },
+    {
+      type: 'Generic',
       anchorLink: '202410',
       title: '2024.10',
       sectionContent: `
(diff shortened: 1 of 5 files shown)


packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
```diff
@@ -13,6 +13,20 @@ const data: LandingTemplateSchema = {
...
+      type: 'Generic',
+      anchorLink: 'support-policy',
+      title: 'Support Policy',
+      sectionContent: '',
+      sectionNotice: [
+        {
+          title: 'End of Support Plan',
+          type: 'Warning',
+          sectionContent:
+            'Starting in April 2025, we will end support for versions on a one-year rolling basis. Support for legacy versions (1.0.0 - 1.7.0) and 2024-04 will end on the next POS release after April 1, 2025. Then, support for 2024-07 will end on the next POS release after July 1, 2025, and so on. Refer to the [migration guide](/docs/api/pos-ui-extensions/migrating) for more information.',
```
saarahafeez, line 26: Support for all legacy versions older than and including 2024-04 (1.0.0, 1.1.2, 1.2.0, 1.3.0, 1.4.0, 1.5.0, 1.6.0, 1.7.0, 2024.04) will be removed on the 2025.04 (April 1, 2025) release. On the 2025.07 (July 1, 2025 release), support for 2024.07 will be removed, and so on.

^ can we change to this to be SUPER clear and precise?

vctrchu:
I wonder if theres a way to put the notice as a "note" so it stands out and doesn't get lost in the version log. We can probably be committed with putting future change log sections underneath this section but just a thought!

saarahafeez:
https://shopify.dev/docs/api/usage/versioning --> "Each stable version is supported for a minimum of 12 months. This means that there are at least nine months of overlap between two consecutive stable versions. When a new stable version is introduced and contains changes that affect your app, you have nine months to test and migrate your app to the new version before support for the previous version is removed."

Instead of updating POS UI Ext dev docs with versions supported, we should have POS UI Ext listed on the above Shopify API versioning. This would be consistent with other Shopify extensibility practices.

PR #2418, opened 2024-10-17
Admin remote dom

vividviolet:
### Background

Update Admin surface to use remote-dom custom components

### Checklist

### DIFF
--- a/packages/ui-extensions-react/src/surfaces/admin/render.tsx
+++ b/packages/ui-extensions-react/src/surfaces/admin/render.tsx
@@ -1,5 +1,5 @@
 import type {ReactElement} from 'react';
-import {render as remoteRender} from '@remote-ui/react';
+import {createRoot} from 'react-dom/client';
 
 import {extension} from '@shopify/ui-extensions/admin';
 import type {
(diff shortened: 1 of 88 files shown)


packages/ui-extensions-react/src/surfaces/admin/render.tsx
```diff
@@ -36,17 +36,15 @@ export function reactExtension<ExtensionTarget extends RenderExtensionTarget>(
...
+        reactRoot.render(
           <ExtensionApiContext.Provider value={api}>
             {element}
           </ExtensionApiContext.Provider>,
-          root,
-          () => {
-            resolve();
-          },
         );
+        resolve(() => reactRoot.unmount());
```
henrytao-me, line 47: So, we decide to return a clean up function for UI extension and just data object for Data extension then.

henrytao-me (approved):
LGTM

packages/ui-extensions-react/src/surfaces/admin/render.tsx
vividviolet: I'm leaning toward this but we haven't decided. This is being discussed in this issue: https://github.com/Shopify/ui-extensions/pull/2353#issuecomment-2426849901

PR #2420, opened 2024-10-18
Add data extensions package with conditional action targets

elanalynn:
### Background

Related to new type of extension: [data extension](https://docs.google.com/document/d/14Y9EoL0b9M8x1K2yEdugBqjrwFeB1vXUXbjOVcMrj74/edit?tab=t.0#heading=h.vqkbcd97wgda)

Closes https://github.com/Shopify/app-ui/issues/1699

### Solution

- Add targets for conditional rendering action extensions in admin
- Expose `dataExtension` registration function
- Copy existing patterns from ui-extensions

### 🎩

Latest snapshot: `"@shopify/data-extensions": "0.0.0-snapshot-20241018231554",`

- Add a data extension that looks like this:

```ts
import {dataExtension} from '@shopify/data-extensions/admin';

export default dataExtension("admin.product-details.action.should-render", () => {
  return {display: false};
});
```

- Use the snapshot build in your data extension

Screenshot 2024-10-21 at 2 37 13 PM

- Make sure to update the target in the toml file too.
- Confirm that `dataExtension` and various `should-render` targets are available with no type errors 

Screenshot 2024-10-21 at 3 47 31 PM

- Confirm that using an invalid target gives you a type error

Screenshot 2024-10-21 at 3 47 44 PM

### Checklist

### DIFF
--- a/packages/data-extensions/README.md
+++ b/packages/data-extensions/README.md
@@ -1,0 +1,27 @@
+# `@shopify/data-extensions`
+
+This package contains the public type definitions and utilities needed to create a Shopify data extension.
+
+## What is a data extension?
+
+A data extension is a JavaScript-based module that can hook in to client-side behaviors on Shopify’s first party UI surface areas. It differs from a UI extension in that it does not establish a long-lived channel or render any UI on the surface. It runs once and returns data.
+
+Currently, this package contains the API for one surface.
+
+- [`admin`](./src/surfaces/admin)
+
+All extensions, regardless of where they appear in Shopify, make use of the same [underlying technology](../../documentation/how-extensions-work.md) and most of the same capabilities (e.g., direct API access, session tokens).
+
+A data extension using “vanilla” JavaScript would be written as follows:
+
+```ts
+import {dataExtension} from '@shopify/data-extensions/admin';
+
+export default dataExtension(
+  'admin.product-details.action.should-render',
+  (root, {data}) => {
+    ...
+    return {display: true}
+  },
+);
+```
--- a/packages/data-extensions/README.md
+++ b/packages/data-extensions/README.md
@@ -4,7 +4,7 @@ This package contains the public type definitions and utilities needed to create
 
 ## What is a data extension?
 
-A data extension is a JavaScript-based module that can hook in to client-side behaviors on Shopify’s first party UI surface areas. It differs from a UI extension in that it does not establish a long-lived channel or render any UI on the surface. It runs once and returns data.
+A data extension is a JavaScript-based module that can hook in to client-side behaviors on Shopify’s first party UI surface areas. It differs from a UI extension in that it does not establish a long-lived channel or render any UI on the surface. It is a stateless function that executes and returns data.
 
 Currently, this package contains the API for one surface.
 
--- a/packages/data-extensions/src/api.ts
+++ b/packages/data-extensions/src/api.ts
@@ -1,0 +1,95 @@
+/**
+ * Union of supported API versions
+ */
+export type ApiVersion = 'unstable';
+
+export interface Auth {
+  /**
+   * Retrieves a Shopify OpenID Connect ID token for the current user.
+   */
+  idToken: () => Promise<string | null>;
+}
+
+/**
+ * GraphQL error returned by the Shopify Admin API.
+ */
+export interface GraphQLError {
+  message: string;
+  locations: {
+    line: number;
+    column: string;
+  };
+}
+
+/**
+ * This defines the i18n.translate() signature.
+ */
+export interface I18nTranslate {
+  /**
+   * This returns a translated string matching a key in a locale file.
+   *
+   * @example translate("banner.title")
+   */
+  <ReplacementType = string>(
+    key: string,
+    options?: Record<string, ReplacementType | string | number>,
+  ): ReplacementType extends string | number
+    ? string
+    : (string | ReplacementType)[];
+}
+
+export interface I18n {
+  /**
+   * Returns a localized number.
+   *
+   * This function behaves like the standard `Intl.NumberFormat()`
+   * with a style of `decimal` applied. It uses the buyer's locale by default.
+   *
+   * @param options.inExtensionLocale - if true, use the extension's locale
+   */
+  formatNumber: (
+    number: number | bigint,
+    options?: {inExtensionLocale?: boolean} & Intl.NumberFormatOptions,
+  ) => string;
+
+  /**
+   * Returns a localized currency value.
+   *
+   * This function behaves like the standard `Intl.NumberFormat()`
+   * with a style of `currency` applied. It uses the buyer's locale by default.
+   *
+   * @param options.inExtensionLocale - if true, use the extension's locale
+   */
+  formatCurrency: (
+    number: number | bigint,
+    options?: {inExtensionLocale?: boolean} & Intl.NumberFormatOptions,
+  ) => string;
+
+  /**
+   * Returns a localized date value.
+   *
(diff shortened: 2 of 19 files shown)


packages/data-extensions/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## What is a data extension?
+
+A data extension is a JavaScript-based module that can hook in to client-side behaviors on Shopify’s first party UI surface areas. It differs from a UI extension in that it does not establish a long-lived channel or render any UI on the surface. It runs once and returns data.
+
+Currently, this package contains the API for one surface.
+
+- [`admin`](./src/surfaces/admin)
+
+All extensions, regardless of where they appear in Shopify, make use the same [underlying technology](../../documentation/how-extensions-work.md) and most of the same capabilities (e.g., direct API access, session tokens).
```
belalsj, line 13:
```suggestion
All extensions, regardless of where they appear in Shopify, make use of the same [underlying technology](../../documentation/how-extensions-work.md) and most of the same capabilities (e.g., direct API access, session tokens).
```

packages/data-extensions/src/api.ts
```diff
@@ -0,0 +1,103 @@
+/**
+ * Union of supported API versions
+ */
+export type ApiVersion =
+  | '2023-04'
```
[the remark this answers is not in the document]
elanalynn: Oh, good point. We shouldn't include old ones.

packages/data-extensions/README.md
```diff
@@ -0,0 +1,27 @@
+# `@shopify/data-extensions`
+
+This package contains the public type definitions and utilities needed to create a Shopify data extension.
+
+## What is a data extension?
+
+A data extension is a JavaScript-based module that can hook in to client-side behaviors on Shopify’s first party UI surface areas. It differs from a UI extension in that it does not establish a long-lived channel or render any UI on the surface. It runs once and returns data.
```
vividviolet, line 7: Does it only run once? Don't we re-run if data changes?

packages/data-extensions/src/api.ts
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Retrieves a Shopify OpenID Connect ID token for the current user.
+   */
+  idToken: () => Promise<string | null>;
+}
+
+/**
+ * GraphQL error returned by the Shopify Admin API.
+ */
+export interface GraphQLError {
```
vividviolet, line 16: We couldn't find a way to share these, eh?

packages/data-extensions/src/surfaces/admin/api/standard/standard.ts
```diff
@@ -0,0 +1,33 @@
+import {ApiVersion, Auth, GraphQLError, I18n} from '../../../../api';
+import type {ExtensionTarget as AnyExtensionTarget} from '../../extension-targets';
+
+/**
+ * The following APIs are provided to all extension targets.
+ */
+export interface StandardApi<ExtensionTarget extends AnyExtensionTarget> {
```
vividviolet, line 7: I think we're missing `data: {selected: [{id: string}]}`?

packages/data-extensions/src/surfaces/admin/globals.ts
```diff
@@ -0,0 +1,8 @@
+import type {ExtensionTargets} from './extension-targets';
+
+export interface ShopifyGlobal {
```
vividviolet, line 3: I think we're missing a `declare global` using this

packages/data-extensions/src/surfaces/admin/globals.ts
```diff
@@ -0,0 +1,8 @@
...
+  dataExtension<ExtensionTarget extends keyof ExtensionTargets>(
```
vividviolet, line 4: Shouldn't this be `run` to make the shopify global?

vividviolet (commented):
Just a couple of minor things

packages/data-extensions/README.md
```diff
@@ -0,0 +1,27 @@
...
+- [`admin`](./src/surfaces/admin)
+
+All extensions, regardless of where they appear in Shopify, make use the same [underlying technology](../../documentation/how-extensions-work.md) and most of the same capabilities (e.g., direct API access, session tokens).
+
+A data extension using “vanilla” JavaScript would be written as follows:
+
+```ts
+import {dataExtension} from '@shopify/data-extensions/admin';
+
+export default dataExtension(
```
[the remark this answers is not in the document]
vividviolet: I think we'd still want the wrapper around the global so we have the space to do additional things around it if needed. We could also call the global `dataExtension` as well. I think the reason why we had `extend` in the ui-extension package is because the wrapper differs between react (reactExtension) vs. vanilla (extension). In the case of data extension we don't have such differences.

cc @henrytao-me @developit  in case they have opinions

packages/data-extensions/src/api.ts
elanalynn: I just pushed up another commit with the changes to make these APIs shareable. It did require changing the `rootDir` to a shared one a level up. I used it in the `ui-extensions` and `data-extensions` standard API definitions, but I was hesitant to make updates throughout checkout to use the shared APIs. Maybe that can be a follow up?

packages/data-extensions/src/surfaces/admin/api/standard/standard.ts
elanalynn: I added it `Data` to the shared APIs and pulled it in from there. We should note that in UI extensions it's not part of the standard API. It's added to action, block, and print individually, but I think it's okay to deviate here. Thoughts?

packages/data-extensions/src/surfaces/admin/globals.ts
elanalynn: I'm not following. I thought we want to expose `dataExtension`. Do we call this `run` and then alias it somewhere else?

packages/data-extensions/src/api.ts
elanalynn: I fought with it for awhile (changed `rootDir`, added `paths`, `includes`, etc. and reorganized the exports) and I thought I had everything building locally, but that broke the other packages when it got to CI. What do you think about making this a follow up clean up task?

packages/data-extensions/README.md
vividviolet: I'm not suggesting we watch changes. I just think the wording could be confusing. The data extension gets executed once for the provided input data. However, it also re-runs when the data changes like (ex. navigate to another product). We also re-run the function whenever the merchant navigates back to the same resource since we don't cache results currently. Maybe we say:

```suggestion
A data extension is a JavaScript-based module that can hook in to client-side behaviors on Shopify’s first party UI surface areas. It differs from a UI extension in that it does not establish a long-lived channel or render any UI on the surface. It is a stateless function that executes and returns data.
```

packages/data-extensions/src/surfaces/admin/globals.ts
vividviolet: This interface is for defining the Shopify global. You are calling `shopify.run` in this line here: https://github.com/Shopify/ui-extensions/pull/2420/files#diff-d1bf82eda258f81a751d34a89e27f700f4e9b0ab279c63bae4e94f31cc300ea4R15

If this is supposed to be `shopify.dataExtension` then we should update it and also add the global declaration.

packages/data-extensions/src/api.ts
vividviolet: Yeah, forget it. I was just wondering if it was possible

vividviolet (approved):

PR #2423, opened 2024-10-18
Remove create-stable-version action and use regexes

MitchLillie:
### Background

This automation didn't quite work due to missing perms, and it was a bit clunky anyways as the PR this attempted to create needs to be merged _before_ the action runs.

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/documentation/versions-and-deploys.md
+++ b/documentation/versions-and-deploys.md
@@ -45,7 +45,7 @@ Like the `unstable` branch, the `internal` branch also publishes snapshot releas
 
 To create a new stable version, you will need to follow these steps:
 
-1. Create a new branch for the calver version you set from the previous step. This branch should be named with the format `YYYY-MM`. You will likely branch off of `unstable` to create this branch.
+1. Create a new branch for the calver version. This branch should be named with the format `YYYY-MM` where the month should be `01`, `04`, `07`, or `10` (we release quarterly). You will branch off of `unstable` to create this branch.
 1. Push your new branch to GitHub. This will trigger the GitHub action that creates a new PR to consume all the changesets you copied over from `unstable` into your new version.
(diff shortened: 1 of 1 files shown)


.github/workflows/create-stable-version.yml
```diff
@@ -33,16 +33,6 @@ jobs:
...
-      - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v5
-        with:
-          token: ${{ secrets.SHOPIFY_GH_ACCESS_TOKEN }}
-          commit-message: 'Add ${{ github.event.inputs.version }} to watched branches'
-          title: 'Add ${{ github.event.inputs.version }} to watched branches'
-          body: 'This PR adds the new stable version ${{ github.event.inputs.version }} to the list of watched branches in the changesets-reminder and deploy GitHub actions.'
-          branch: 'update-watched-branches-${{ github.event.inputs.version }}'
-          base: '${{ github.event.inputs.version }}'
-
```
jamesvidler, line 45: You are removing the PR that gets created, but you still create the branch with the updated GH action files? I'm not sure if I understand correctly.

MitchLillie: Ahh I see what you mean. Yes good call, we don't need that either.

jamesvidler (commented):
I see a slight problem with this existing workflow. We update the "latest version" variable a little too early in the process. Sometimes a stable version might be cut for days before it actually ships. 

Consider that we create a branch for 2025-01 and the "latest version" is updated to that. But, while it is in review a patch was shipped for 2024-10. 2024-10 is still technically the latest version, but it won't get tagged as "latest" because the "latest version" variable has already been updated to `2025-01`.

Maybe we can find another way to identify the "latest" version?

MitchLillie:
@jamesvidler Good point. If we remove that part, all the action did was create a new branch which isn't very helpful. I updated the PR to remove the new action entirely and use the new regexes where appropriate.

Issue #1495, opened 2023-11-02
[POS UI Extensions] Return UUID when Cart API's Line item level promises are resolved


dsnvs:
## Please list the related package(s)

retail-ui-extensions

## If this related to specific APIs or components, please list them here

Cart API's Line Item Level action

## Is your feature request related to a problem? Please describe.

Our apps rely on managing the line items in the cart (add a new line item, remove line item), and with the current Cart API, chained operations are cumbersome and severely increase the complexity of the code.

For example, adding new line items with properties to the cart requires us first to add the line item to the cart, then listen to cart changes to try and identify the new line item (which becomes very tricky for custom sales), so we can grab the new UUID and apply the line item property.

Maybe I am missing something here and there is a simple way to do this, but with my understanding of React Hooks (which is not that high), React Hooks' paradigm combined with the `useCartSubscription()` hook being limited to one subscriber at a time, doing this is a big hassle. In most scenarios, chaining the actions is not a hard task, but it makes the extension way more complex with data moving in multiple directions and a single `context` that has to manage the cart.

## Describe the changes you are looking for

Having the `addLineItem()` and `addCustomSale()` functions reaturn the UUIDs of the line item when the promise is resolved would eliminate this problem almost entirely.

Returning the quantity would also be super helpful so we could remove a line item from the cart if quantity was zero.

kyle-villeneuve:
This is super important for us, without it it leads to a lot of very unidiomatic code

vctrchu:
@adamwooding @dsnvs @kyle-villeneuve We have updated the [`addLineItem` and `addCustomSale`](https://github.com/Shopify/ui-extensions/pull/2526/files) to return UUIDS. It will be available for use officially in ui-extension version 2025-01 and POS 9.26 starting on 1/26/2025. Thank you for your patience.

PR #2425, opened 2024-10-21
Return the new line item uuid upon add line item

js-goupil:
### Background

This is a popular request, and a low hanging fruit and it seems like it would help out the devs a lot. https://github.com/Shopify/ui-extensions/issues/1495

### Solution
On the POS side, we will be returning the UUID of the new line item.

### Checklist


js-goupil:
While we're at it, anywhere else this might be useful?

vctrchu:
I like the idea of fixing low hanging fruit, my only concern is that I know there are problems with Promises that return values so we would need to tophat this to confirm...

js-goupil:
Closing for now due to other priorities

PR #2428, opened 2024-10-24
Make src in PrintAPI non-optional

vctrchu:
### Background

As discussed in the transition to build https://vault.shopify.io/gsd/reviews/39439#feedback-gsd_review_39439, we want to keep consistency and always require a src URL regardless if a PrintPreview component is used.


packages/ui-extensions/src/surfaces/point-of-sale/api/print-api/print-api.ts
```diff
@@ -3,9 +3,9 @@
  */
 export interface PrintApiContent {
   /** Trigger a print dialog
-   * @param src the source to print. Omit to use the current PrintPreview contents.
+   * @param src the source URL of the content to print. A src is required regardless if you are using the PrintPreview component.
```
NathanJolly, line 6: I'd suggest adding the requirements here and in the PrintPreview component, just so it's written down somewhere. This might be an example:

```suggestion
   * @param src the source URL of the content to print. This URL must be a path of the app backend. Valid document types are text, HTML, image, and PDF.
```

PR #2487, opened 2024-11-21
Defaulting to string type for enabled features passed to the templates

aeperea:
### Background

We are now passing obfuscated flags as part of the `__enabledFeatures` property in the `CustomerSegmentTemplateApi` interface. Having hardcoded types defeat the purpose of obfuscation, so we are defaulting to a string type for these flags.

### Solution

Switching types to string[]

### 🎩

- ...

### Checklist

### DIFF
--- a/.changeset/famous-parents-punch.md
+++ b/.changeset/famous-parents-punch.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Defaulting to string type for enabled features passed to the templates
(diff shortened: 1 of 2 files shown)


LTiger14 (approved):
I think it's fine given it's internal and backwards compatible anyway Also, that way we can pass unreleased features to our internal app.

PrescilliaA:
> given it's internal

@LTiger14 I am trying to remember but I think we might be using the same API for 3P extensions? 🤔

PR #2491, opened 2024-11-25
add collapsedSummary prop to AdminBlock 

belalsj:
### Background
Adding a new prop for dev to add summary text when the Block Extension has no content.

#### Example
Screenshot 2024-11-25 at 18 07 29

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components/AdminBlock/AdminBlock.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components/AdminBlock/AdminBlock.ts
@@ -7,11 +7,9 @@ export interface AdminBlockProps {
   title?: string;
 
   /**
-   * The summary to display when the app block is collapsed.
-   *
-   * @deprecated App blocks no longer have a collapsed state, so this prop is no longer supported.
+   * The summary to display when the app block is collapsed. Summary longer than 30 characters will be truncated.
    */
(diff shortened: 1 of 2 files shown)


packages/ui-extensions/src/surfaces/admin/components/AdminBlock/AdminBlock.ts
```diff
@@ -9,9 +9,14 @@ export interface AdminBlockProps {
   /**
    * The summary to display when the app block is collapsed.
    *
-   * @deprecated App blocks no longer have a collapsed state, so this prop is no longer supported.
+   * @deprecated This prop is no longer supported. Use `collapsedSummary` instead.
    */
   summary?: string;
```
vividviolet, line 14: Can you clarify why we want to introduce a new prop instead of just re-using the existing one?

AaronSpringut: We dont expose summary anywhere public in public docs nor do we support collapsing of blocks currently, so this is not a "remove then add", for developers this is just an "add". 

The new name makes it much more clear when the value of this prop will be displayed.

vividviolet (approved):

vividviolet (approved):

PR #2498, opened 2024-11-27
Only allow .render targets to run reactExtension

belalsj:
### Background
Using `reactExtension` for non-react extensions causes an infinite loop

### Solution
Only allow `.render` targets to run reactExtension

### Checklist


.changeset/neat-paws-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/ui-extensions-react': patch
+---
+
+Only allow .render targets to run reactExtension
```
vividviolet, line 5: Can you clarify that this is for Admin only?

vividviolet (approved):

PR #2504, opened 2024-12-03
Added `BaseInput` and `BaseIntent` for point-of-sale

fatbattk:
### Background

Closes https://github.com/Shopify/pos-next-react-native/issues/47313

### Solution

- Moved `api` and `components` folder under a new `render` folder.
- Moved `types` out from `api` folder to `point-of-sale`.
- Added a `run` folder with `input` and `intent` which has the base types.
- Added `device` type and added `currency` to `Session` type.

### 🎩

1) `yalc` to POS app and use the types there. `BaseInput` or `BaseIntent`.
1) On POS RN, `./scripts/ui-extensions/yalc-local-extensions add`

### Checklist


packages/ui-extensions/src/surfaces/point-of-sale/run/input/BaseInput.ts
```diff
@@ -0,0 +1,13 @@
...
+  DeviceApiContent,
+  Session,
+} from '../../../point-of-sale';
+import type {CurrencyCode} from '../../../../shared';
+
+export interface BaseInput {
+  connectivity: ConnectivityState;
+  device: DeviceApiContent;
+  locale: string;
+  session: Session & {currency: CurrencyCode};
```
js-goupil, line 12: I think we should just include currency in the `Session` api. This is a wanted feature for UI rendering extensions as well!

packages/ui-extensions/src/surfaces/point-of-sale/run/input/BaseInput.ts
```diff
@@ -0,0 +1,13 @@
+import type {
+  ConnectivityState,
+  DeviceApiContent,
+  Session,
+} from '../../../point-of-sale';
+import type {CurrencyCode} from '../../../../shared';
+
+export interface BaseInput {
+  connectivity: ConnectivityState;
+  device: DeviceApiContent;
```
js-goupil, line 10: DeviceApiContent has a bunch of getter functions which are very "API-like". I think we should re-write this into an interface that just has read values, and POS can call the functions on its side and return the values there.

packages/ui-extensions/src/surfaces/point-of-sale/run/intent/BaseIntent.ts
```diff
@@ -0,0 +1,8 @@
+interface ExtensionError extends Error {
+  level: 'error' | 'warning';
```
js-goupil, line 2: Wondering what the purpose of error vs warning will be on the POS side. Do we have a recommendation as to when either will be appropriate?

js-goupil (commented):
Good start, but we should consolidate currencyCode into the Session interface. Also should avoid having method based APIs in our input

packages/ui-extensions/src/surfaces/point-of-sale/run/intent/BaseIntent.ts
NathanJolly: https://docs.google.com/document/d/1km0Oyz_QZTWUdFtTWAv3-1qmVNsaD_shY4VAmKwSZ2g/edit?tab=t.0#bookmark=id.39h1v6blsjs5 This is to trigger the yellow or red states. We can expound on it in docs etc

packages/ui-extensions/src/surfaces/point-of-sale/run/input/BaseInput.ts
fatbattk: added.

ProdigyXable: In the near future, different locations can have the locale/currency which may differ than what is set in the store's admin.

js-goupil (approved):

packages/ui-extensions/src/surfaces/point-of-sale/types/session.ts
```diff
@@ -24,4 +26,9 @@ export interface Session {
    * Note that this staff member ID may be different to the User ID, as the staff member who enters their PIN may be different to the User who logged onto POS.
    */
   staffMemberId?: number;
+
+  /**
+   * The currency code associated with the location currently in on POS.
+   */
+  currency: CurrencyCode;
```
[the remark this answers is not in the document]
js-goupil: I think as soon as we update on the POS side the linter will complain when it doesn't see this value getting passed in

PR #2526, opened 2024-12-12
Return the uuid for add line item actions

vctrchu:
### Background

### Solution

### 🎩

- ...

### Checklist


packages/ui-extensions/src/surfaces/point-of-sale/render/api/cart-api/cart-api.ts
```diff
@@ -75,14 +75,16 @@ export interface CartApiContent {
 
   /** Add a custom sale to the cart
    * @param customSale the custom sale object to add to the cart
+   * @returns the uuid of the line item added to the cart
    */
-  addCustomSale(customSale: CustomSale): Promise<void>;
+  addCustomSale(customSale: CustomSale): Promise<string | undefined>;
```
NathanJolly, line 80: When do these return `undefined`?

vctrchu: When no UUID is found

js-goupil (approved):

PR #2534, opened 2024-12-16
Add discount functions settings api docs

devisscher:
### Background

This PR adds a section for Discount Function Settings API, similar to the Validation Settings API.

Related PR: https://github.com/Shopify/web/pull/151973

### Solution

Adding the interface and data provided to the extension point.

### 🎩

[Spin Link](https://shopify-dev.ui-extensions-5t93.thomas-devisscher.us.spin.dev/docs/api/admin-extensions/unstable/api/target-apis/discount-function-settings-api)

For info on how to setup docs: https://app-ui.docs.shopify.io/guides/documentation/ui-extensions

1. Go to Discount Function Settings API and review content.

### Checklist


packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/metafields.ts
```diff
@@ -0,0 +1,76 @@
+const supportedDefinitionTypes = [
```
olavoasantos, line 1: Does this need to be a `const`? I feel it might be ideal to keep this as a type to avoid increasing the runtime size of the library.

PR #2539, opened 2024-12-17
Add pos.purchase-complete.observe

NathanJolly:
### Background

Part of https://github.com/Shopify/pos-next-react-native/issues/47315

### Solution

This lays the groundwork for adding the `pos.purchase-complete.observe` target.

### 🎩

Pull the companion branch, yalc this all over the place, and make an extension! Run it in dev mode and click the "Other Extension Targets" tile. Click the button and your extension will run.

https://github.com/Shopify/pos-next-react-native/tree/njo/support-purchase-complete-observe

```
import {eventExtension, PurchaseCompleteInput} from '@shopify/ui-extensions/point-of-sale'

export default eventExtension('pos.purchase-complete.observe', (input: PurchaseCompleteInput) => {

    console.log(JSON.stringify(input, null, 2));

    return {
        errors: [
            {level: 'error', message: 'a test error message', name: 'a test error name'},
            {level: 'warning', message: 'a test warning message', name: 'a test warning name'}
        ]
    }
});
```

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/reference/targets/pos.purchase-complete.observe.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/reference/targets/pos.purchase-complete.observe.doc.ts
@@ -1,0 +1,14 @@
+import type {ReferenceEntityTemplateSchema} from '@shopify/generate-docs';
+import {generateCodeBlock} from '../helpers/generateCodeBlock';
+import {ExtensionTargetType} from '../types/ExtensionTargetType';
+
+const data: ReferenceEntityTemplateSchema = {
+  name: ExtensionTargetType.PosCustomerDetailsActionMenuItemRender,
+  description: 'An event extension target that observes completed checkouts',
+  category: 'Targets',
+  isVisualComponent: false,
+  related: [],
+  type: 'Target',
+};
+
+export default data;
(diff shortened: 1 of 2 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/event/input/PurchaseCompleteInput.ts
```diff
@@ -0,0 +1,11 @@
+import {BaseInput} from './BaseInput';
+
+export interface PurchaseCompleteInput extends BaseInput {
+  lineItems: any[];
```
js-goupil, line 4: I think the way this extends BaseInput means that all these fields will be at the root of the input which might be awkward. ie you would see something like `input.session.shopId` and `input.lineItems` which feels strange. Should we instead put all these fields in an object like this:

```
cart: {
  lineItems: any[],
  discounts: any[];
  attributedStaff: any;
  orderId: any;
  draftCheckoutId: any;
  taxLines: any[];
  shippingLines: any[];
}
```

js-goupil (commented):
Noticed that your PR description says "groundwork" and some things are typed to any, is the idea that we're gonna follow up with typing?

js-goupil (approved):
Saw your comment on slack. Please consider putting the fields in an object before merging though!

PR #2542, opened 2024-12-18
Adds row and column gap for POS stack

js-goupil:
### Background

Adds [row and column gap ](https://github.com/Shopify/ui-api-design/blob/177092940cd7202d5d3a81973ef9781f6eb6c53d/ui-api-design/shared/inner-layout.ts#L24-L33) overrides for the Stack component. 

This is a really useful addition for when a Stack wraps to another line as it allows the dev to customize spacing along both axis.

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/render/components/shared/inner-layouts.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/render/components/shared/inner-layouts.ts
@@ -8,6 +8,24 @@ export interface GapProps {
    * @default 'none'
    */
   gap?: SpacingKeyword;
+
+  /**
+   * Adjust spacing between elements in the block axis.
+   *
+   * This overrides the row value of `gap`.
+   *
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/render/components/shared/inner-layouts.ts
```diff
@@ -8,6 +8,24 @@ export interface GapProps {
...
   gap?: SpacingKeyword;
+
+  /**
+   * Adjust spacing between elements in the block axis.
+   *
+   * This overrides the row value of `gap`.
+   *
+   * @default '' - meaning no override
+   */
+  rowGap?: SpacingKeyword | '';
```
NathanJolly, line 19: If this is optional, isn't an empty string redundant with undefined?

NathanJolly (approved):

PR #2543, opened 2024-12-18
Njo/revert event target

NathanJolly:
### Background

This backs off the new observe target stuff. We should use it as a reference for when we want to re-add it though!

### Solution

Simple revert


egenerat (approved):
Revert looks good to me, not tested

vctrchu (approved):

PR #2559, opened 2025-01-08
Updates localized field hooks in `Unstable`

rcaplanshopify:
### Background

This PR optimizes the `useLocalizedFields` hook and introduces the `useLocalizedField` hook. It also updates the related docs.

### Solution

### 🎩

[Unstable localized field API docs](https://shopify-dev.checkout-web-api-docs-es7w.rick-caplan.us.spin.dev/docs/api/checkout-ui-extensions/unstable/apis/localized-fields)
Screenshot 2025-01-21 at 1 59 34 PM

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/checkout/reference/apis/localized-fields.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/checkout/reference/apis/localized-fields.doc.ts
@@ -26,6 +26,12 @@ const data: ReferenceEntityTemplateSchema = {
         'Returns the current localized fields and re-renders your component if the values change.',
       type: 'UseLocalizedFieldsGeneratedType',
     },
+    {
+      title: 'useLocalizedField',
+      description:
+        'Returns the current localized fields and re-renders your component if the values change.',
+      type: 'UseLocalizedFieldGeneratedType',
+    },
   ],
   defaultExample: getExample('localized-fields/default', ['jsx', 'js']),
   related: getLinksByTag('apis'),
--- a/packages/ui-extensions/docs/surfaces/checkout/reference/apis/localized-fields.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/checkout/reference/apis/localized-fields.doc.ts
@@ -29,7 +29,7 @@ const data: ReferenceEntityTemplateSchema = {
     {
       title: 'useLocalizedField',
       description:
-        'Returns the current localized fields and re-renders your component if the values change.',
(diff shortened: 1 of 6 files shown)


packages/ui-extensions/src/surfaces/checkout/api/standard/standard.ts
```diff
@@ -62,7 +62,7 @@ export interface Extension<Target extends ExtensionTarget = ExtensionTarget> {
   /**
    * The API version that was set in the extension config file.
    *
-   * @example '2024-04', '2024-07', '2024-10', '2025-01', 'unstable'
+   * @example '2024-07', '2024-10', '2025-01', '2025-04', 'unstable'
```
jamesvidler, line 65: I'm not sure if we want to add this yet. It's not a valid version yet so could be a little misleading. Usually we add this days before publishing.

packages/ui-extensions/src/shared.ts
```diff
@@ -917,6 +918,7 @@ export type StorefrontApiVersion =
   | '2024-07'
   | '2024-10'
   | '2025-01'
+  | '2025-04'
```
jamesvidler, line 921: I'm not sure if we want to add this yet. It's not a valid version yet so could be a little misleading. Usually we add this days before publishing.

packages/ui-extensions/src/shared.ts
```diff
@@ -24,6 +24,7 @@ export type ApiVersion =
...
```
jamesvidler, line 27: I'm not sure if we want to add this yet. It's not a valid version yet so could be a little misleading. Usually we add this days before publishing.

packages/ui-extensions/docs/surfaces/checkout/reference/apis/localized-fields.doc.ts
```diff
@@ -26,6 +26,12 @@ const data: ReferenceEntityTemplateSchema = {
         'Returns the current localized fields and re-renders your component if the values change.',
       type: 'UseLocalizedFieldsGeneratedType',
     },
+    {
+      title: 'useLocalizedField',
+      description:
+        'Returns the current localized fields and re-renders your component if the values change.',
```
jamesvidler, line 32:
```suggestion
        'Returns the requested localized field and re-renders your component if the value changes.',
```

rcaplanshopify:
The version updates have been dropped, and the typo has been fixed.

Screenshot 2025-01-21 at 3 00 03 PM

PR #2583, opened 2025-01-24
Add billing address hook to customer accounts

oleksandroliynyk:
### Background

### Solution

### 🎩

- ...

### Checklist


brianshen1990 (approved):
Do we want to backfill existing stable versions?

oleksandroliynyk:
> Do we want to backfill existing stable versions?

Yes, we also have another api we want to backfill (SKU field)

packages/ui-extensions-react/src/surfaces/customer-account/hooks/tests/billing-address.test.ts
```diff
@@ -0,0 +1,21 @@
+import {MailingAddress} from '@shopify/ui-extensions/customer-account';
+
+import {useBillingAddress} from '..';
+
+import {mount, createMockStatefulRemoteSubscribable} from './mount';
+
+describe('useBillingAddress', () => {
```
robin-drexler, line 7: could we add a test that tests the error case? 🙏 

Like we did here:https://github.com/Shopify/ui-extensions/blob/e739ca66785e6bda9fb576bed27dabbe7e57aa8c/packages/ui-extensions-react/src/surfaces/customer-account/hooks/tests/shopping-address.test.tsx#L7

It's also odd that we have a shopping and a shipping address test where one tests the happy path and one tests the error case. Those should be in the same file.

robin-drexler (approved):

PR #2585, opened 2025-01-24
WIP: Replace `BaseInput` with `BaseMetaInput`

fatbattk:
### Background

Spawned from, https://github.com/Shopify/pos-next-react-native/pull/51443 which automatically prepends `BaseInput` data to every worker run call on POS.

Goal here is so there is a clear separation for,
1) event target input types from,
1) the "meta data" input types.

Coupling them makes it difficult to separate type concerns on POS.

### Solution

- Move `BaseInput` properties to `BaseMetaInput`.
- Left `BaseInput` as placeholder for all event targets `*Inputs`.
- The `BaseMetaInput` types are not needed as arguments for `ExtensionRunner.run()`, only for `UnifiedExtensionWorker.run()`.
- With this PR, typing on POS is simpler with `ExtensionRunner.run()` referencing the correct `*Inputs` type, and optionally `UnifiedExtensionWorker.run()` can reference the combined typings (currently typed as `any`).

### 🎩

1) Check out this branch locally.
1) On POS RN, run `./scripts/ui-extensions/yalc-local-extensions add`.
1) Validate on IDE lint. This should not have type errors.
    ```ts
    // example:
    ExtensionRunner.run('pos.cash-tracking-session-start.event.observe', {
      cashTrackingSessionStart: {
        id: getId(cashTrackingSession.id),
        openingTime: cashTrackingSession.openingTime,
      },
    });
    ```
| Before | After |
|--|--|
| image | image |

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/event/input/BaseInput.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/event/input/BaseInput.ts
@@ -1,8 +1,12 @@
 import type {ConnectivityState, Device, Session} from '../../../point-of-sale';
 
-export interface BaseInput {
+// meta-type data that's automatically included by POS.
+export interface BaseMetaInput {
   connectivity: ConnectivityState;
   device: Device;
   locale: string;
   session: Session;
 }
+
+// required input by all event targets.
+export interface BaseInput {}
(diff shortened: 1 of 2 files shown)


js-goupil:
I think there's some confusion. The metadata in BaseInput that is being moved over is and should be passed in for all targets. It provides important information on the shop, user, and overall session. In what context would we not need it ?

PR #2587, opened 2025-01-24
Add `sku` to lineItem api in customer account extension

lihaokx:
### Background

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
@@ -639,6 +639,11 @@ export interface ProductVariant extends BaseMerchandise {
   title: string;
 
   /**
+   * The product variant's sku.
+   */
+  sku?: string;
+
+  /**
    * The product variant's subtitle.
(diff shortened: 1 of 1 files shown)


brianshen1990 (approved):
do we want it in any stable version?

packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
```diff
@@ -638,6 +638,11 @@ export interface ProductVariant extends BaseMerchandise {
    */
   title: string;
 
+  /**
+   * The product variant's sku.
+   */
+  sku?: string;
+
```
brianshen1990, line 645: hm... check is `sku: string;`, it won't be undefined https://shopify.dev/docs/api/checkout-ui-extensions/2025-01/apis/cart-lines

lihaokx:
> do we want it in any stable version?

Yes, I think so, but it will be in another pr

brianshen1990 (approved):

robin-drexler (approved):

PR #2610, opened 2025-02-03
Add app owned metafield docs to `unstable`

avocadomayo:
### Background

Add documentation for an upcoming change in the Checkout Extensibility `appMetafields` API to support app-owned metafields.

### Solution

### 🎩

- ...

### Checklist


packages/ui-extensions/docs/surfaces/checkout/reference/examples/metafields/default.example.toml
```diff
@@ -0,0 +1,6 @@
+# other configs omitted
+
+[[extensions.metafields]]
+# tip: you can use $app:some-namespace to further segment your data
```
aledalgrande, line 4:
```suggestion
# tip: you can append a string to $app, e.g. $app:some-namespace, to further segment your data
```

packages/ui-extensions/docs/surfaces/checkout/reference/examples/metafields/default.example.tsx
```diff
@@ -0,0 +1,35 @@
...
+  }).filter(
+    (entry) =>
+      entry.target.id === productVariantId,
+  );
+
+  return (
+    energyRating && (
+      <Text>
+        Energy rating:{' '}
+        {energyRating.metafield.value}
```
aledalgrande, line 31:
```suggestion
        {energyRating.metafield.value || 'energy rating not present'}
```

aledalgrande: same for the other example

packages/ui-extensions/docs/surfaces/checkout/staticPages/configuration.doc.ts
```diff
@@ -373,10 +373,38 @@ When developing a local extension, you can remove the \`block_progress\` capabil
...
+      sectionSubContent: [
+        {
+          title: 'App owned metafields',
+          sectionContent: `
+[App owned metafields](/docs/apps/build/custom-data/ownership#reserved-prefixes) are supported. You can use app owned metafields when your app needs to control the data and visibility of the metafield.
+
+Your extension can access app owned metafields that are requested in its toml using the \`$app\` format. Your extension can only access app owned metafields that belong to its parent app.
+
+> Caution:
+> When accessing app owned metafields, you must use the \`$app\` format. The fully qualified reserved namespace format such as \`app--{your-app-id}[--{optional-namespace}]\` is not supported.
```
aledalgrande, line 394: Don't know if this was part of the discussion, but I feel like mentioning fully qualified here is not right since we only want to show the One True Way™ to devs going forward. Maybe this should move to the version migration page?

avocadomayo: I think we should leave it in so that it can help developers understand which formats are _not_ accepted. It would be a nice companion to the TOML validation we have shipped. Btw per release review feedback, we have removed the version migration page since we felt it did not add much value and the existing content was old

PR #2613, opened 2025-02-05
[Admin Components] 29.0.0

gwyneplaine:
### Background

### Solution

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components.d.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components.d.ts
@@ -1,9 +1,13 @@
-/** VERSION: 0.28.0 **/
+/** VERSION: 0.29.0 **/
 /* eslint-disable @typescript-eslint/ban-types */
 /* eslint-disable @typescript-eslint/no-namespace */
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/src/surfaces/admin/components.d.ts
```diff
@@ -1,9 +1,13 @@
-/** VERSION: 0.28.0 **/
+/** VERSION: 0.29.0 **/
 /* eslint-disable @typescript-eslint/ban-types */
 /* eslint-disable @typescript-eslint/no-namespace */
 /* eslint-disable @typescript-eslint/member-ordering */
 // eslint-disable-next-line @typescript-eslint/triple-slash-reference, spaced-comment
 /// <reference lib="DOM" />
+import {ReactNode} from 'react';
+
+import * as preact from 'preact';
```
gwyneplaine, line 9: ran this using the latest in https://github.com/Shopify/admin-ui-foundations/pull/1918 
@vividviolet is importing from these an libraries expected result?

PR #2634, opened 2025-02-12
Fix empty bulkCartUpdate typing in docs

vctrchu:
Resolves https://github.com/Shopify/shopify-dev/issues/53252

### Background

The omit operators don't play well when generating dev docs. I will update the type to exclude the variables instead. Since the docs are evaluated based on the interfaces, they are no generated json file diffs. I ran npm run docs:point-of-sale to be sure.

https://shopify-dev.ui-extensions-pp9g.victor-chu.us.spin.dev/docs/api/pos-ui-extensions/unstable/apis/cart-api#cartapi-propertydetail-bulkcartupdate

Before
Screenshot 2025-02-12 at 12 59 18 PM

After
Screenshot 2025-02-12 at 1 14 51 PM

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
@@ -12,10 +12,14 @@ export interface Cart {
   properties: Record<string, string>;
 }
 
-export type CartUpdateInput = Omit<
-  Cart,
-  'subtotal' | 'taxTotal' | 'grandTotal'
->;
+export interface CartUpdateInput {
+  note?: string;
+  cartDiscount?: Discount;
+  cartDiscounts: Discount[];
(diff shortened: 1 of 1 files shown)


NathanJolly (approved):
Can you lodge this as a bug with the team that makes the doc generation tooling?

vctrchu:
@NathanJolly 

> Can you lodge this as a bug with the team that makes the doc generation tooling?

Yes, I'll reach out to them to notify them of the issue

vctrchu:
Issue lodged here https://github.com/Shopify/shopify-dev/issues/53575#issue-2852274121

stefanvermaas:
Just to add to this, it would be great to get some examples for creating, editing, and deleting these cart items in bulk, too. With this PR, the docs are added, but it's still not clear how to use them and what the use cases are.

PR #2639, opened 2025-02-14
feat(products): updating Product Configuration extension points to use block extension api & components

kmdavis:
Part of https://github.com/Shopify/temp-project-mover-Archetypically-20250320131408/issues/241

### Background

When the product configuration extension points were first created, all data necessary for rendering had to passed in. This is no longer the case, as ui extensions now have access to a graphql api. [The prior PR in this chain](https://github.com/Shopify/web/pull/158391) added the product or variant id as `selected[0].id` (the same pattern used by block extensions), and this PR updates the types to reflect this new field, and marks the existing product/productVariant objects as deprecated (we'll remove them in a few months, once all third party apps are no longer using it)

### Checklist

### DIFF
--- a/.changeset/tender-boxes-own.md
+++ b/.changeset/tender-boxes-own.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Update Product Configuration extension points to use block extension api and data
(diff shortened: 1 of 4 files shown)


wesbaker (dismissed):
nitpick: Testing dismiss changes

hannakebedom (approved):

PR #2643, opened 2025-02-18
Add `pos.receipt-footer.block.render`

fatbattk:
Part of, https://github.com/Shopify/retail-hq-store-management/issues/755

- Add `pos.receipt-footer.block.render` render target.
    - Has `OrderApi`.
    - Follow up PR will introduce new receipt block.
- Remove unused `pos.prepare-receipt.event.inject`.

### Background

- Removal done referencing, https://github.com/Shopify/ui-extensions/pull/2597

### Solution

- Chose bare bones version from other block render targets.
- Created new subcategory `Receipts` since none fits this target. Is the documentation links correct?
- Did this under the `2025.04` release.

### 🎩

- ...

### Checklist


packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
```diff
@@ -127,6 +127,10 @@ export interface ExtensionTargets {
       ActionApi,
     BlockComponents
   >;
+  'pos.receipt-footer.block.render': RenderExtension<
+    StandardApi<'pos.receipt-footer.block.render'>,
+    BlockComponents
+  >;
```
Alex-Palad, line 133: We'd have to update `BlockComponents` to new components after we introduce ReceiptBlock components

Alex-Palad (approved):

packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
fatbattk: yup already in other PR. but now I wonder if this target should limit all the components here as well vs just in the block.

ProdigyXable: Agreed that, when this PR is over, the allowed components should be a specific as possible if this target is meant to only allow `ReceiptBlock`-related components as children.

Looks like it is accounted for via https://github.com/Shopify/ui-extensions/pull/2644/commits/c1778a266f889c6a7a42d1ce04c298032944d9c6

ProdigyXable (approved):

js-goupil (approved):

PR #2644, opened 2025-02-19
Add `POSReceiptBlock` component

fatbattk:
Resolves https://github.com/Shopify/retail-hq-store-management/issues/755

Part 2 of 2. [Part 1 here.](https://github.com/Shopify/ui-extensions/pull/2643)

- Add `POSReceiptBlock` with children limitations and documentations.

### Background

`POSReceiptBlock` is a block wrapper specific for the POS receipt section and does not do anything UI-wise other than gating the allowed children.

### Solution

- Defined acceptable children is currently set to `Image` and `Text` only. New components will be introduced in later PRs.
- Documentation is WIP since no screenshots + other descriptions.
- Did this under the `2025.04` release.

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
@@ -10,6 +10,9 @@ import {DraftOrderApi} from './render/api/draft-order-api/draft-order-api';
 
 type SmartGridComponents = AnyComponentBuilder<Pick<Components, 'Tile'>>;
 type ActionComponents = AnyComponentBuilder<Pick<Components, 'Button'>>;
+type ReceiptComponents = AnyComponentBuilder<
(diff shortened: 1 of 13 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
```diff
@@ -25,6 +25,7 @@ type BlockComponents = AnyComponentBuilder<
     | 'Stack'
     | 'Text'
     | 'TimePicker'
+    | 'POSReceiptBlock'
```
Alex-Palad: Do we want all Block targets to use this component?

fatbattk: good point. don't think it should.
will separate out to it's own.

NathanJolly (approved):

Alex-Palad (approved):

PR #2646, opened 2025-02-22
Remove "pos.cash-tracking-session-cancel.event.observe" event target

fatbattk:
### Background

Added previously but not used in POS. After discussions we do not see a use for cash tracking canceled event target.

### Solution

### 🎩

- ...

### Checklist


Alex-Palad (approved):

vctrchu (commented):
Is there a doc where the decision was made? Just want to make sure we don't have to re-add it again

fatbattk:
@vctrchu it's a comment [on this Doc](https://docs.google.com/spreadsheets/d/1BpnYutnV6oJ2PTqXAdbqyvkxeHYMH-vrMhMYZeaoUPg/edit?gid=0#gid=0), except I don't see it now (not sure if should restore from previous version), but the gist is that there is no way to "cancel" a cash tracking session (errors excluded).

PR #2648, opened 2025-02-24
[Admin UI Components] Update component definitions 24 02 2025

gwyneplaine:
### Background

Add NumberField and Select component and associated definitions to `component.d.ts`
https://github.com/Shopify/admin-ui-foundations/pull/2166

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components.d.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components.d.ts
@@ -2017,7 +2017,7 @@ interface SelectProps$1
   extends GlobalProps,
     AutocompleteProps<AnyAutocompleteField>,
     Pick<FieldDecorationProps, 'icon'>,
-    Omit<FieldProps, 'defaultValue'>,
+    Exclude<FieldProps, 'defaultValue'>,
     FocusEventProps {
   /**
    * The options a user can select from.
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/src/surfaces/admin/components.d.ts
```diff
@@ -1291,6 +1291,48 @@ interface FieldDecorationProps {
...
+   * Sets the type of controls displayed in the field.
+   *
+   * - `stepper`: displays buttons to increase or decrease the value of the field by the stepping interval defined in the `step` property.
+   * Appropriate mouse and [keyboard interactions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/spinbutton_role#keyboard_interactions) to control the value of the field are enabled.
+   * - `none`: no controls are displayed and users must input the value manually. Arrow keys and scroll wheels can’t be used either to avoid accidental changes.
+   * - `auto`: the presence of the controls depends on the surface and context.
+   *
+   * @default 'auto'
+   */
+  controls?: 'auto' | 'stepper' | 'none';
```
billfienberg, line 1334: It looks like we never implemented `controls` in `admin-ui-foundations` ([source](https://github.com/Shopify/admin-ui-foundations/blob/594b25175e1da3678cdf79fae9e6fdf417f42ea5/admin-ui-components/src/components/NumberField/properties.ts#L11)). 

If the implementation doesn't currently support `controls`, then should we omit `controls` from the remote definition? 

CC @vividviolet @alex-page

kyledurand: I think I noticed that this type def file copies our types as well as the types from ui-api-design, it doesn't mean `controls` is referenced from our props. I could be wrong though

vividviolet: It picks up the shared ui-api-design stuff but it `controls` is showing up in the remote definition then we need to omit it from our definition. In this particular case we do need to exclude `controls` the [type definition](https://github.com/Shopify/ui-extensions/blob/2d61ea71d8311ca261792fa000833acb5dd058ee/packages/ui-extensions/src/surfaces/admin/components.d.ts#L4923) extends from [`NumberFieldProps$1`](https://github.com/Shopify/ui-extensions/blob/2d61ea71d8311ca261792fa000833acb5dd058ee/packages/ui-extensions/src/surfaces/admin/components.d.ts#L1897) which extends from `NumberConstraintsProps` so it pulls in `controls`.

billfienberg: > In this particular case we do need to exclude `controls`...

For now, should we manually remove the unimplemented types from the remote definitions?

vividviolet: Let's just leave it in for now (there's no harm since it would be a no-op) until we can fix the generator or the types to exclude `controls`. I'm looking at it again and our types don't reference `controls` at all so I'm not sure why it would be in the remove definition. Can you make a follow up task Bill?

billfienberg: Follow up task: https://github.com/Shopify/admin-ui-foundations/issues/2169

PR #2663, opened 2025-03-04
Add Sale/Exchange/Return types and update transaction complete.

fatbattk:
### Background

Part of, https://github.com/Shopify/temp-project-mover-Archetypically-20250612122555/issues/34
These changes came from developing with it extensively on POS. Most reflect the data that POS has accessible based on the transaction type (Sale/Exchange/Return).

Did not add documentation since these are all types/inputs and not previously "released". Let me know if otherwise.

Not addressing the new naming in this PR. Instead will be handled via https://github.com/Shopify/retail-hq-store-management/issues/764

### Solution

- Add a `BaseTransactionComplete` that replaces previous `Checkout` type.
- Add 4 input types `ExchangeTransactionInput`, `ReprintReceiptInput`, `SaleTransactionInput`, `ReprintReceiptInput` that extends from `BaseTransactionComplete`.
- Add `TransactionType` type.
- Add `taxLines` to `Cart` type.
- Add `Money` type.
- Add `StoreCredit` to `paymentMethods`.
- Update `ShippingLine` and `TaxLine` type to use `Money` type and optional some fields.
- Remove `Checkout` type.

### 🎩

- ...

### Checklist


packages/ui-extensions/src/surfaces/point-of-sale/input.ts
```diff
@@ -1,4 +1,5 @@
 export type {BaseInput} from './event/input/BaseInput';
+export type {ReprintReceiptInput} from './event/input/ReprintReceiptInput';
```
fatbattk, line 2: `ReprintReceiptInput` is not part of `TransactionCompleteInput` since it is not a sale-related transaction, so exporting it for separate use on POS.

Alex-Palad (approved):

packages/ui-extensions/src/surfaces/point-of-sale/event/input/ExchangeTransactionInput.ts
```diff
@@ -0,0 +1,9 @@
+import {BaseTransactionComplete} from '../../types/base-transaction-complete';
+import {LineItem} from '../../types/cart';
+
+export interface ExchangeTransactionInput extends BaseTransactionComplete {
```
Alex-Palad, line 4: ExchangeTransactionData?

fatbattk: not planning to rename it here, will do mass-rename in https://github.com/Shopify/retail-hq-store-management/issues/764

js-goupil (approved):

PR #2665, opened 2025-03-04
[CI] Update actions/cache and actions/checkout to latest

gwyneplaine:
CI Failing due to actions/cache being out of date (previously on v1, latest is v4).
Bumping of actions/checkout is purely boyscouting.


alex-page (approved):

alex-page:
There are a couple of other GitHub actions still using older versions. We should bump those as well.

PR #2673, opened 2025-03-07
Fio/update types

Fionoble:
### Background

Adds types for form to 2025-07-rc

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components.d.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components.d.ts
@@ -1,4 +1,4 @@
-/** VERSION: 0.39.0 **/
+/** VERSION: 0.39.1 **/
 /* eslint-disable @typescript-eslint/ban-types */
 /* eslint-disable @typescript-eslint/no-namespace */
 /* eslint-disable @typescript-eslint/member-ordering */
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/src/surfaces/admin/components.d.ts
```diff
@@ -6281,6 +6284,38 @@ export interface AdminPrintActionJSXProps
   id?: string;
 }
 
+declare const tagName = 's-form';
```
billfienberg, line 6287: @vividviolet, should `Form`'s `tagName` have a `$*` suffix like all of the other `tagName` declarations?

vividviolet: That's all automated by the build tool. It doesn't have the alias because it doesn't have a collision since all the other ones have the suffix. All components call the tag name `tagName` so some of them need to be renamed to not collide.

vividviolet (approved):

packages/ui-extensions/src/surfaces/admin/components.d.ts
Fionoble: Isn't that just to avoid duplicate vars?

PR #2674, opened 2025-03-10
Deprecated FormattedTextField

js-goupil:
Part of https://github.com/Shopify/temp-project-mover-Archetypically-20250612104008/issues/53

### Background
This TextField is now deprecated, in favour for the TextField component.

### Solution
I took the liberty of renaming `NewTextFieldProps` to `TextField` props since that name is now available without conflict. 

### 🎩
- You can build the package and drop it into the node modules of an existing project's ui-extensions build folder. `yarn build` at the root of the ui-extensions package.
- Make sure the field is marked as deprecated 

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/reference/components/FormattedTextField.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/reference/components/FormattedTextField.doc.ts
@@ -7,7 +7,7 @@ const generateCodeBlockForComponent = (title: string, fileName: string) =>
 const data: ReferenceEntityTemplateSchema = {
   name: 'FormattedTextField',
   description:
-    'Use a formatted text field when you require additional functionality such as the text field input type or a custom validator.',
+    'This component is deprecated. Please use the [`TextField`](/docs/api/pos-ui-extensions/components/textfield) component instead. Use a formatted text field when you require additional functionality such as the text field input type or a custom validator.',
   isVisualComponent: true,
   type: 'component',
   definitions: [
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
@@ -46,6 +46,9 @@ Refer to the [migration guide](/docs/api/pos-ui-extensions/migrating) for more i
 - Removed in POS version: N/A
 - Release day: N/A
 
+### Breaking Changes
+- Renamed \`NewTextFieldProps\` to \`TextFieldProps\`.
+
 ### Features
 
 - Added support for the ${TargetLink.PosTransactionCompleteObserve} target.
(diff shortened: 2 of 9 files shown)


NathanJolly (approved):
Nice, ty.

packages/ui-extensions/src/surfaces/point-of-sale/render/components/TextField/TextField.ts
```diff
@@ -29,8 +29,8 @@ export type EmbeddedElementProps =
   | SuccessProps
   | PasswordProps;
 
-export interface NewTextFieldProps extends InputProps {}
+export interface TextFieldProps extends InputProps {}
```
NathanJolly, line 32: I believe this will be a breaking change, right? Will have to document as such for 2025-04.

NathanJolly (commented):
Sorry, I approved then realized this is a breaking change and needs to be noted in the version doc.

packages/ui-extensions/src/surfaces/point-of-sale/render/components/TextField/TextField.ts
js-goupil: Already working on a docs PR 😇

js-goupil:
@NathanJolly Good catch. Applied in latest commit. This is what it looks like 
Screenshot 2025-03-10 at 1 47 45 PM

packages/ui-extensions/docs/surfaces/point-of-sale/reference/components/FormattedTextField.doc.ts
```diff
@@ -7,7 +7,7 @@ const generateCodeBlockForComponent = (title: string, fileName: string) =>
 const data: ReferenceEntityTemplateSchema = {
   name: 'FormattedTextField',
   description:
-    'Use a formatted text field when you require additional functionality such as the text field input type or a custom validator.',
+    'This component is deprecated. Please use the [`TextField`](/docs/api/pos-ui-extensions/components/textfield) component instead. Use a formatted text field when you require additional functionality such as the text field input type or a custom validator.',
```
js-goupil, line 10: @vctrchu do you remember if there's a way to mark the component as deprecated? ie some yellow label shows up to make it more obvious?

js-goupil: Nope doesn't look like it

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
```diff
@@ -46,6 +46,9 @@ Refer to the [migration guide](/docs/api/pos-ui-extensions/migrating) for more i
 - Removed in POS version: N/A
 - Release day: N/A
 
+### Breaking Changes
```
vctrchu, line 49: Screenshot 2025-03-10 at 12 59 35 PM
I don't see this here?

js-goupil: We removed the deprecation from breaking changes because it's not actually a breaking change

vctrchu (approved):

PR #2681, opened 2025-03-12
Add `QRCode` component for POS

fatbattk:
### Background

Resolves https://github.com/Shopify/retail-hq-store-management/issues/768

### Solution

- Add `QRCode` component + docs + thumbnail + examples.
  - **NOTE:** only property is `value: string`. Can expand in future if necessary.
- Replace `Image` with `QRCode` as acceptable children for `POSReceiptBlock`.
- Add to version log as `2025-04` developer preview.
- Update targets type signature for `pos.receipt-footer.block.render` with a type (vs api).
  - **NOTE:** type signature for `transaction` is not surfaced in the documentations. This needs to be done, BUT WHERE since it's not an API.
  - ~(_unrelated to PR_) There's underlying typing issues with `useApi`, etc where autocomplete just return the full API list even though it's not in that target. 🤷🏻~

### 🎩

image

### Checklist


Alex-Palad (approved):

packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
```diff
@@ -131,7 +136,7 @@ export interface ExtensionTargets {
     BlockComponents
   >;
   'pos.receipt-footer.block.render': RenderExtension<
-    StandardApi<'pos.receipt-footer.block.render'> & OrderApi,
+    StandardApi<'pos.receipt-footer.block.render'> & TransactionCompleteData,
```
js-goupil:
```suggestion
   TransactionCompleteData,
```

We need to remove the StandardApi because right now it's adding all of the extra sessionApi, toastApi, and a bunch of things that do "mutation" things to POS (that's why we consider that "API"). Even though this target renders and behaves a bit like a UI Extension Target, it's intended to be static, so we only need `TransactionCompleteData`, which extends `BaseData` and therefore will pass in connectivity, session, and other base objects

Screenshot 2025-03-13 at 1 26 22 PM

js-goupil (changes requested):
Just based off of the StandardApi being included, we don't need it here, nor do we want it because this target still behaves like an "event" target even though it "renders". It's a weird hybrid I guess...

packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
Alex-Palad: Oh, and also, it is `TransactionCompleteData | {transaction: ReprintReceiptData}`

fatbattk: ok, removed!
that's interesting now my autocomplete shows `transaction` too. wasn't able to get it yesterday 🤔

Alex-Palad: You'll also need {key: string} to make TS happy
```js
'pos.customer-details.block.render': RenderExtension<
    {
      [key: string]: any;
    } & (TransactionCompleteData | {transaction: ReprintReceiptData}),
    BlockComponents
  >;
```

js-goupil (approved):

packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
fatbattk: good call on the `ReprintReceiptData`, but trying to avoid the `any` object since that is a hack for the current `useApi` usage.

looks like there's issues with the typing after all. will try to find solution for abit if not just mass allow like that.

PR #2677, opened 2025-03-12
Use correct props for DateField

MitchLillie:
### Background

DateField was using the DatePicker props, which is incorrect.

[We heard about this on slack here.](https://shopify.slack.com/archives/C04K7BZDH3N/p1741687157987469)

### Solution

(Describe your solution, why this approach was chosen, and what the alternatives/impacts may be)

### 🎩

- ...

### Checklist

- [ ] I have :tophat:'d these changes
- [ ] I have updated relevant documentation

PR #2683, opened 2025-03-13
Update DateField examples

MitchLillie:
### Background

Follow up to https://github.com/Shopify/ui-extensions/pull/2677 and actually update the examples.

### Solution

### 🎩

- ...

### Checklist


packages/ui-extensions/src/surfaces/admin/components/DateField/examples/range-datefield.example.ts
```diff
@@ -3,7 +3,10 @@ import {extend, DateField} from '@shopify/ui-extensions/admin';
 extend('Playground', (root) => {
   const dateField = root.createComponent(
     DateField,
-    { selected: {start: '2023-11-08', end: '2023-11-10' } },
+    {
+      label: 'DateField',
+      value: {start: '2023-11-08', end: '2023-11-10'},
```
vividviolet: This isn't valid for DateField - it expects only a string. We should delete this example.

packages/ui-extensions-react/src/surfaces/admin/components/DateField/examples/range-datefield.example.tsx
```diff
@@ -2,15 +2,21 @@ import React from 'react';
...
   render,
   DateField,
-  type Selected
 } from '@shopify/ui-extensions-react/admin';
 
 render('Playground', () => <App />);
 
 function App() {
-  const [selected, setSelected] = React.useState<Selected>({start: '2023-11-08', end: '2023-11-10' });
+  const [selected, setSelected] = React.useState({
```
vividviolet: Same here, this isn't supported

packages/ui-extensions/src/surfaces/admin/components/DateField/examples/range-datefield.example.ts
MitchLillie: Ah. I was checking in the lit components directly which do seem to support these. But yes apparently the actual component does not support it. I definitely remember how all of this works 😅

vividviolet (approved):

PR #2700, opened 2025-03-20
Add missing docs for Checkbox, OrderedList, UnorderedList, PasswordField, Select and URLField

alex-page:
Audited the shipped components and updated the documentation to include complete components

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/admin/create-doc-files.sh
+++ b/packages/ui-extensions/docs/surfaces/admin/create-doc-files.sh
@@ -45,7 +45,15 @@ for componentName in "$@"; do
     if [ ! -f "$tsx_example" ]; then
         cat << EOF > "$tsx_example"
   <s-${lowercaseComponentName}></s-${lowercaseComponentName}>
-}
+EOF
+        echo "Created TSX example file for ${componentName}"
+    fi
+
+    # Create HTML example file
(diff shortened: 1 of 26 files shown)


alex-page:
@fio I would like to make some changes tomorrow after going through generating docs:
- [ ] The examples are duplicated multiple times across `basic-badge.example.html` `basic-badge.example.tsx` and `preview.html` we should be able to consolidate these
- [ ] We shouldn't need to add CSS to each example. We should have a live code example that adds some styles for our implementation instead of forcing us to add the entire HTML. We could resolve this by having a wrapper and children that get injected into the iframe. This will help with: center alignment by default for components, consistent whitespace around the components, updating the CDN once instead of multiple times.
- [ ] When running `gen-docs:admin` then `yarn docs:ab` an error is shown from `Unable to find code file, '../../../../../../ui-extensions-react/src/surfaces/admin/components/Checkbox/examples/basic-checkbox.example.html`. We should update this script to work

Fionoble (approved):

Fionoble:
The Checkbox one is weird, it looks like it used the old generator logic for some bizarre reason. 

Yea, I'd love to eventually get rid of the redundancies of the examples/previews. However, JSX is not exactly HTML so in some cases (Clickable is a decent example). However it is a superset so we could do some conversion while processing it. Jason had some gists he was willing to share. I'd prefer to clean this up once things are a bit more stable though.

PR #2706, opened 2025-03-24
Clean up and consolidate documentation

alex-page:
### Background

Follow the following guidance for examples and consolidate duplicate code as much as possible.

Do: Follow existing example content on [polaris.shopify.com](http://polaris.shopify.com/)
Do: Create examples of UI that fits in the Admin. <s-banner tone="info">USPS has updated their rates</s-banner>
Do: Showcase what is possible with the component in the default example
Do: Avoid complexity and ensure all examples are entry level (adding script tags, regex and more. Leave complexity to tutorials)
Don't: Use words like "simple" or "basic" to explain concepts or examples
Don't: Explain the component or it's state in the example (the code already does this) <s-banner tone="info">This is a banner with info tone</s-banner>
Don't: Create examples that do not follow Admin Design Guidelines (interactive Badges)
Don't: Add unnecessary composition of components. Ensure the component demoed is the focus. (exception would be parent child relationships)

### Solution

- [x] Consolidate language across examples to be `jsx`
- [x] Remove "Simple" and "Basic"
- [x] Consolidate default example titles to be "Default Example"
- [x] Consolidate example code into `${component}/example/default.html`
- [x] JSX examples use same markup as HTML by default
- [x] Update link in `build-ab-docs.sh`
- [x] Update scaffolding in `create-doc-files.sh` to align with above changes
- [x] Remove all `basic-${component}.example.html` files
- [x] Update `tsconfig` to exclude `examples` directory instead of `*.example.*` files
- [x] Add `"htmlWhitespaceSensitivity": "ignore",` to `.prettierrc` to ensure HTML example whitespace is retained
- [x] Remove `relative` and `external` examples from `Link` (same as default)
- [x] Update examples to use real content from the Admin (like polaris.shopify.com)
- [x] Added three example templates linked to the `language` (follow up PR here to decouple it https://github.com/Shopify/generate-docs/pull/332

### Checklist


packages/ui-extensions/src/surfaces/admin/components/MoneyField/examples/default.html
```diff
@@ -0,0 +1,7 @@
+<s-money-field
+  label="Regional Price"
+  placeholder="99.99"
+  currencyCode="EUR"
+  details="Recommended price for the European market"
+/>
+/>
```
Fionoble:
```suggestion
```

PR #2709, opened 2025-03-25
Revert "Deprecated FormattedTextField"

js-goupil:
Reverts Shopify/ui-extensions#2674

We are reverting because we are not shipping breaking changes to 2025-04


packages/ui-extensions-react/src/surfaces/point-of-sale/components/FormattedTextField/FormattedTextField.ts
```diff
@@ -1,9 +1,6 @@
 import {FormattedTextField as BaseFormattedTextField} from '@shopify/ui-extensions/point-of-sale';
 import {createRemoteReactComponent} from '@remote-ui/react';
 
-/**
- * @deprecated Use the `TextField` component instead.
- */
```
NathanJolly: This is not a breaking change and can be left in.

NathanJolly (approved):
Can leave the deprecations in

NathanJolly (approved):

PR #2714, opened 2025-03-26
[Customer][EXT] add country to customer account Standard API.

ottocho:
### Background

part of https://github.com/shop/issues-customer-accounts-and-login/issues/402

> As a 3P developer, I can access the current country of a customer, so I can implement country-specific logic in my extension blocks.

### Solution

Data will be provided by PR here https://github.com/shop/world/pull/24651

We will add country to localization in unstable.

### 🎩

[see this PR](https://github.com/shop/world/pull/24651)

### Checklist


packages/ui-extensions/src/shared.ts
```diff
@@ -859,6 +859,14 @@ export type CountryCode =
   | 'ZW'
   | 'ZZ';
 
+export interface Country {
```
robin-drexler, line 863: We should talk to c1 folks about this because they have a copy of that package in their repo and might accidentally override stuff when they copy theirs over. 😅  cc @jamesvidler

robin-drexler (approved):

packages/ui-extensions/src/shared.ts
jamesvidler: Yes, the changes to the checkout surface here would need to be manually applied to [checkout-web](https://github.com/shop/world/tree/main/areas/clients/checkout-web/packages/ui-extensions/src/surfaces/checkout) as well so that we don't overwrite these.

When customer-account eventually moves over to the monorepo, I wonder if we can all share the same private package (which is useful as a staging ground) and then we don't need to worry about this stuff anymore 😅.

packages/ui-extensions/src/surfaces/customer-account/api/standard-api/standard-api.ts
```diff
@@ -161,6 +162,13 @@ export interface Localization {
    * extension (that is, the one matching your .default.json file).
    */
   extensionLanguage: StatefulRemoteSubscribable<Language>;
+
+  /**
+   * The country context of the buyer sees in the customer account.
+   * It will update if the buyer changes the country in the customer account
+   * If the country is unknown, then the value is undefined.
+   */
+  country: StatefulRemoteSubscribable<Country | undefined>;
```
brianshen1990, line 171: hm.. seems in this PR https://github.com/shop/world/pull/24651 Micha thinks we should add it to buyer identity part..I will have another review once the concern is resolved

packages/ui-extensions/src/shared.ts
ottocho: good news is that customer-account has been in monorepo since yesterday!

packages/ui-extensions/src/surfaces/customer-account/api/standard-api/standard-api.ts
ottocho: that's not _buyer identity_. Misha was talking about `useBuyerContextData` where we get to know what country the customer is currently active on customer account.

packages/ui-extensions/src/shared.ts
jamesvidler: Oh snap! I had no idea 🥳

ottocho: I wanted to confirm my understanding and determine what I should test next.

1. I copied the source from the ui-extensions to checkout-web ([draft like this](https://github.com/shop/world/pull/25319)), but I noticed quite many changes. It seems like it might not be up to date. How often do we manually sync changes between these two?

2. Should I only apply the changes I made here to checkout-web and test to see if everything works correctly? However, this approach doesn't seem very easy given the amount of manual testing involved.

jamesvidler: > I copied the source from the ui-extensions to checkout-web (https://github.com/shop/world/pull/25319), but I noticed quite many changes. It seems like it might not be up to date. How often do we manually sync changes between these two?

I would only copy over the specific changes you are introducing here that are in the `checkout` surface. We haven't brought over changes from the private checkout to the public packages in a bit as there's a bunch of remote-dom work in there. We'll be porting over "some" hand picked changes from our private package to the public package early next week (see [slack thread](https://shopify.slack.com/archives/C0683EBQ55M/p1743104873072919?thread_ts=1742825573.749629&cid=C0683EBQ55M) for context).

> Should I only apply the changes I made here to checkout-web and test to see if everything works correctly? However, this approach doesn't seem very easy given the amount of manual testing involved.

Yes, only bring over changes you are making within the `checkout` surface and the shared code you have. You shouldn't really need to test anything here as its just re-arranging some types.

ottocho: I have this minor PR in checkout-one for this change. thanks.

https://github.com/shop/world/pull/26722

brianshen1990 (approved):

.changeset/funny-dragons-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
```
brianshen1990, line 2: you also updated `@shopify/ui-extensions-react`, do you want to add it to the changeset?

ottocho: thanks right let me adjust this

jamesvidler (approved):

PR #2729, opened 2025-04-02
Simplify docs

alex-page:
A part of https://github.com/Shopify/admin-ui-foundations/issues/2534 and https://github.com/Shopify/admin-ui-foundations/issues/2442

This PR does a number of changes:
- [x] Consolidate Admin Doc generation into one script (ensures we CI test both scripts as well)
- [x] Consolidate examples into one non-interactive example
- [x] Replace shell script with NodeJS
- [x] Only adds `member.isOptional = true;` if it is a `entry.isVisualComponent`
- [x] Generates the live preview and html code tab from the example provided
- [x] Removes the `JSX` and `HTML` tab titles
- [x] Groups the code into reusable functions
- [x] Only reads the generated_docs_json file once to make all necessary modifications
- [x] Copies the final result of `generated` directory to `shopify-dev` (instead of each file individually)
- [x] Adds a `AdminReferenceEntityTemplateSchema` which includes `layout` (which is removed on build) which is used to generate live template code
- [x] Update create docs template
- [x] Removes all events from example code

Follow up PRS:
- [x] Update docs in `shopify-dev` in the `polaris-docs-2025` branch https://github.com/Shopify/shopify-dev/pull/55853
- [x] Update the `docs:sync` script in `extensibility` https://github.com/Shopify/extensibility/pull/350

Future ideas:
- [ ] Consolidate the `.docs.ab.ts` and `.docs.ts` files for Admin. We don't get a lot of benefit from both of them. We just need a way to say `Page` is App Bridge only and override values for App Bridge if necessary. The `tsconfig` does a lot of this work already.


package.json
```diff
@@ -14,7 +14,6 @@
     "predeploy": "yarn build",
     "deploy": "changeset publish",
     "predeploy:unstable": "yarn build",
-    "docs:ab": "yarn workspace @shopify/ui-extensions docs:ab",
```
Fionoble, line 17: FYI: Removing this will also break the `docs:sync` command in extensibility.

packages/ui-extensions/src/surfaces/admin/globals.ts
```diff
@@ -14,3 +18,16 @@ declare global {
   // @ts-ignore
   const shopify: ShopifyGlobal;
 }
+
+export interface AdminReferenceEntityTemplateSchema
```
Fionoble: We should move this to a separate file. This is the global object that gets exposed to developers using admin-extensions. We shouldn't add doc specific types in here. Maybe a `docs-types.ts` or something.

packages/ui-extensions/docs/surfaces/admin/build-docs.mjs
```diff
@@ -0,0 +1,254 @@
...
+  await fs.writeFile(filePath, JSON.stringify(jsonData, null, 2));
+};
+
+const generateFiles = async (tsconfig, outputDir, isExtensions = true) => {
+  const scripts = [
+    `yarn tsc --project ${docsRelativePath}/${tsconfig} --moduleResolution node  --target esNext  --module CommonJS`,
+    `yarn generate-docs --input ./${srcRelativePath} --typesInput ./${srcRelativePath} --output ./${outputDir}`,
+  ];
+
+  if (isExtensions) {
```
Fionoble, line 153: We aren't checking for the args in this file so it will always default to unstable. We need to make it 2025-07-rc

alex-page: We get the version on line 8 https://github.com/Shopify/ui-extensions/blob/82ad833438a0625d8a2fd466e3708463b9abb47a/packages/ui-extensions/docs/surfaces/admin/build-docs.mjs#L8

I tophatted it with `yarn docs:admin 2025-07-rc` and it worked fine. Let me know if I am missing something.

PR #2734, opened 2025-04-02
Add `Discount.currency` optional property

fatbattk:
### Background

We should support multi-currencies. On POS, discounts can be multi-currency. Updating for `Discount` interface used by event targets.

### Solution

- Add optional `currency` type to existing `Discount` interface.
- Add to version notes under `2025-07`.
- Did not use existing [Money](https://github.com/Shopify/ui-extensions/blob/unstable/packages/ui-extensions/src/surfaces/point-of-sale/types/money.ts) or made required because it will break typing for existing users of `Discount`. Should it be a breaking change though?

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
@@ -39,6 +39,20 @@ Refer to the [migration guide](/docs/api/pos-ui-extensions/migrating) for more i
     },
     {
       type: 'Generic',
+      anchorLink: '202507',
+      title: '2025.07',
+      sectionContent: `
+- Added in POS version: 10.6
+- Removed in POS version: N/A
+- Release day: N/A
+
+### Features
+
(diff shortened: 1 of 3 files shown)


Alex-Palad (approved):

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
vctrchu, line 1: As we discussed, since this missed the 2025-04 cut-off. We will have to put the optional currency notes under 2025-07 version docs

fatbattk: updated.
> Added in POS version: 10.6

also is this accurate? referred the POS release schedule.

vctrchu (approved):

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
vctrchu: Yes

PR #2747, opened 2025-04-08
Add type defs for Switch

kyledurand:


packages/ui-extensions/src/surfaces/admin/components.d.ts
```diff
@@ -1,4 +1,4 @@
-/** VERSION: 0.46.0 **/
+/** VERSION: 0.47.2 **/
```
kyledurand, line 1: Odd version bump here. Any idea why @vividviolet ?

sam-b-rose: This corresponds to the current Admin UI components version. This looks correct!

sam-b-rose (approved):

packages/ui-extensions/src/surfaces/admin/components.d.ts
kyledurand: Ah that makes sense

PR #2759, opened 2025-04-10
Prepare types/docs for 2025-07-rc for customer accounts

oluwatimio:
### Background

This PR is a starting point for our 2025-07-rc

There are 2 commits
- first commits added by @oluwatimio , all api, conponents, targets needed for 2-25-07
- second commit, added by @brianshen1990
  - make sure there is no type issue => checkout is not there yet, so we have to copy some types
  - make sure we compile global types so that developers can use them

### tophat

Demo video: https://screenshot.click/11-31-f69tx-dsff7.mp4 

Use any existing extension
1. update `api_version = "2025-07"` in toml file
2. update `ui-extension` version to `"@shopify/ui-extensions": "0.0.0-snapshot-20250411132127",` 
3. run `yarn install` on app root
4. touch a file called `shopify.d.ts` on app root
5. add something like this based on your own test extension
```
import '/Users/brianshen/remotedomhey/node_modules/@shopify/ui-extensions/index.js';

//@ts-ignore
declare module './extensions/remotedombughunt/src/OrderStatusBlock.jsx' {
  const globalThis: typeof import('./node_modules/@shopify/ui-extensions/build/ts/surfaces/customer-account/targets/customer-account.order-status.block.render.d.ts');
  const shopify: import('./node_modules/@shopify/ui-extensions/build/ts/surfaces/customer-account/targets/customer-account.order-status.block.render.d.ts').Api;
}

//@ts-ignore
declare module './extensions/remotedombughunt/src/ProfileBlock.jsx' {
  const globalThis: typeof import('./node_modules/@shopify/ui-extensions/build/ts/surfaces/customer-account/targets/customer-account.profile.block.render.d.ts');
  const shopify: import('./node_modules/@shopify/ui-extensions/build/ts/surfaces/customer-account/targets/customer-account.profile.block.render.d.ts').Api;
}
``` 
6. go to extension code, add `/// <reference types="../../../shopify.d.ts" />` at the begining of the file
7. refresh ts server, now you can see we have all types 

### Checklist


packages/ui-extensions/docs/surfaces/customer-account/reference/apis/customer-privacy.doc.ts
```diff
@@ -0,0 +1,74 @@
+import type {ReferenceEntityTemplateSchema} from '@shopify/generate-docs';
```
oluwatimio, line 1: We still have a lot of our examples to update

packages/ui-extensions/loom.config.ts
```diff
@@ -42,6 +42,13 @@ export default createPackage((pkg) => {
             );
             completedSurfaces.add('admin');
           }
+          if (!completedSurfaces.has('customer-account')) {
+            buildTargetsDefinitions(
+              resolve(process.cwd(), 'packages/ui-extensions'),
+              'customer-account',
+            );
+            completedSurfaces.add('customer-account');
+          }
```
brianshen1990, line 51: add customer account

packages/ui-extensions/src/shared.ts
```diff
@@ -860,6 +860,14 @@ export type StorefrontApiVersion =
...
   | 'unstable';
 
+export interface Country {
+  /**
+   * The ISO-3166-1 code for this country.
+   * @see https://www.iso.org/iso-3166-country-codes.html
+   */
+  isoCode: CountryCode;
+}
+
```
brianshen1990, line 870: from unstable version

packages/ui-extensions/buildTargetDts.ts
```diff
@@ -159,22 +159,39 @@ function updateReferences(
...
 }
 
+function getTargets(sourceFile: SourceFile, surface: string) {
+  if (surface === 'customer-account') {
+    return [
+      sourceFile.getInterface('OrderStatusExtensionTargets')!,
+      sourceFile.getInterface('CustomerAccountExtensionTargets')!,
+    ];
+  }
+  return [sourceFile.getInterface('ExtensionTargets')!];
```
brianshen1990: customer account `ExtensionTargets` is a type and it is empty, it is just  `OrderStatusExtensionTargets |  CustomerAccountExtensionTargets` , it has no property.

packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
```diff
@@ -1,8 +1,19 @@
...
+  onPress?(): void;
+  to?: string;
+  loading?: boolean;
+  loadingLabel?: string;
+  disabled?: boolean;
+  accessibilityLabel?: string;
+  accessibilityRole?: 'button' | 'submit';
+  overlay?: any;
+  kind?: 'primary' | 'secondary' | 'plain';
+}
```
brianshen1990, line 16: add it here manually since checkout is not here

packages/ui-extensions/src/surfaces/customer-account/components/CustomerAccountAction.d.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export interface CustomerAccountActionProps {
+  /**
+   * Sets the heading of the Action container.
+   */
+  heading: string;
+  /**
+   * Sets the Primary action button of the container. This component must be a button component.
+   */
+  primaryAction?: ComponentChild;
```
brianshen1990: updated this to `ComponentChild` same thing as admin https://github.com/Shopify/admin-ui-foundations/blob/f2b76887e6e40e46e9c18d9e36aab4a28aa38254/admin-ui-components/src/ui-extensions-components/AdminAction/meta.ts#L11-L12

packages/ui-extensions/src/surfaces/customer-account/components/shared.ts
```diff
@@ -4,3 +4,11 @@ export interface IdProps {
    */
   id?: string;
 }
+
+export type Size =
```
brianshen1990, line 8: manually copied here from checkout since it is not here

packages/ui-extensions/src/surfaces/customer-account/extension-targets.ts
```diff
@@ -40,7 +36,7 @@ export interface OrderStatusExtensionTargets {
   'customer-account.order-status.block.render': RenderExtension<
     OrderStatusApi<'customer-account.order-status.block.render'> &
       StandardApi<'customer-account.order-status.block.render'>,
-    AnyComponent
+    StandardComponents
```
brianshen1990, line 40: switched to `StandardComponents` as checkout components are not ready in this branch

packages/ui-extensions/buildTargetDts.ts
vividviolet: Admin's is also a type - it should work the same way for Customer Account? We should avoid hardcoding things here if we can.

brianshen1990: For admin, it is 
```
export interface ExtensionTargets {
  'admin.customers.segmentation-templates.render':  ...
  ...
}
```

for customer account, we are
```
export type ExtensionTargets = OrderStatusExtensionTargets &
  CustomerAccountExtensionTargets;
export interface OrderStatusExtensionTargets {
  'customer-account.order-status.block.render':...
  ...
}
export interface CustomerAccountExtensionTargets {
  'customer-account.page.render': ...
   ...
} 
```

so
1 `getInterface` will only get interface value. we are using `type`, so we will get nothing from the interface
2. when we `getProperty` , admin's  ExtensionTargets have property directly.  And even if I update customer account's ExtensionTargets to an interface, it has no DIRECT property, and it will lead to more complex and inconsistent code here.

i prefer we keep the logic intact, and just add intended target interfaces for each surface like what I did in this PR.

vividviolet: I think we should just change the tool to account for this use case. You can do these small changes:
```ts
export interface ExtensionTargets
  extends OrderStatusExtensionTargets,
    CustomerAccountExtensionTargets {}
```

```ts
function getTargets(
  sourceFile: SourceFile,
  _surface: string,
): InterfaceDeclaration[] {
  const base = sourceFile.getInterface('ExtensionTargets')!;
  const structure = base.getStructure();
  const allInterfaces = [base];

  if (Array.isArray(structure?.extends)) {
    structure.extends.forEach((inherited) => {
      if (typeof inherited === 'string') {
        allInterfaces.push(sourceFile.getInterface(inherited)!);
      }
    });
  }

  return allInterfaces;
}
```

brianshen1990: let me have a look, thanks ~~

brianshen1990: updated!

jamesvidler (approved):
I tophatted these changes to the `buildTargetDts.ts` and `loom.config.ts` for the [checkout surface as well](https://github.com/Shopify/ui-extensions/pull/2765) and got the global API types working.

packages/ui-extensions/src/surfaces/customer-account/components/CustomerAccountAction.d.ts
vividviolet: Do you want these to actually be components as props? We only have these in Admin to generate the remote definition as "slots". These props are not removed from the custom elements properties as seen [here](https://github.com/Shopify/ui-extensions/blob/9adfc12838a0a713e16086d6f33bf3072919a20a/packages/ui-extensions/src/surfaces/admin/components/AdminAction.d.ts#L98)

packages/ui-extensions/src/surfaces/customer-account/components/Page/examples/basic-Page-js.example.ts
```diff
@@ -0,0 +1,71 @@
...
+      },
+      'Secondary action 3',
+    ),
+  );
+
+  const breadcrumbActions = root.createFragment();
+  await breadcrumbActions.append(
+    root.createComponent(Button, {
+      accessibilityLabel: 'Button',
+      onPress: () => {},
```
vividviolet, line 55: These are still `onPress`? I thought these have all been updated to `onClick`?

vividviolet (commented):
Looking good! Just a few questions

packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
oluwatimio: ButtonProps have changed but im guessing we will update this later

packages/ui-extensions/src/surfaces/customer-account/components/CustomerAccountAction.d.ts
oluwatimio: I think normally should be a remote fragment. Do we want to use ComponentChild instead?

oluwatimio: Also I believe the components are not props instead they'd be slots

packages/ui-extensions/src/surfaces/customer-account/components/Menu/Menu.doc.ts
```diff
@@ -0,0 +1,80 @@
+import type {ReferenceEntityTemplateSchema} from '@shopify/generate-docs';
+
+const data: ReferenceEntityTemplateSchema = {
```
oluwatimio: Do we want to keep Menu examples @brianshen1990?

packages/ui-extensions/src/surfaces/customer-account/components/Page/examples/basic-Page-js.example.ts
oluwatimio: Button has been updated @brianshen1990 that being said, we dont have checkout's changes here yet and since this is a checkout component. We should probably update this soon especially once their changes are here

packages/ui-extensions/src/surfaces/customer-account/api/docs.ts
brianshen1990: we will switch to use checkout's button once it is ready in rc branch

packages/ui-extensions/src/surfaces/customer-account/components/Page/examples/basic-Page-js.example.ts
brianshen1990: use checkout's types and updating examples will follow up!

packages/ui-extensions/src/surfaces/customer-account/components/Menu/Menu.doc.ts
brianshen1990: let me remove them and the re-add them later!

packages/ui-extensions/src/surfaces/customer-account/components/CustomerAccountAction.d.ts
brianshen1990: hey @vividviolet , 

> These props are not removed from the custom elements properties as seen [here](https://github.com/Shopify/ui-extensions/blob/9adfc12838a0a713e16086d6f33bf3072919a20a/packages/ui-extensions/src/surfaces/admin/components/AdminAction.d.ts#L98)

sorry I get a little bit confused, are you intended to say: these props are removed from the custom elements properties

vividviolet: Sorry, i meant to say
> These props are ~not~ removed from the custom elements properties as seen [here](https://github.com/Shopify/ui-extensions/blob/9adfc12838a0a713e16086d6f33bf3072919a20a/packages/ui-extensions/src/surfaces/admin/components/AdminAction.d.ts#L98)

oluwatimio: > We only have these in Admin to generate the remote definition as "slots"

This is what we do as well. They are slots https://github.com/shop/world/blob/main/areas/clients/customer-account-web/app/foundation/extensibility/sandbox/elements/CustomerAccountAction.ts#L21

So we wont use this as a prop

brianshen1990: I remove them ahha~~

brianshen1990: discussed with @oluwatimio , we will add it to component.ts and them omit it here once checkout button is ready so that we can have a proper test~

brianshen1990 (approved):

PR #2765, opened 2025-04-11
Prepare `shopify` global API types per target for checkout 2025-07-rc

jamesvidler:
### Background

Closes https://github.com/Shopify/temp-project-mover-Archetypically-20250512095102/issues/160.

See this 🎥 [video for more details](https://share.descript.com/view/xSqizp66xpj).

Checkout must expose types in a specific way for them in be automatically included in this TS reference `shopify.d.ts` and it should be specific to the target that the extension is rendering in:
Screenshot 2025-04-11 at 3 46 12 PM

### Solution

Since the `2025-07-rc` doesn't yet have the `checkout` surface included, this PR needed to add a current cut of the checkout surface and make some adjustments.

**Note**: I excluded the components entirely since they still need some work and must provide their own `<component>.d.ts` files so that their aren't any dependencies on the `@shopify/ui-api-design` repo. This would align with how admin and customer account are doing this. See this [issue](https://github.com/Shopify/temp-project-mover-Archetypically-20250512095102/issues/188) for additional context.

Some changes are being [backported to checkout-web@main](https://github.com/shop/world/pull/35580).

### 🎩
1. Follow [these steps](https://docs.google.com/document/d/1W-AeJl1v281512LYRAhpAG9FH8RsTXqfKX2zhrjS6jQ/edit?tab=t.by6i7l7z98ds#heading=h.lr188tolgi4o) to generate a remote-dom extension.

2. Then, change the `package.json` for the extension to use:
```typescript
"@shopify/ui-extensions": "0.0.0-snapshot-20250411192040"
```

3. Run `shopify app dev` or `shopify app build` to ensure that the `/// <reference types="../../../shopify.d.ts" />` file is automatically added to the top of the extension code.

### Checklist


packages/ui-extensions/buildTargetDts.ts
```diff
@@ -22,6 +22,10 @@ function copyComponentDefinitions({
   buildPath: string;
   surface: string;
 }) {
+  if (surface === 'checkout') {
+    // HACK: Checkout doesn't yet have `<component>.d.ts` files, so we can't process them
+    return true;
+  }
```
jamesvidler, line 28: Allow this to run even if we don't get the types for components.

packages/ui-extensions/buildTargetDts.ts
```diff
@@ -159,22 +163,44 @@ function updateReferences(
...
+    structure.extends.forEach((inherited) => {
+      if (typeof inherited === 'string') {
+        allInterfaces.push(sourceFile.getInterface(inherited)!);
+      }
+    });
+  }
+
+  return allInterfaces;
+}
+
```
jamesvidler: Same fix applied in this [customer-accounts PR](https://github.com/Shopify/ui-extensions/pull/2759). Thank you @brianshen1990 🙏.

packages/ui-extensions/loom.config.ts
```diff
@@ -16,7 +16,7 @@ export default createPackage((pkg) => {
   const completedSurfaces = new Set<string>();
   pkg.entry({root: './src/index.ts'});
   pkg.entry({name: 'preact', root: './src/preact.ts'});
-  // pkg.entry({name: 'checkout', root: './src/surfaces/checkout.ts'});
+  pkg.entry({name: 'checkout', root: './src/surfaces/checkout.ts'});
```
jamesvidler: Enable the checkout surface to build target defs

packages/ui-extensions/package.json
```diff
@@ -37,6 +43,15 @@
...
     },
+    "./checkout": {
+      "types": "./build/ts/surfaces/checkout.d.ts",
+      "esnext": "./build/esnext/surfaces/checkout.esnext",
+      "import": "./build/esm/surfaces/checkout.mjs",
+      "require": "./build/cjs/surfaces/checkout.js"
+    },
+    "./purchase.*": {
+      "default": "./build/ts/surfaces/checkout/targets/purchase.*.d.ts"
+    },
```
jamesvidler, line 60: Checkout targets all start with `purchase.*` and not `checkout.*`. We will not support types for legacy targets such as `Checkout::Dynamic::Render`.

packages/ui-extensions/package.json
```diff
@@ -51,6 +66,8 @@
     "./build/cjs/preact.js"
   ],
   "devDependencies": {
+    "@remote-ui/async-subscription": "^2.1.16",
+    "@remote-ui/core": "^2.2.5",
```
jamesvidler: Required for some type references, mainly for remote subscribables.

packages/ui-extensions/src/surfaces/checkout/components.ts
jamesvidler, line 1: Removed components because I couldn't get this package to build since it had a dependancy on the ui-api-design package.

packages/ui-extensions/src/surfaces/checkout/shared.ts
```diff
@@ -1,11 +1,74 @@
-import type {ComponentsBuilder, AnyComponentBuilder} from '../../shared';
+export type AnyComponent =
+  | 'Abbreviation'
+  | 'Badge'
+  | 'Banner'
+  | 'BlockLayout'
+  | 'BlockSpacer'
+  | 'BlockStack'
```
jamesvidler, line 7: To satisfy TypeScript, these need to be a union of strings.

packages/ui-extensions/src/surfaces/checkout/extension-targets.ts
jamesvidler, line 1: Need to rename this specifically to `extension-targets.ts` so that the `buildTargetDts` script can find it.

packages/ui-extensions/src/surfaces/checkout/extension-targets.ts
```diff
@@ -680,8 +680,9 @@ export interface RunnableExtensionTargets {
   >;
 }
 
-export type ExtensionTargets = RenderExtensionTargets &
-  RunnableExtensionTargets;
+export interface ExtensionTargets
+  extends RenderExtensionTargets,
+    RunnableExtensionTargets {}
```
jamesvidler, line 685: The `buildTargetDts` script expects `ExtensionTargets` to be an interface.

packages/ui-extensions/docs/surfaces/checkout/build-docs.sh
```diff
@@ -92,5 +92,15 @@ if [ -d $SHOPIFY_DEV_PATH ]; then
...
+}
+
+if [ -d $SHOPIFY_DEV_PATH ]; then
+  copy_generated_docs_to_shopify_dev
+else
+  # We could be in the monorepo and need to go up several more directories to find shopify-dev
+  SHOPIFY_DEV_PATH="../../../../../../shopify-dev"
+  copy_generated_docs_to_shopify_dev
 fi
+
```
brianshen1990, line 106: will follow the same pattern later~

brianshen1990 (approved):

packages/ui-extensions/src/surfaces/checkout/globals.ts
```diff
@@ -0,0 +1,16 @@
...
+  ): void;
+  reload(): void;
+}
+
+declare global {
+  // conflicts with build/ts/globals.d.ts
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  const shopify: ShopifyGlobal;
+}
```
kumar303, line 16: I don't think this is right. Instead, I'm thinking the generated `shopify.d.ts` would define`ShopifyGlobal`. For example:

```ts
declare module './extensions/checkout-ui/src/Checkout.jsx' {
  const shopify: import('module-path-to/surfaces/checkout/targets/purchase.checkout.block.render.d.ts').Api;
}
```

^ In this case, `Api` should define the `extend()` and `reload()` properties by inheriting some base object that's shared by all targets. What do you think?

packages/ui-extensions/docs/surfaces/checkout/build-docs.sh
```diff
@@ -92,5 +92,15 @@ if [ -d $SHOPIFY_DEV_PATH ]; then
...
-  echo "Not copying docs to shopify-dev because it was not found at $SHOPIFY_DEV_PATH."
+    echo "Not copying docs to shopify-dev because it was not found at $SHOPIFY_DEV_PATH."
+  fi
+}
+
+if [ -d $SHOPIFY_DEV_PATH ]; then
+  copy_generated_docs_to_shopify_dev
+else
+  # We could be in the monorepo and need to go up several more directories to find shopify-dev
+  SHOPIFY_DEV_PATH="../../../../../../shopify-dev"
```
kumar303, line 103: If you change the line above to `SHOPIFY_DEV_PATH=~/src/github.com/Shopify/shopify-dev` then it will work in both spin and local.

packages/ui-extensions/package.json
kumar303: These were production dependencies before https://github.com/Shopify/ui-extensions/blob/d4020313fd35ad838e2fd06b433487b67e4f12ad/packages/ui-extensions/package.json#L66-L69

Was that because we imported actual code from them?

packages/ui-extensions/src/surfaces/checkout/api/standard/README.md
kumar303, line 1: Hmm, maybe we should delete this file since everything in here is outdated.

kumar303 (approved):
🎩 this is working great in an extension. I think the global definition needs a tweak but otherwise everything looks good.

The generated `shopify.d.ts` has some unexpected imports but they work just fine. We can follow up on that later.

oliverigor (approved):
All changes here makes sense to me :D

packages/ui-extensions/docs/surfaces/checkout/build-docs.sh
jamesvidler: I'll need to double-check this, because I don't think that is necessarily true when its running in a Github action. Will look at this outside of this PR.

packages/ui-extensions/package.json
jamesvidler: That's right. There were some functions such as `createRemoteRoot` that are not used anymore. Actually, this PR doesn't actually use `"@remote-ui/core"` at all so I will remove that. We have `"@remote-ui/async-subscription"` as a devDependancy because we use the `StatefulRemoteSubscribable` type.

packages/ui-extensions/src/surfaces/checkout/globals.ts
jamesvidler: You are right, it is missing the base object that should include `extend()` and `reload()`. I wonder if this was a conscious decision that admin made? There's no good reason for a UI extension to actually use those APIs directly, so maybe hiding it from the types (and consequently the docs) is a good thing?

jamesvidler: Confirmed in [slack thread](https://shopify.slack.com/archives/C07SL56R6RZ/p1744656023549559). These shouldn't be included.

PR #2772, opened 2025-04-15
fix vanillaJS version of global types for customer accounts

brianshen1990:
### Background

preact version works perfect, but we have some type issue for vanillaJS version. Now we declare correct types for the components.

Screenshot 2025-04-15 at 3 49 21 PM

The vanillaJS can get all the props now. 

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/components/PolicyModal.d.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/components/PolicyModal.d.ts
@@ -25,13 +25,9 @@ export interface PolicyModalProps {
   type: PolicyType;
 }
 
-declare class PolicyModalComponent
-  extends HTMLElement
-  implements PolicyModalProps {}
-
 declare global {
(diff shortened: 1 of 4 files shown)


packages/ui-extensions/src/surfaces/customer-account/components/PolicyModal.d.ts
```diff
@@ -25,13 +25,9 @@ export interface PolicyModalProps {
...
 }
 
-declare class PolicyModalComponent
-  extends HTMLElement
-  implements PolicyModalProps {}
-
 declare global {
   interface HTMLElementTagNameMap {
-    ['s-policy-modal']: PolicyModalComponent;
+    ['s-policy-modal']: HTMLElement & PolicyModalProps;
```
oluwatimio, line 30: We dont have this component just yet but i guess will come in future

oluwatimio (approved):

PR #2764, opened 2025-04-11
Add a checkout/preact package for hooks

kumar303:
⚠️ Some links are internal to Shopify.

### Background

Fixes https://github.com/Shopify/temp-project-mover-Archetypically-20250512095102/issues/186 by adding a new hooks library.

### Solution

* copied all hooks from `@shopify/ui-extensions-react/checkout` @ [2025-04](https://shopify.dev/docs/api/checkout-ui-extensions/2025-04) (this [commit](https://github.com/Shopify/ui-extensions/commit/d4020313fd35ad838e2fd06b433487b67e4f12ad))
* I **did not** remove any hooks ([learn why](https://docs.google.com/document/d/1Asx-neu8JgEnCeB_Z7lbk61PLydA56YxHrwpzP6EzzQ/edit?tab=t.0#heading=h.truxg8qtku6))
  * ...with one exception: I removed `useOrder` because that had already been removed in `2025-07` ([details](https://github.com/shop/issues-checkout/issues/1658))
* exposed them from a new entrypoint: `@shopify/ui-extensions/checkout/preact`
* fixed type imports for checkout APIs
* replaced `react` imports with `preact/hooks`
* replaced React API context with global `shopify` in `useApi()`
* removed unnecessary React things in `useTranslate()`
* **deferred** getting the test suite working (it doesn't run in CI for some reason 😅). Follow up: https://github.com/Shopify/temp-project-mover-Archetypically-20250512095102/issues/196

### 🎩

I'm checking compilation like

```
yarn
pushd packages/ui-extensions
../../node_modules/typescript/bin/tsc --build --watch
```

Here's a diff I ran between the old and new hooks libraries.

```bash
diff -r -w \
  ~/tmp/ui-extensions/packages/ui-extensions-react/src/surfaces/checkout/hooks \
  ~/src/github.com/Shopify/ui-extensions/packages/ui-extensions/src/surfaces/checkout/preact
```

<details>
<summary>👁️ diff output</summary>

```diff
Only in /Users/kumar/src/github.com/Shopify/ui-extensions/packages/ui-extensions/src/surfaces/checkout/preact: .DS_Store
diff -r -w /Users/kumar/tmp/ui-extensions/packages/ui-extensions-react/src/surfaces/checkout/hooks/api.ts /Users/kumar/src/github.com/Shopify/ui-extensions/packages/ui-extensions/src/surfaces/checkout/preact/api.ts
1c1
< import {useContext} from 'react';
---
[...]

PR #2777, opened 2025-04-16
add customer account preact hooks support for rc version

brianshen1990:
### Background

There are 2 commits
1.  cop paste all customer account legacy hooks from unstable version so that we can make sure we will not miss any existing hooks. 
2. update the implementation and type imports a little bit to make sure it works 

### Solution

same as checkout https://github.com/Shopify/ui-extensions/pull/2764

### 🎩

Use https://github.com/Shopify/ui-extensions/pull/2777#issuecomment-2810270571

And this is how it looks like
Screenshot 2025-04-16 at 1 55 39 PM

Screenshot 2025-04-16 at 1 55 50 PM

```
import { render } from "preact";
import {
  useExtension,
} from "@shopify/ui-extensions/customer-account/preact";

export default function () {
  render(<Extension />, document.body);
}

function Extension() {
  const extension = useExtension();
  return (
    <s-page>
      <s-button slot="primaryAction" onClick={() => console.log("Primary action")} >PRIMARY BUTTONE</s-button>
      {JSON.stringify(extension)}
    </s-page>
  )
};

```

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/preact/app-metafields.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/preact/app-metafields.ts
@@ -1,0 +1,65 @@
+import type {
+  AppMetafieldEntryTarget,
+  Metafield,
+  AppMetafieldEntry,
+  RenderOrderStatusExtensionTarget,
+} from '@shopify/ui-extensions/customer-account';
+import {useMemo} from 'react';
+
+import {
+  CustomerAccountUIExtensionError,
+  ExtensionHasNoFieldError,
+} from '../errors';
+
+import {useApi} from './api';
+import {useSubscription} from './subscription';
+
+interface AppMetafieldFilters {
+  id?: AppMetafieldEntryTarget['id'];
+  type?: AppMetafieldEntryTarget['type'];
+  namespace?: Metafield['namespace'];
+  key?: Metafield['key'];
+}
+
+type AppMetafieldFilterKeys = keyof AppMetafieldFilters;
+
+/**
+ * Returns the metafields configured with `shopify.ui.extension.toml`.
+ * @arg {AppMetafieldFilters} - filter the list of returned metafields
+ */
+export function useAppMetafields<
+  Target extends RenderOrderStatusExtensionTarget = RenderOrderStatusExtensionTarget,
+>(filters: AppMetafieldFilters = {}): AppMetafieldEntry[] {
+  const api = useApi<Target>();
+  const extensionTarget = api.extension.target;
+
+  if (!('appMetafields' in api)) {
+    throw new ExtensionHasNoFieldError('appMetafields', extensionTarget);
+  }
+
+  const appMetafields = useSubscription(api.appMetafields);
+
+  return useMemo(() => {
+    if (filters.key && !filters.namespace) {
+      throw new CustomerAccountUIExtensionError(
+        'You must pass in a namespace with a key',
+      );
+    }
+
+    const filterKeys = Object.keys(filters) as AppMetafieldFilterKeys[];
+
+    if (filterKeys.length) {
+      return appMetafields.filter((app) => {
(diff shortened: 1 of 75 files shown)


packages/ui-extensions/src/surfaces/customer-account/preact/country.ts
```diff
@@ -0,0 +1,24 @@
...
+import {useSubscription} from './subscription';
+import {ExtensionHasNoFieldError} from './errors';
+
+/**
+ * Returns the country associated with either the current order on the order status page
+ * or the selected country in the customer account interface,
+ * and automatically re-renders your component if the country changes.
+ */
+export function useLocalizationCountry<
+  Target extends RenderExtensionTarget = RenderExtensionTarget,
```
ottocho, line 14: 👍

packages/ui-extensions/src/surfaces/customer-account/preact/tests/authenticated-account.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+  };
+}
+
+function createMockPurchasingCompany() {
+  return {
+    company: {
+      id: `gid://shopify/Company/${faker.number.int({min: 1})}`,
+    },
+    location: {
+      id: `gid://shopify/CompanyLocation/${faker.number.int({min: 1})}`,
```
ottocho, line 22: 👍

ottocho (approved):
Confirmed that our changes for markets project were migrated here from unstable

packages/ui-extensions/src/surfaces/customer-account/preact/app-metafields.ts
```diff
@@ -0,0 +1,65 @@
...
+  return useMemo(() => {
+    if (filters.key && !filters.namespace) {
+      throw new CustomerAccountUIExtensionError(
+        'You must pass in a namespace with a key',
+      );
+    }
+
+    const filterKeys = Object.keys(filters) as AppMetafieldFilterKeys[];
+
+    if (filterKeys.length) {
```
oluwatimio, line 51: im guessing this is the same as the logic from before?

packages/ui-extensions/src/surfaces/customer-account/preact/subscription.ts
```diff
@@ -0,0 +1,45 @@
...
+ * and subscribes to update the value when changes occur in the checkout.
+ *
+ * > Note:
+ * > You generally shouldn’t need to use this directly, as there are dedicated hooks
+ * > for accessing the current value of each individual resource in the checkout.
+ */
+export function useSubscription<Value>(
+  subscription: StatefulRemoteSubscribable<Value>,
+): Value {
+  const [, setValue] = useState(subscription.current);
```
oluwatimio, line 18: Should we use subscription.value here instead? since we want to have a signal like interface cc @vividviolet

vividviolet: I suggested that in the other [PR](https://github.com/Shopify/ui-extensions/pull/2764#discussion_r2047717937) but it looks like these hooks for supposed to be feature-parity with the react ones.

oluwatimio (approved):
Thanks for doing this @brianshen1990 ! I took a look at the api's they are the same as before except using the shopify global now ❤️ . I also had a question about using .value instead of .current

packages/ui-extensions/src/surfaces/customer-account/preact/app-metafields.ts
brianshen1990: copy pasted, no change, you can review the 2nd commit

brianshen1990: Screenshot 2025-04-17 at 11 37 14 AM

PR #2770, opened 2025-04-15
Fix event types

vividviolet:
### Background
Fix https://github.com/Shopify/app-ui/issues/2211

### 🎩

1. Follow this [doc](https://docs.google.com/document/d/1kdqif_HgyH_rZi6prY8p1YbvdrtJZ6lZH-zGGhGWu-o/edit?tab=t.0#heading=h.1f3xod8i38e6) to create a block extension
2. Open the package.json of the extension and update to use the `0.0.0-snapshot-20250421182514` snapshot with the fixed types. Note that the types for the snapshot was generated with this [PR](https://github.com/Shopify/admin-ui-foundations/pull/2613) for building individual d.ts files for each component.
3. Run `pnpm install` to install the snapshot
4. Update your extension code to be as follows:
```tsx
import {render} from 'preact';
export default function extension() {
  render(<Extension />, document.body);
}

function Extension() {
  return (
    <s-admin-block title="My Block Extension">
      <s-form onSubmit={(event) => {
        event.waitUntil((async () => {
          console.log('saving');
          close();
        })());
      }}>
      <s-button onClick={(event) => {
        console.log('clicked', event);
      }}>Test button</s-button>
      </s-form>
    </s-admin-block>
  );
}
```
5. Verify that you get the event typed as either `Event` for the Button onClick handler
image
7. Verify that you get the event typed as `ExtendableEvent` for the Form onSubmit handler
image

### Checklist

- [x] I have :tophat:'d these changes
- [x] I have updated relevant documentation

PR #2801, opened 2025-04-23
[Ui components]: Component dts files

oliverigor:
# I'll port the changes from [here](https://github.com/shop/world/pull/37803) once we're ready and bring only dts files over here


packages/ui-extensions/src/surfaces/checkout/components/Abbreviation/Abbreviation.d.ts
```diff
@@ -0,0 +1,26 @@
...
+/* eslint-disable import/no-deprecated */
+// eslint-disable-next-line @typescript-eslint/triple-slash-reference, spaced-comment
+/// <reference lib="DOM" />
+import type {AbbreviationProps$1} from './components-shared.d.ts';
+
+export interface AbbreviationProps extends Pick<AbbreviationProps$1, 'title'> {
+}
+export interface AbbreviationElement extends AbbreviationProps, Omit<HTMLElement, 'id' | 'title'> {
+}
+declare global {
```
vividviolet: Just noticed that you're missing the preact jsx types. Also, we discovered that preact overwrites conflicting handler props that you have to omit (see related [PR](https://github.com/Shopify/ui-extensions/pull/2770) for Admin). Doing something like this should work
```ts
declare module 'preact' {
  namespace createElement.JSX {
    interface IntrinsicElements {
      's-abbreviation': Omit<
        HTMLAttributes<HTMLElement>,
        Extract<keyof HTMLAttributes<HTMLElement>, `on${Capitalize<string>}`>
      > &
      AbbreviationProps;
    }
  }
}
```

oliverigor: Oh... noice, will add this in the [checkout-web](https://github.com/shop/world/pull/37803) side then, thanks trish :D

jamesvidler: I tophatted this, but the component types still aren't working and that is because of what Trish mentioned above ☝️ relating to the preact jsx types.

Screenshot 2025-04-24 at 10 21 44 AM

oliverigor: ok, I'll work on he checkout-web side to make the additions.

How about the JS ones ?? does that work fine ? or same ?

PR #2809, opened 2025-04-28
Update component definitions from Box-esque element type changes

gwyneplaine:
Run `build:definitions` from https://github.com/Shopify/admin-ui-foundations/pull/2740


packages/ui-extensions/src/surfaces/admin/components/AdminBlock.d.ts
```diff
@@ -39,6 +39,7 @@ export interface ClickOptions {
  */
 declare const BaseClass: typeof globalThis.HTMLElement;
 declare abstract class PreactCustomElement extends BaseClass {
+  #private;
```
gwyneplaine, line 42: Looks like this is added to every individual [componentName].d.ts file
Is this intended? @billfienberg

gwyneplaine:
The changes are all relatively predictable / expected 
The only one I'm unclear on is the addition of a bunch of `#private` tags on all component `d.t.s` files.

packages/ui-extensions/src/surfaces/admin/components/AdminBlock.d.ts
billfienberg: 🤔 I'm not sure if that's intended. I added a `no-ts-private-protected` linting rule in https://github.com/Shopify/admin-ui-foundations/pull/2691, but [opted not to auto-fix](https://github.com/Shopify/admin-ui-foundations/pull/2691/#discussion_r2054755219) any `protected`/`private` linting errors. 

I'm looking through @vividviolet's **Build individual component d.ts** [PR](https://github.com/Shopify/admin-ui-foundations/pull/2613), but I'm not sure why/where the rogue `#private` is coming from in the individual `[componentName].d.ts` files.

vividviolet: This was always there and is already part of admin-ui-components. I had custom code to remove it from ui-extensions but I [deleted it](https://github.com/Shopify/admin-ui-foundations/pull/2613/files#diff-7f58b3e2afb1a54c6f0ee1d31c2b3cc4bf52f8a7120295e28d98b341bf72c886L22) thinking we don't need it. Can you add it back to `cleanContent` and run the build script again?

PR #2812, opened 2025-04-28
Fix example

Fionoble:
This pull request includes a small change to the `form-default-value.jsx` file. The change adds an import for the `render` function from the `preact` library, which was previously missing.


packages/ui-extensions/docs/surfaces/admin/staticPages/examples/form-default-value.jsx
vividviolet, line 1: While you're in here can you fix this [issue](https://docs.google.com/document/d/1W5OVIv1KKCOmwsfMNhoMpgQ8W6K-dZz5zbDUw7Y3pMU/edit?tab=t.0#bookmark=kix.ha8cx3hfdttj)?

vividviolet: Also, we should add `name` for the fields for now otherwise the CSB won't work.

Fionoble: Should we fix that in web? Add logic that if the input doesn't have a name we assign a temp name?

vividviolet: Yeah, we definitely should fix it but since it's broken now we should add it. Having a name is best practice anyway. Eventually it would be nice to add support for `FormData` so having a name would allow that to work.

vividviolet (approved):

alex-page (approved):

billfienberg (approved):

packages/ui-extensions/docs/surfaces/admin/staticPages/examples/form-default-value.jsx
Fionoble: Ah ok, that makes sense

Fionoble: Added a blurb about name being required to trigger the forms as well

PR #2831, opened 2025-05-02
Omit slots from JSX props for Admin components

vividviolet:
### Background

Slot props were showing back up in some Admin component we need to remove them to avoid type errors like this:
 
image

I also ran the generator again so we picked up some new types from the component source.

### 🎩
1. Consume the library snapshot
2. Create an Admin Action remote dom extension
3. Verify that you no longer see type errors for missing slot props

### Checklist


packages/ui-extensions/src/surfaces/admin/components/AdminAction.d.ts
```diff
@@ -102,11 +102,14 @@ declare module 'preact' {
     interface IntrinsicElements {
       [tagName]: Omit<
         HTMLAttributes<HTMLElement>,
-        Extract<keyof HTMLAttributes<HTMLElement>, `on${Capitalize<string>}`>
+        | Extract<keyof HTMLAttributes<HTMLElement>, `on${Capitalize<string>}`>
+        | 'primaryAction'
+        | 'secondaryActions'
       > &
-        ReactProps;
+        AdminActionJSXProps;
```
sam-b-rose, line 109: It looks like even though we `Omit` the props above, they are being added back here when using the intersection  operator.

vividviolet:
no longer relevant

PR #2837, opened 2025-05-05
Add storage API to point-of-sale extension targets base input

Alex-Palad:
### Background

Part 1/4 of https://github.com/Shopify/temp-project-mover-Archetypically-20250612122555/issues/187

Closes https://github.com/Shopify/temp-project-mover-Archetypically-20250612122555/issues/276

### Solution

Adds storage API types to allow extensions store data in POS App

Implemented API from https://github.com/Shopify/ui-api-design/pull/909

### 🎩

- ...

### Checklist

- [x] I have 🎩'd these changes

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
@@ -14,6 +14,7 @@ import type {Components} from './shared';
 import type {AnyComponentBuilder} from '../../shared';
 import type {ReprintReceiptData} from './event/data/ReprintReceiptData';
 import type {TransactionCompleteData} from './event/data/TransactionCompleteData';
+import {StorageApi} from './render/api/storage-api/storage-api';
 
 type SmartGridComponents = AnyComponentBuilder<Pick<Components, 'Tile'>>;
 type ActionComponents = AnyComponentBuilder<Pick<Components, 'Button'>>;
@@ -161,10 +162,8 @@ export interface ExtensionTargets {
   'pos.receipt-footer.block.render': RenderExtension<
     // NOTE: key/any type is cause of no arg useApi() that includes all target types.
     //   stop using useApi() with no args, instead specify the target type explicitly.
-    {[key: string]: any} & (
-      | TransactionCompleteData
-      | {transaction: ReprintReceiptData}
-    ),
+    {[key: string]: any} & StorageApi &
+      (TransactionCompleteData | {transaction: ReprintReceiptData}),
     ReceiptComponents
   >;
 }
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
@@ -1,0 +1,65 @@
+export interface StorageExceededError extends Error {
+  name: 'StorageExceededError';
+  code: 'RecordsCount' | 'RecordSize';
+}
+export interface Storage<
+  BaseStorageTypes extends Record<string, any> = Record<string, unknown>,
+> {
+  /**
+   * Sets the value of a key in the storage.
+   *
+   * @param key - The key to set the value for.
+   * @param value - The value to set for the key.
+   * Can be any primitive type supported by `JSON.stringify`.
+   * @throws StorageExceededError if the extension exceeds its allotted storage limit.
+   * @throws StorageExceededError if value exceeds its allotted storage limit.
+   */
+  set<
+    StorageTypes extends BaseStorageTypes = BaseStorageTypes,
+    Keys extends keyof StorageTypes = keyof StorageTypes,
+  >(
+    key: Keys,
+    value: StorageTypes[Keys],
+  ): Promise<void>;
+
+  /**
+   * Gets the value of a key in the storage.
+   *
+   * @param key - The key to get the value for.
+   * @returns The value of the key.
+   * If no value for the key exists, the resolved value is undefined.
+   */
+  get<
+    StorageTypes extends BaseStorageTypes = BaseStorageTypes,
+    Keys extends keyof StorageTypes = keyof StorageTypes,
+  >(
+    key: Keys,
+  ): Promise<StorageTypes[Keys] | undefined>;
+
+  /**
+   * Clears the storage.
+   */
+  clear: () => Promise<void>;
+
+  /**
+   * Deletes a key from the storage.
+   *
+   * @param key - The key to delete.
+   */
+  delete<
+    StorageTypes extends BaseStorageTypes = BaseStorageTypes,
+    Keys extends keyof StorageTypes = keyof StorageTypes,
+  >(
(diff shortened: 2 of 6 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
```diff
@@ -0,0 +1,42 @@
+export interface Storage {
+  /**
+   * Sets the value of a key in the storage.
+   *
+   * @param key - The key to set the value for.
+   * @param value - The value to set for the key.
+   * @throws Error if storage API is not available or the method is not implemented.
+   * Value can be any primitive type supported by `JSON.stringify`.
+   */
+  set: <StoredValue = any>(key: string, value: StoredValue) => Promise<void>;
```
js-goupil: These methods look good, but it looks like we are expected to have more implemented according to [this doc](https://github.com/Shopify/ui-api-design/pull/909/files)

js-goupil (commented):
I think there's value in setMany and getMany, considering these calls are async, we want to make sure the 3P devs have access to more performant way to doing things in bulk

packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
Alex-Palad: As we discussed before, we'd like to implement only minimal set of methods. We allow partners use methods defined in ui-extensions/point-of-sale, so we don't have to implement all of those methods.
However, this must be compatible with the proposed API.

packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
fatbattk, line 1: should this follow the types, jsdoc, etc. from [the api design?](https://github.com/Shopify/ui-api-design/pull/909/files#diff-0a7d57b868fd2bddebefba552d761a627f09fa36aa7ced08518b011f9e7edf1a)

fatbattk (commented):
do we need to add the `StandardApi` for event/observe targets too?

also missed this one render, https://github.com/Shopify/ui-extensions/blob/unstable/packages/ui-extensions/src/surfaces/point-of-sale/targets.ts#L161

Alex-Palad:
> do we need to add the `StandardApi` for event/observe targets too?

We have BaseData for event/observe

> also missed this one render, https://github.com/Shopify/ui-extensions/blob/unstable/packages/ui-extensions/src/surfaces/point-of-sale/targets.ts#L161

Good catch. Added it to block.render

packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
```diff
@@ -164,6 +165,7 @@ export interface ExtensionTargets {
     {[key: string]: any} & (
       | TransactionCompleteData
       | {transaction: ReprintReceiptData}
+      | StorageApi
```
fatbattk: should this just use `StandardApi` now?

Alex-Palad: No, we don't send StandardApi to block.render, because it is not an actual render target. We can't use useApi hook there

fatbattk: oh yeah `pos.receipt-footer.block.render` is only retrieved.

packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
vctrchu: ^

packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
```diff
@@ -0,0 +1,61 @@
+export interface Storage<
+  BaseStorageTypes extends Record<string, any> = Record<string, unknown>,
+> {
+  /**
+   * Sets the value of a key in the storage.
+   *
+   * @param key - The key to set the value for.
+   * @param value - The value to set for the key.
+   * Can be any primitive type supported by `JSON.stringify`.
+   * @throws StorageExceededError if the extension exceeds its allotted storage limit.
```
vctrchu, line 14: I think we need to create this interface?

vctrchu: https://github.com/Shopify/ui-api-design/blob/a43c4d69314dc299ac3cd67578aa78b9770dfe6f/ui-api-design/extensions/apis/Storage/index.ts#L128C1-L130C2

packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
vctrchu: Sorry, can you expand more on why a minimal set of methods? Is it to reduce work on the POS? I might be missing context on the use cases.

vctrchu: For context in the screen shot, "First interface" is your PR.
Screenshot 2025-05-13 at 5 11 07 PM

The second interface is the API design, I feel like `update` could be handy?

js-goupil: @Alex-Palad my worry is that partners might loop through data and call `set`, or `get` in a loop which is highly inefficient when going across the RPC layer. For context we've seen them do this when trying to apply cart properties on 50 line items and we ended up with cart updates that were taking 20 seconds on Android devices. What's stopping us from implementing `setMany` and `getMany`?

packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
Alex-Palad: Good point, I added an error interface

packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
Alex-Palad: We follow it. We don't implement all methods from that interface (we deploy only minimal set of methods, and will add more if needed in future)

packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
Alex-Palad: >Sorry, can you expand more on why a minimal set of methods? Is it to reduce work on the POS? I might be missing context on the use cases.

Is too reduce risks.
`update` method needs transactions, whaat if they never commit transaction? 
`setMany` -validation risks. What if one value is valid, and 2nd is to large. Shall we reject both or one?

It is easy to add new methods than to remove existing in future.


>>my worry is that partners might loop through data and call set, or get in a loop

This is a key-value storage. I don't see any reason to do it in a loop. For offline transactions - they will reuse a single record, and will push a new value there. (There shouldn't be race conditions. I don't think we will complete next transactions before they can save previous)

fatbattk (approved):

js-goupil (approved):

PR #2838, opened 2025-05-05
DatePicker (and other small updates)

frehner:
From following the steps here https://github.com/Shopify/admin-ui-foundations/blob/main/documentation/extensions-architecture.md#how-to-make-a-component-available-in-remote-dom I'm adding DatePicker

But also, apparently a bunch of other small changes that hadn't gone to this repo either?


packages/ui-extensions/src/surfaces/admin/components/DatePicker.d.ts
vividviolet, line 1: I think you're missing the docs for DatePicker. You can run `gen-docs:admin DatePicker` to get a template started

vividviolet (commented):
We also need to make the DatePicker available as a component. If it's a standard component we should add it [here](https://github.com/Shopify/ui-extensions/blob/2025-10-rc/packages/ui-extensions/src/surfaces/admin/components/StandardComponents.ts).

packages/ui-extensions/src/surfaces/admin/components/DatePicker.d.ts
```diff
@@ -0,0 +1,182 @@
...
+      > &
+        DatePickerJSXProps;
+    }
+  }
+}
+
+declare const tagName = 's-date-picker';
+export interface DatePickerJSXProps
+  extends Partial<DatePickerProps>,
+    Pick<DatePickerProps$1, 'id'> {
```
vividviolet, line 169: This should be changed to `Partial<Pick<DatePickerProps$1, 'id'>>`? or Maybe this?
```ts
interface DatePickerJSXProps
  extends Partial<DatePickerProps & Pick<DatePickerProps$1, 'id'>>, 
```

vividviolet (commented):
Looking good aside from the type error for `id`
image

packages/ui-extensions/src/surfaces/admin/components/DatePicker.d.ts
frehner: This seems the same as every other component, not sure why it's not happy here.

Maybe you need to restart your TS Server for it to pickup the changes? 🤷

vividviolet: Yup, it was a different issue due to me manually copying the built files since I couldn't install the snapshot

vividviolet (approved):

PR #2853, opened 2025-05-08
support checkout component types for customer account

brianshen1990:
### Background

Customer account ui extensions can also use component from checkout. By default, the build only get all the components from that specific surface's source path. 

In this PR, I make the build scripts more flexible. For components, now we can specify more than 1 source paths. 

### 🎩

- Create an new customer account ui extension follow [this doc](https://docs.google.com/document/d/1hrK6JNHBi8YbEGMoaQiFQqDC8qN9dvzWVTWmHCugMDU/edit?tab=t.0#heading=h.lr188tolgi4o  )
- update extension package.json's ui-extension version to `0.0.0-snapshot-20250508171603`  ([snapshot version from this PR](https://github.com/Shopify/ui-extensions/pull/2853#issuecomment-2863758168))
- now , enjoy the checkout component types!

Screenshot 2025-05-08 at 1 43 03 PM

### Checklist

### DIFF
--- a/packages/ui-extensions/buildTargetDts.ts
+++ b/packages/ui-extensions/buildTargetDts.ts
@@ -14,15 +14,17 @@ import type {
 import {Project} from 'ts-morph';
 
 function copyComponentDefinitions({
-  srcPath,
+  srcPaths,
   buildPath,
   surface,
 }: {
-  srcPath: string;
+  srcPaths: string[];
   buildPath: string;
   surface: string;
 }) {
-  const componentsSrcPath = join(srcPath, 'components');
+  const componentsSrcPaths = srcPaths.map((srcPath) =>
+    join(srcPath, 'components'),
+  );
   const componentsBuildPath = join(
     buildPath,
     `ts/surfaces/${surface}/components`,
@@ -31,23 +33,28 @@ function copyComponentDefinitions({
     mkdirSync(buildPath, {recursive: true});
   }
 
-  if (!existsSync(componentsSrcPath)) {
+  const exists = componentsSrcPaths.every((path) => existsSync(path));
+
+  if (!exists) {
     // eslint-disable-next-line no-console
     console.log('No components to copy');
     return false;
   }
 
-  const components = readdirSync(componentsSrcPath);
-  components
-    .filter((file) => {
-      return file.endsWith('d.ts');
-    })
-    .forEach((file) => {
-      copyFileSync(
-        join(componentsSrcPath, file),
-        join(componentsBuildPath, file),
-      );
-    });
+  componentsSrcPaths.forEach((componentsSrcPath) => {
+    const components = readdirSync(componentsSrcPath);
+    components
+      .filter((file) => {
+        return file.endsWith('d.ts');
+      })
+      .forEach((file) => {
+        copyFileSync(
+          join(componentsSrcPath, file),
+          join(componentsBuildPath, file),
+        );
+      });
+  });
+
   return true;
 }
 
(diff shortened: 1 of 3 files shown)


packages/ui-extensions/buildTargetDts.ts
```diff
@@ -14,15 +14,17 @@ import type {
 import {Project} from 'ts-morph';
 
 function copyComponentDefinitions({
-  srcPath,
+  srcPaths,
```
brianshen1990, line 17: please use hide whitespace to review.  

There is no different if there is only 1 `srcPath` in `srcPaths`. The only difference happens when there are more. for example, for customer account target, we will use components from both checkout and customer account surface.

packages/ui-extensions/buildTargetDts.ts
```diff
@@ -234,7 +244,17 @@ export function buildTargetsDefinitions(directory: string, surface: string) {
   const project = new Project();
   const buildPath = resolve(directory, 'build');
   const srcPath = resolve(directory, `src/surfaces/${surface}`);
-  const success = copyComponentDefinitions({srcPath, buildPath, surface});
+
+  const checkoutSrcPath = resolve(directory, `src/surfaces/checkout`);
+
+  const componentSrcPaths =
+    surface === 'customer-account' ? [srcPath, checkoutSrcPath] : [srcPath];
```
brianshen1990: this is the key updates:  for customer account target, we will use components from both checkout and customer account surface.

packages/ui-extensions/src/surfaces/customer-account/components/StandardComponents.ts
```diff
@@ -1,4 +1,7 @@
+import {AnyComponent} from '../../checkout';
+
 export type StandardComponents =
+  | AnyComponent
```
brianshen1990, line 4: we will use all checkout components. 

An notice is that, checkout has mode components types defined in AnyComponent than real component ts files ( many of them are still in porgree ), we will see some warning logs indicating some components are not found when run `yarn build` , for example
```
Component ChoiceList not found in undefined
Component DatePicker not found in undefined
```

andrewmcgov (commented):
Is it expected that not all type definitions are available yet?

For example its working for button and banner, but not modal, email-field, number-field, and a few others. 

Screenshot 2025-05-08 at 2 22 44 PM

packages/ui-extensions/buildTargetDts.ts
```diff
@@ -234,7 +244,17 @@ export function buildTargetsDefinitions(directory: string, surface: string) {
   const project = new Project();
   const buildPath = resolve(directory, 'build');
   const srcPath = resolve(directory, `src/surfaces/${surface}`);
-  const success = copyComponentDefinitions({srcPath, buildPath, surface});
+
+  const checkoutSrcPath = resolve(directory, `src/surfaces/checkout`);
+
+  const componentSrcPaths =
```
vividviolet: I would prefer that we pass things in from loom instead of hardcoding custom stuff here.
```suggestion
  const componentSrcPaths = [
    srcPath,
    ...additionalComponentPaths.map((path) => resolve(directory, path)),
  ];
```

And then in the loom config you can do
```ts
buildTargetsDefinitions(
  resolve(process.cwd(), 'packages/ui-extensions'),
  'customer-account',
  ['src/surfaces/checkout'],
);

**Review by vividviolet — COMMENTED** [review]
Just one suggestion - otherwise I'm good with the change

File: packages/ui-extensions/buildTargetDts.ts (hunk 4 above)
Comment (brianshen1990) [reply packages/ui-extensions/buildTargetDts.ts]: ah! thanks for the suggestion, will make a update 🙇

**Comment by brianshen1990** [comment]
> Is it expected that not all type definitions are available yet?
> 
> For example its working for button and banner, but not modal, email-field, number-field, and a few others.
> 
> Screenshot 2025-05-08 at 2 22 44 PM

Hey Andrew, these types come from here https://github.com/Shopify/ui-extensions/tree/1a6bc05b0e125e2fb662add5fe341b53e6747546/packages/ui-extensions/src/surfaces/checkout/components , as you can see, Modal, EmailField, NumberField are not there yet. Once checkout add it, we will have these new type~

File: packages/ui-extensions/buildTargetDts.ts (hunk 4 above)
Comment (brianshen1990) [reply packages/ui-extensions/buildTargetDts.ts]: updated!

**Review by vividviolet — APPROVED** [review]
Tophat'd 👍 
image

File: packages/ui-extensions/loom.config.ts [hunk 5]
```diff
@@ -61,6 +61,7 @@ export default createPackage((pkg) => {
             buildTargetsDefinitions(
               resolve(process.cwd(), 'packages/ui-extensions'),
               'customer-account',
+              [`src/surfaces/checkout`],
```
robin-drexler: nit: why is this a template string? 😅

robin-drexler (approved):

packages/ui-extensions/loom.config.ts
brianshen1990: copy pasted, updating now
Screenshot 2025-05-08 at 3 33 05 PM

PR #2854, opened 2025-05-08
Update Admin UI component definitions

sam-b-rose:
Updates the UI Extension component definitions with restructured responsive props that include comments.

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components/AdminAction/examples/default.tsx
+++ b/packages/ui-extensions/src/surfaces/admin/components/AdminAction/examples/default.tsx
@@ -1,13 +1,13 @@
 <s-admin-action title="My App Action">
   Modal content
   <s-button
-    slot="primaryAction"
+    slot="primary-action"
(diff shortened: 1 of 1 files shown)


packages/ui-extensions/src/surfaces/admin/components/Image.d.ts
```diff
@@ -50,37 +50,37 @@ export type BoxBorderStyles = Extract<
...
+export type ResponsiveBoxProps = MakeResponsivePick<
+  RequiredBoxProps,
+  | 'padding'
+  | 'paddingBlock'
+  | 'paddingBlockStart'
+  | 'paddingBlockEnd'
+  | 'paddingInline'
+  | 'paddingInlineStart'
+  | 'paddingInlineEnd'
+>;
```
kyledurand: This seems like a possible side effect. Can you tophat and see if there's some weird padding coming out in image docs? Might be worth digging into because image should only be accessing border props off box

PR #2855, opened 2025-05-08
add events sections

kyledurand:
Part of https://github.com/Shopify/admin-ui-foundations/issues/2645
image

This removes events from properties (not sure if this causes issues with extensions @vividviolet please let me know)


vividviolet (approved):
Looks good to me. The components for the client lib are using the separate [definitions](https://github.com/Shopify/ui-extensions/blob/e7c615f64f35b233e3b954f094e84a8222310757/packages/ui-extensions/src/surfaces/admin/components/Clickable.d.ts#L344) so we should be fine to edit the `components.d.ts` for the docs.

PR #2858, opened 2025-05-13
Docs: new setup for shared component content

jffortier:
This PR creates the foundation to share components content with other surfaces. It currently use the content from Checkout, but in a follow up PR, I'll start aligning the components description and involve Admin / Customer Account in the process. I'll also use the sharedContent on the Admin side at that point. I don't want to disturb their content yet.


packages/ui-extensions/src/docs/shared/components/Banner.ts
```diff
@@ -0,0 +1,11 @@
+import {SharedReferenceEntityTemplateSchema} from '../docs-type';
+
+const data: SharedReferenceEntityTemplateSchema = {
+  name: 'Banner',
+  description:
```
oliverigor, line 5: to avoid the `\n\n` we can write this descriptions as markdown so 
```suggestion
  description:`
  Use s-banner... .
  You can use...
 `
```

we have that example on clipboard item

packages/ui-extensions/src/docs/shared/components/Button.ts
```diff
@@ -0,0 +1,11 @@
+import type {SharedReferenceEntityTemplateSchema} from '../docs-type';
+
+const sharedContent: SharedReferenceEntityTemplateSchema = {
```
oliverigor: is there a reason why some is `data` and the rest is sharedContent ?

packages/ui-extensions/src/docs/shared/components/Heading.ts
```diff
@@ -0,0 +1,11 @@
+import type {SharedReferenceEntityTemplateSchema} from '../docs-type';
+
+const data: SharedReferenceEntityTemplateSchema = {
+  name: 'Heading',
+  description:
+    "The `s-heading` component renders hierarchical titles similar to HTML's `h1-h6` elements. Heading levels are automatically determined by parent `s-section` nesting depth starting from h2. Always prefer using the `heading` property on `s-section` and `s-banner` components before constructing custom layout with `s-heading`.",
```
oliverigor:
```suggestion
    "The `s-heading` component renders hierarchical titles similar to HTML’s `h1-h6` elements. Heading levels are automatically determined by parent `s-section` nesting depth starting from h2. Always prefer using the `heading` property on `s-section` and `s-banner` components before constructing custom layout with `s-heading`.",
```

no `’` ?

packages/ui-extensions/src/docs/shared/components/Stack.ts
```diff
@@ -0,0 +1,11 @@
+import type {SharedReferenceEntityTemplateSchema} from '../docs-type';
+
+const data: SharedReferenceEntityTemplateSchema = {
+  name: 'Stack',
+  description:
+    "`s-stack` structures layout elements along the [block or inline axes](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_display/Block_and_inline_layout_in_normal_flow) of the page. It is useful for organizing buttons, creating layouts that adjust to screen size, and controlling spacing between elements.\n\nIt's important to note that Stack does not include any padding by default. If you need padding around your stacked elements, use `base` to apply the default padding. When spacing becomes limited, Stack will always wrap children to a new line.\n\nWhen using `s-stack`, use smaller gaps between small elements and larger gaps between big ones. Maintain a consistent spacing in stacks across all pages of your app.",
```
oliverigor, line 6: I think descriptions like this will benefit a lot on being written as markdown, will read better imo

oliverigor (approved):
left somme comments/suggestions but nothing blocking

Thanks for the help on this brother :D

packages/ui-extensions/src/docs/shared/components/Button.ts
jffortier: The reason is AI. I dont know why it did this.

packages/ui-extensions/src/docs/shared/components/Stack.ts
jffortier: yeah, I'll use markdown when we rewrite them.

packages/ui-extensions/src/docs/shared/components/Banner.ts
jffortier: yeah, I'll use markdown when we rewrite them.

PR #2865, opened 2025-05-14
ignore native html props for customer account elements

brianshen1990:
ignore native html props for customer account elements

no title anymore:
Screenshot 2025-05-14 at 11 13 21 AM

heading is still here:
Screenshot 2025-05-14 at 11 14 03 AM

Vanilla JS version
Screenshot 2025-05-15 at 1 23 45 PM

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/components/Page.d.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/components/Page.d.ts
@@ -10,17 +10,23 @@ export interface PageProps {
   subheading?: string;
 }
 
+export interface PageElement extends HTMLElement, PageProps {}
+
 declare global {
   interface HTMLElementTagNameMap {
-    ['s-page']: HTMLElement & PageProps;
+    ['s-page']: PageElement;
   }
 }
 
 declare module 'preact' {
+  interface BaseProps {
+    children?: preact.ComponentChildren;
+    slot?: Lowercase<string>;
+  }
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace createElement.JSX {
     interface IntrinsicElements {
-      ['s-page']: HTMLAttributes<HTMLElement> & PageProps;
+      ['s-page']: BaseProps & PageProps;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


packages/ui-extensions/src/surfaces/customer-account/components/Page.d.ts
```diff
@@ -12,15 +12,19 @@ export interface PageProps {
 
 declare global {
   interface HTMLElementTagNameMap {
-    ['s-page']: HTMLElement & PageProps;
+    ['s-page']: PageProps;
```
simontaisne: I think we still need to extend `HTMLElement` here; otherwise, non preact extensions will lose a bunch of properties like `addEventListener()`.

simontaisne (commented):
Are you also going to update the interfaces in `world`?

simontaisne:
Also, FYI, here's the equivalent PR for checkout: https://github.com/shop/world/pull/56242.

packages/ui-extensions/src/surfaces/customer-account/components/Page.d.ts
robin-drexler: @simontaisne can you elaborate on this more? Wouldn't the same be true for button or am I missing something? https://github.com/shop/world/pull/56242/files#diff-846cda7003c339964ea0e81d58cb7c1223223ab6bb4c82bf1b0fb0e23cd46b8eR49

brianshen1990:
> Are you also going to update the interfaces in `world`?

customer account is using public ui-extensions, we don't have the private version in world

packages/ui-extensions/src/surfaces/customer-account/components/Page.d.ts
simontaisne: So essentially:
- For vanilla JS, we're extending `HTMLElement` to retain essential functionality like `addEventListener()` or `textContent`, even if some native HTML attributes don't apply. We'll aim to tidy this up later, as it's a lower priority.
- For the Preact version, we're moving away from extending HTMLElement to prevent confusion from TypeScript attribute suggestions.

simontaisne (approved):
LGTM 👍

PR #2870, opened 2025-05-15
Restructure docs for simple descriptions

alex-page:
### Background

We are trying to align descriptions across Polaris surfaces. Descriptions are also used for cards. This creates multiple issues with the current approach:
- Markdown titles get rendered without using the correct heading levels
- Content inside the descriptions cannot be easily shared as it contains context relevant to Admin but not to other surfaces
- Markdown titles show up in the component listing page

### Solution

This PR
- [x] Moves all markdown titles into sub sections and removes them from descriptions
- [x] Ensures descriptions are only ever a small 1-2 sentence bit of content that can be reused across surfaces
- [x] Ensures it uses the structure that is available in shopify-dev

The tradeoff for this is that the documentation lives below the properties. I think this is fine and if we wanted to adjust that we could do it in other ways. We should ensure the description is simple and sharable.

### Follow up

- This allows us to safely remove the logic that was stripping mark down titles from descriptions on listing pages
- It also means that that logic won't break again if a developer used a `##` or `###` as it only stripped `####` titles

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components/Banner/shared.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components/Banner/shared.ts
@@ -1,25 +1,33 @@
 const shared = {
   name: 'Banner',
-  description: `
-  Use \`s-banner\` to communicate important status updates or required actions that users must take. Banners will automatically adjust their design to match the context in which they are used.
-
-#### Outside of a section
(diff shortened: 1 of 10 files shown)


sam-b-rose:
This looks great—nice work! 💯

PR #2874, opened 2025-05-15
publish pre version

robin-drexler:
Publish pre version: https://changesets-docs.vercel.app/en/prereleases

I wonder if there's a way to easily test this with less risk

### DIFF
--- a/.github/workflows/rc-snapshot.yml
+++ b/.github/workflows/rc-snapshot.yml
@@ -41,6 +41,6 @@ jobs:
         run: yarn build
 
       - name: Deploy snapshot version
-        run: yarn changeset version --snapshot ${{ steps.extract_branch.outputs.branch }} && yarn changeset publish --tag ${{ steps.extract_branch.outputs.tag }} --no-git-tag
+        run: yarn changeset version && yarn changeset publish --tag ${{ steps.extract_branch.outputs.tag }} --no-git-tag
         env:
           GITHUB_TOKEN: ${{ secrets.SHOPIFY_GH_ACCESS_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/rc-snapshot.yml
```diff
@@ -41,6 +41,6 @@ jobs:
         run: yarn build
 
       - name: Deploy snapshot version
-        run: yarn changeset version --snapshot ${{ steps.extract_branch.outputs.branch }} && yarn changeset publish --tag ${{ steps.extract_branch.outputs.tag }} --no-git-tag
+        run: yarn changeset version && yarn changeset publish --tag ${{ steps.extract_branch.outputs.tag }} --no-git-tag
```
robin-drexler, line 44: I left the `--tag` here because otherwise the tag would end up being `rc` of which we have multiple at the same time so having a single `rc` tag seems odd.

I'm not 100% certain `changeset publish` wil respect it when in pre mode.

packages/ui-extensions/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/ui-extensions",
-  "version": "2025.10.0",
+  "version": "2025.10.0-rc.0",
```
robin-drexler, line 3: I ran a `changesets version` locally when this was `2025.10.0` and it'd create a `2025.11.0.rc-0` version. 

Changing this here to `2025.10.0-rc.0` before running `changeset pre enter rc` made `changeset version` change it to `2025.10.0-rc.1` which is what we would want.

robin-drexler:
let me close and (maybe) re-open a new one. The approach in here doesn't work

PR #2875, opened 2025-05-15
2025 10 rc pre release

robin-drexler:
https://changesets-docs.vercel.app/en/prereleases

Creates a pre release version when a commit is merged into an -rc branch.


packages/ui-extensions/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/ui-extensions",
-  "version": "2025.10.0",
+  "version": "2025.10.0-rc.0",
```
robin-drexler, line 3: I ran a `changesets version` locally when this was `2025.10.0` and it'd create a `2025.11.0.rc-0` version. 

Changing this here to `2025.10.0-rc.0` before running `changeset pre enter rc` made `changeset version` change it to `2025.10.0-rc.1` which is what we would want.


Also if there's a changeset with a `major` change, this will raise the year to the next version automatically.

.changeset/pre.json
```diff
@@ -0,0 +1,8 @@
+{
+  "mode": "pre",
+  "tag": "rc",
```
robin-drexler, line 3: This will also publish the npm package under the `rc` dist-tag.  If we wanted to have a `2025-10-rc` tag, we could look into adjusting the ` Set 'latest' NPM dist tag` step to also deal with rc versions, but I'd like to see the whole thing work first before investing into this.

.changeset/pre.json
```diff
@@ -0,0 +1,8 @@
...
+  "initialVersions": {
+    "@shopify/ui-extensions": "2025.10.0-rc.0"
+  },
+  "changesets": []
```
vividviolet, line 7: What's this for?

.changeset/pre.json
vividviolet, line 1: Now that we have this we should also delete `.github/workflows/rc-snapshot.yml` otherwise it will be also run.

.github/workflows/deploy.yml
```diff
@@ -11,6 +11,7 @@ on:
       - 2024-07
       - 2024-10
       - 20[0-9][0-9]-[01][1470]
+      - 20[0-9][0-9]-[01][1470]-rc
```
vividviolet, line 14: I don't think we should add to this because it's for stable releases. Shouldn't we have a new workflow that does the prerelease steps? Ideally we wouldn't need a Version PR to ship these prereleases

.changeset/pre.json
robin-drexler: [this was generated by `changeset pre enter next` ](https://changesets-docs.vercel.app/en/prereleases#:~:text=bun-,npm%20run%20changeset%20pre%20enter%20next,-npm%20run%20changeset)

It gets populated when `changeset version` is run. I imagine its to keep track of what needs to be released? I'm not entirely sure tbh.

.github/workflows/deploy.yml
robin-drexler: > Ideally we wouldn't need a Version PR to ship these prereleases

I'm not sure how else this would work? I _think_ only snapshot releases support this. But I'm happy to be proven wrong

.changeset/pre.json
robin-drexler: I thought about this, but thought it'd be useful to have both. A snapshot for everything we merge and then the proper rc version when we actually want to cut a proper public RC release.

.github/workflows/deploy.yml
vividviolet: Ok, I'm good with that. I think the Version PRs can only be approved by certain ppl though

vividviolet (approved):

PR #2876, opened 2025-05-15
Components description edits

jffortier:
Update descriptions for our components. These will be shared by all surfaces so they are generic and could apply to merchant/buyers/customers audiences.

This PR only includes the components we are releasing for Editions in Checkout. The other components for Admin will come in another PR.

Post edition, we'll update ui-api-design with them and hopefully make it the source of truth.

### Checklist

### DIFF
--- a/packages/ui-extensions/src/docs/shared/components/Heading.ts
+++ b/packages/ui-extensions/src/docs/shared/components/Heading.ts
@@ -3,7 +3,7 @@ import type {SharedReferenceEntityTemplateSchema} from '../docs-type';
 const data: SharedReferenceEntityTemplateSchema = {
   name: 'Heading',
   description:
-    'The `s-heading` component renders hierarchical titles similar to HTML’s `h1-h6` elements. Heading levels are automatically determined by parent `s-section` nesting depth starting from `h2`. Always prefer using the `heading` property on `s-section` and `s-banner` components before constructing custom layout with `s-heading`.',
+    'Renders hierarchical titles to communicate the structure and organization of page content. Heading levels adjust automatically based on nesting within parent Section components, ensuring a meaningful and accessible page outline.',
   category: 'Polaris web components',
   subCategory: 'Titles and text',
   related: [
--- a/packages/ui-extensions/src/docs/shared/components/Text.ts
+++ b/packages/ui-extensions/src/docs/shared/components/Text.ts
@@ -3,7 +3,7 @@ import type {SharedReferenceEntityTemplateSchema} from '../docs-type';
 const data: SharedReferenceEntityTemplateSchema = {
   name: 'Text',
   description:
-    'The `s-text` component is used for inline text styling, similar to HTML `span` elements, not for paragraph-level content. Use it to apply different visual tones and text styles to specific words or phrases within a `s-paragraph`, such as a `strong` type or `critical` tone.',
+    'Displays inline text with specific visual styles or tones. Use to emphasize or differentiate words or phrases within a Paragraph or other block-level components.',
   category: 'Polaris web components',
   subCategory: 'Titles and text',
   related: [
--- a/packages/ui-extensions/src/surfaces/admin/components/Box/shared.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components/Box/shared.ts
@@ -1,7 +1,7 @@
 const shared = {
   name: 'Box',
   description:
-    'A generic container that provides a flexible alternative for custom designs not achievable with existing components. Use it to apply styling such as backgrounds, padding, or borders, or to nest and group other components. The contents of Box always maintain their natural size, making it especially useful within layout components that would otherwise stretch their children.',
+    'Use `s-box` to build custom interfaces with your own design language.',
   thumbnail: '/assets/templated-apis-screenshots/admin/components/box.png',
   isVisualComponent: true,
   subSections: [
--- a/packages/ui-extensions/src/surfaces/admin/components/Box/shared.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components/Box/shared.ts
@@ -1,7 +1,7 @@
 const shared = {
   name: 'Box',
   description:
-    'Use `s-box` to build custom interfaces with your own design language.',
+    'A generic container that provides a flexible alternative for custom designs not achievable with existing components. Use it to apply styling such as backgrounds, padding, or borders, or to nest and group other components. The contents of Box always maintain their natural size, making it especially useful within layout components that would otherwise stretch their children.',
   thumbnail: '/assets/templated-apis-screenshots/admin/components/box.png',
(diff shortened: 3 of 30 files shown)


packages/ui-extensions/src/surfaces/admin/components/Box/shared.ts
```diff
@@ -1,7 +1,7 @@
 const shared = {
   name: 'Box',
   description:
-    'Use `s-box` to build custom interfaces with your own design language.',
+    'A generic container that provides a flexible alternative for custom designs not achievable with existing components. Use it to apply styling such as backgrounds, padding, or borders, or to nest and group other components. The contents of Box always maintain their natural size, making it especially useful within layout components that would otherwise stretch their children.',
```
oliverigor:
```suggestion
    'A generic container that provides a flexible alternative for custom designs not achievable with existing components. Use it to apply styling such as backgrounds, padding, or borders, or to nest and group other components. The contents of `Box` always maintain their natural size, making it especially useful within layout components that would otherwise stretch their children.',
```
?

packages/ui-extensions/src/docs/shared/components/Heading.ts
```diff
@@ -3,7 +3,7 @@ import type {SharedReferenceEntityTemplateSchema} from '../docs-type';
 const data: SharedReferenceEntityTemplateSchema = {
   name: 'Heading',
   description:
-    'The `s-heading` component renders hierarchical titles similar to HTML’s `h1-h6` elements. Heading levels are automatically determined by parent `s-section` nesting depth starting from `h2`. Always prefer using the `heading` property on `s-section` and `s-banner` components before constructing custom layout with `s-heading`.',
+    'Renders hierarchical titles to communicate the structure and organization of page content. Heading levels adjust automatically based on nesting within parent Section components, ensuring a meaningful and accessible page outline.',
```
oliverigor, line 6:
```suggestion
    'Renders hierarchical titles to communicate the structure and organization of page content. Heading levels adjust automatically based on nesting within parent `Section` components, ensuring a meaningful and accessible page outline.',
```
? or are we doing this only when mentioning the custom element ?

packages/ui-extensions/src/docs/shared/components/Text.ts
```diff
@@ -3,7 +3,7 @@ import type {SharedReferenceEntityTemplateSchema} from '../docs-type';
 const data: SharedReferenceEntityTemplateSchema = {
   name: 'Text',
   description:
-    'The `s-text` component is used for inline text styling, similar to HTML `span` elements, not for paragraph-level content. Use it to apply different visual tones and text styles to specific words or phrases within a `s-paragraph`, such as a `strong` type or `critical` tone.',
+    'Displays inline text with specific visual styles or tones. Use to emphasize or differentiate words or phrases within a Paragraph or other block-level components.',
```
oliverigor, line 6:
```suggestion
    'Displays inline text with specific visual styles or tones. Use to emphasize or differentiate words or phrases within a `Paragraph` or other block-level components.',
```

packages/ui-extensions/src/surfaces/checkout/components/Heading/Heading.doc.ts
```diff
@@ -28,13 +28,29 @@ const data: ReferenceEntityTemplateSchema = {
   },
   subSections: [
     {
-      type: 'Generic',
+      title: 'Useful for',
+      type: 'Generic' as const,
```
oliverigor: why do we need the as const here ?

packages/ui-extensions/src/surfaces/checkout/components/Text/Text.doc.ts
```diff
@@ -28,11 +28,22 @@ const data: ReferenceEntityTemplateSchema = {
...
+      type: 'Generic' as const,
       anchorLink: 'best-practices',
-      title: 'Best Practices',
-      sectionContent:
-        '- Create contrast between more and less important text with properties such as `color`.',
+      sectionContent: `
+- Use plain and clear terms.
+- Don't use jargon or technical language.
+- Don't use different terms to describe the same thing.
+- Don't duplicate content.`,
```
oliverigor:
```suggestion
- Use plain and clear terms.
- Don’t use jargon or technical language.
- Don’t use different terms to describe the same thing.
- Don’t duplicate content.`,
```

oliverigor (approved):
:shipit:

sam-b-rose (approved):
Description changes are concise and consistent.

Only thing I would note, @oliverigor suggested adding backticks around the component names. In in admin surface, I've seen components referenced using the element tag name (i.e. `s-text`).

I don't think we should use backticks if we are going to use the TitleCased names (i.e. Box) since it would suggest this is a React-like component `<Box>`.

So, my opinion is we either we use code fence for tag elements or we don't use them at all:

- Box not `Box`
- or `<s-box>` not s-box

oliverigor:
Yeah, it was more like a question from me, if that's what we're are going for:

> Box not Box
> or <s-box> not s-box

I agree with that too :D

packages/ui-extensions/src/docs/shared/components/Heading.ts
jffortier: Yeah, I left these like this intentionally. They are not referencing the code, just the component's name.

packages/ui-extensions/src/surfaces/checkout/components/Heading/Heading.doc.ts
jffortier: Unsure, I copied it from the Amdin side. I can check.

packages/ui-extensions/src/docs/shared/components/Heading.ts
jffortier: It aligns with the comment from Sam below that I saw after answering here.

packages/ui-extensions/src/surfaces/checkout/components/Heading/Heading.doc.ts
oliverigor: I tested out, it's not erroring anything in our side, so I think we should be good without.

packages/ui-extensions/src/docs/shared/components/Heading.ts
sam-b-rose, line 6:
```suggestion
    'Renders hierarchical titles to communicate the structure and organization of page content. Heading levels adjust automatically based on nesting within parent Section components, ensuring a meaningful and accessible page outline. Use only when heading properties on section or banner components aren't sufficient.',
```

jffortier: This sentence is present in the Consideration section down the page. It mentions s-section only, so I'll add `s-banner` or rephrase it so it does not need to include all future components that exposes a `heading` property.

PR #2880, opened 2025-05-16
Use shared component data

sam-b-rose:
This PR:
- [x] Removes `related` links from all component pages. See https://github.com/Shopify/admin-ui-foundations/issues/2853 for context (aligned with @jffortier and @AaronSpringut)
- [x] Adds shared documentation for `Page`
- [x] Consolidates Admin component documentation to use `globalShared` and removed `shared`
- [x] Extensions only documentation is under `*.ext.doc.ts`
- [x] App home only documentation is under `*.ab.doc.ts`

### Background

Part of https://github.com/Shopify/admin-ui-foundations/issues/2647

### Solution

Pull from the shared component data to ensure name, description, categories, subcategories, and related components are consistent.

### Checklist


packages/ui-extensions/src/surfaces/admin/components/Banner/Banner.ab.doc.ts
```diff
@@ -1,11 +1,12 @@
 import {AdminReferenceEntityTemplateSchema} from '../../docs-types';
+import {setRelatedUrl} from '../../uitls';
+
+import globalShared from '../../../../docs/shared/components/Banner';
 import shared from './shared';
 
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
 const data: AdminReferenceEntityTemplateSchema = {
+  ...setRelatedUrl(globalShared, 'app-home'),
```
vividviolet: Could we do this instead?

```ts
const data: AdminReferenceEntityTemplateSchema = {
  ...globalShared,
  ...setRelatedUrl('app-home'),
```

It would be cleaner that way as `setRelatedUrl` doesn't need anything from `globalShared`

alex-page: Related URL's are removed from Admin. See: https://github.com/Shopify/admin-ui-foundations/issues/2853

alex-page (approved):

PR #2895, opened 2025-05-19
Ship one column component layout pages to all components

alex-page:
This PR: 
- [x] Adds better error messaging with `error.toString()`
- [x] Adds a `globalShared` for shared docs for every component
- [x] Moves   `related: [],  isVisualComponent: true,  isOneColumnLayout: true,` to `globalShared`. By doing this we change checkout components to use the new one column layout for component docs.
- [x] Removes `requires: '', type: '',` from checkout component docs

TODO
- [ ] Double check the clipboard component can override the `isVisualComponent: true` value

### DIFF
--- a/packages/ui-extensions/docs/surfaces/admin/build-docs.mjs
+++ b/packages/ui-extensions/docs/surfaces/admin/build-docs.mjs
@@ -238,7 +238,7 @@ try {
 
   await fs.rm(tempComponentDefs);
 } catch (error) {
-  console.error(error);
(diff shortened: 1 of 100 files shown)


sam-b-rose (approved):
Nice consolidation, looks good!

gwyneplaine (approved):

alex-page:
Going to close this as the learn team needs to own shipping the component layout across checkout and admin.

PR #2907, opened 2025-05-20
[UI components] Add docs for Section component

andrewmcgov:
### Background

Added docs for Section component, matching the docs from the Admin surgace.

### 🎩

Test changes with dev docs. 

### Checklist


packages/ui-extensions/src/surfaces/checkout/components/Section/Section.doc.ts
```diff
@@ -0,0 +1,47 @@
...
+- Creating consistency across pages.`,
+    },
+    {
+      title: 'Considerations',
+      type: 'Generic',
+      anchorLink: 'considerations',
+      sectionContent: `- When adding headings inside sections they automatically use a specific style, which helps keep the content organized and clear.
+- Built-in spacing is added between nested sections, as well as between heading and content.
+- **Level 1:** Display as responsive cards with background color, rounded corners, and shadow effects. Includes outer padding that can be removed when content like \`s-table\` needs to expand to the full width of the section.
+- **Nested sections:** Don't have any background color or effects by default.`,
```
jffortier:
```suggestion
      sectionContent: `When adding headings inside sections they automatically use a specific style, which helps keep the content organized and clear.`,
```

We should remove these 3 bullets or rewrite them to be generic enough so that they work in both Customer Account and Checkout. I also don't agree with being super descriptive about the look of Section like Admin does. We should not need to describe the exact details of the design, but just that the style will adapt to its context or nested level. I think the description of the component already mentions that.

Sorry if I directed you to copy this, I didnt read each bullet before I did.

packages/ui-extensions/docs/surfaces/checkout/staticPages/using-polaris-web-components.doc.ts
```diff
@@ -114,8 +114,8 @@ const data: LandingTemplateSchema = {
             tabs: [
               {
                 title: 'JavaScript',
-                code: './examples/properties-vs-attributes-jsx-props.js',
```
oliverigor, line 117: This was needed since we ignore js files when generating docs and copying them over to shopify-dev, thanks @andrewmcgov for flagging this.

packages/ui-extensions/src/surfaces/checkout/components/Section/Section.doc.ts
andrewmcgov: No problem at all. 

I've removed the bottom three points, you are right that the nesting level context is mentioned in the shared description: 

> Sections have contextual styling that automatically adapts based on nesting depth.

oliverigor (approved):
:shipit:

jffortier (approved):
Not a huge fan of the screenshots provided by @rockytori, but we can revisit as a fast follow post Edition. My worry is that Section will not look like that in the majority of the cases where partners will test their extension. It will be a unstyled section, **unless** their partner store has customized the styles of section through the Branding API.

PR #2921, opened 2025-05-21
Export storage API interfaces

Alex-Palad:
### Background

Export Storage API interface to use it in POS App

Also, update `entries` method to return Array instead of  IterableIterator, since Worker serializer can't serialize Iterators

Refactor Storage API (do not include it as a part of BaseData, create new BaseAPI)
We need it for a better POS App experience. BaseData does not depend on extension, but APIs do
https://github.com/Shopify/pos-next-react-native/pull/58468#discussion_r2105079430

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
@@ -1,6 +1,11 @@
-export interface StorageExceededError extends Error {
-  name: 'StorageExceededError';
-  code: 'RecordsCount' | 'RecordSize';
+export class StorageError extends Error {
+  public name = 'StorageError';
+  constructor(
+    public code: 'RecordsCount' | 'RecordSize' | 'KeyType' | 'KeySize',
+    message: string,
+  ) {
+    super(message);
+  }
 }
 export interface Storage<
   BaseStorageTypes extends Record<string, any> = Record<string, unknown>,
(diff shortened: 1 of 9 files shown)


Alex-Palad:
> We detected some changes in `packages/*/package.json` or `packages/*/src`, and there are no updates in the `.changeset` directory. If the changes are user-facing and should cause a version bump, run `yarn changeset` to track your changes and include them in the next release CHANGELOG. If you are making simple updates to repo configuration, examples, or documentation, you do not need to add a changeset.

It is a part of `developer preview` [update](https://github.com/Shopify/ui-extensions/pull/2837/files#diff-778f35c71d60f8a623aa6ddda8dfab8cb477137ac58fef307f6683a0e3c30cb5), so no changeset needed

vctrchu:
@Alex-Palad Isn't a change set required to kick off a new unstable build?

vctrchu (approved):

Alex-Palad:
> @Alex-Palad Isn't a change set required to kick off a new unstable build?

Oh, good point, didn't think about that

packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
```diff
@@ -1,6 +1,11 @@
-export interface StorageExceededError extends Error {
-  name: 'StorageExceededError';
-  code: 'RecordsCount' | 'RecordSize';
+export class StorageError extends Error {
```
fatbattk, line 1: should this be outside the `/types` folder.

Alex-Palad: I thought about that. It is actually a type. I made it a class so tat it is easier to throw an exception

Alex-Palad: I just can't find a better place for it :)

fatbattk (approved):

js-goupil (approved):

PR #2926, opened 2025-05-21
Add onBlur prop to SearchBar

kangjoa:
### Background

Related to issue #[54912](https://github.com/Shopify/temp-project-mover-Archetypically-20250612104008/issues/76). **Problem:** The SearchBar component had an `onFocus` callback without a matching `onBlur`, which created potential UX issues.

TLDR:
- Expose `onBlur` prop to `SearchBar`
- Part of cross-repo change:
    - This PR needs to be merged before these PRs:
        - [pos-e2e-ui-extension PR Shopify/pos-next-react-native#144](https://github.com/Shopify/pos-e2e-ui-extension/pull/144)
        - [pos-next-react-native PR Shopify/pos-next-react-native#61908](https://github.com/Shopify/pos-next-react-native/pull/61908)
 
### Details
|Before|After|
|--|--|
| image|image|
   
### Solution
- Adds `onBlur` to the `SearchBarProps` interface to expose the existing blur functionality from the underlying `InputField` component. This completes the onFocus/onBlur pair.

### 🎩

1. Update `package.json` to use unstable-snap in your ui extension
2. Checkout the  [pos-next-react-native PR Shopify/pos-next-react-native#61908](https://github.com/Shopify/pos-next-react-native/pull/61908), use tophat or checkout and run POS locally
3. Checkout [ui-extensions PR Shopify/pos-next-react-native#2926](https://github.com/Shopify/ui-extensions/pull/2926) and run 'shopify app dev'
4. Run `dev extension [url from Developer Console]` on POS
5. Go to SearchBar Screen and test onBlur (paired with onFocus so switches from True to False and back while focusing or blurring)

### Checklist

### DIFF
--- a/.changeset/itchy-forks-lay.md
+++ b/.changeset/itchy-forks-lay.md
@@ -1,0 +1,6 @@
+---
+'@shopify/ui-extensions': minor
+'@shopify/ui-extensions-react': minor
+---
+
+SearchBar - add onBlur handler
--- a/packages/ui-extensions/src/surfaces/point-of-sale/render/components/SearchBar/SearchBar.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/render/components/SearchBar/SearchBar.ts
@@ -18,6 +18,10 @@ export interface SearchBarProps {
    */
   onFocus?: () => void;
   /**
+   * A callback when focus is removed.
+   */
+  onBlur?: () => void;
+  /**
    * Whether the text can be changed.
    */
   editable?: boolean;


vctrchu (approved):

vctrchu:
I think theres a typo in your PR description? It says this PR is required to be merged before but its refers to the same one. Did you mean to link to the pos-e2e-ui-extension PR?

Screenshot 2025-06-03 at 9 27 49 AM

.changeset/itchy-forks-lay.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/ui-extensions': minor
+'@shopify/ui-extensions-react': minor
```
henryStelle, line 3: I think this might be a patch since it is something that should have already been there. It feels more bug than feature to me.

vctrchu: Valid, however since we adding a new callback I would still consider this as a new feature as we are exposing a _new_ callback and not fixing an existing one.

henryStelle (approved):

packages/ui-extensions/src/surfaces/point-of-sale/render/components/SearchBar/SearchBar.ts
```diff
@@ -17,6 +17,10 @@ export interface SearchBarProps {
    * A callback when the input is focused.
    */
   onFocus?: () => void;
+  /**
+   * The callback when focus is removed.
```
fatbattk: nit: should it follow others with `A callback...`

fatbattk (approved):

NathanJolly (approved):

.changeset/itchy-forks-lay.md
katriciab: Out of curiosity @kangjoa @vctrchu, are there any docs on how to create changesets and overall how they work? I only see this one: https://pos.docs.shopify.io/extensions#repository-and-packages and the note: "Merging a PR with a changeset into unstable causes the CI/CD pipeline for ui-extensions to release a new unstable version". I assume this is the same changeset.

vctrchu: @katriciab Yes it looks like that note explains the high level idea. The in-depth documentation is in the ui-extension repo below:

https://github.com/Shopify/ui-extensions/blob/5bc30524cd9d24fe9478dccee5f17aacb4a36aa5/documentation/versions-and-deploys.md?plain=1#L13-L33

I wonder if this would be beneficial to link to on our internal docs?

katriciab: Ah ok, thanks for the link to the versions-and-deploys. I was looking for `itchy-forks-lay` 😅 (wrong keyword and `yarn changeset` command was what i needed) and didn't find anything even in the ui-extensions docs.

packages/ui-extensions/src/surfaces/point-of-sale/render/components/SearchBar/SearchBar.ts
kangjoa: I'll change this to follow the other ones, thanks! 

I had copied this comment from the `point-of-sale/render/components/shared/InputField.ts` ([link](https://github.com/Shopify/ui-extensions/blob/5bc30524cd9d24fe9478dccee5f17aacb4a36aa5/packages/ui-extensions/src/surfaces/point-of-sale/render/components/shared/InputField.ts#L54)). For documentation, do we generally follow the style of what's in the current doc (vs. "shared" documentation)? 

example from /shared:
```js
  /**
   * The callback when focus is removed.
   */
  onBlur?: () => void;

[...]

  /**
   * The callback when input is focused.
   */
  onFocus?: () => void;
```

fatbattk: ah wasn't aware of the difference with `shared`. typically following the same file's conventions is fine unless it's obviously wrong.

vctrchu: @kangjoa I would just stick with the patterns that exist in the file we are working with. 

My bad on referencing the shared folder!

PR #2927, opened 2025-05-21
fix navigation hook

robin-drexler:
Fixes `useNavigationCurrentEntry` returning stale values
See https://github.com/Shopify/ui-extensions/pull/2927/files#r2101046001 for more context

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/preact/live-navigation.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/preact/live-navigation.ts
@@ -7,20 +7,27 @@ import {useEffect, useReducer} from 'preact/hooks';
 export function useNavigationCurrentEntry<
   Target extends RenderCustomerAccountFullPageExtensionTarget = RenderCustomerAccountFullPageExtensionTarget,
 >(): NavigationHistoryEntry {
-  const {currentEntry, removeEventListener, addEventListener} =
-    useApi<Target>().navigation;
+  const navigation = useApi<Target>().navigation;
 
-  const [entry, update] = useReducer(() => currentEntry, currentEntry);
+  const [entry, update] = useReducer(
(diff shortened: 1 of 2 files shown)


packages/ui-extensions/src/surfaces/customer-account/preact/live-navigation.ts
```diff
@@ -7,20 +7,27 @@ import {useEffect, useReducer} from 'preact/hooks';
 export function useNavigationCurrentEntry<
   Target extends RenderCustomerAccountFullPageExtensionTarget = RenderCustomerAccountFullPageExtensionTarget,
 >(): NavigationHistoryEntry {
-  const {currentEntry, removeEventListener, addEventListener} =
-    useApi<Target>().navigation;
+  const navigation = useApi<Target>().navigation;
 
-  const [entry, update] = useReducer(() => currentEntry, currentEntry);
+  const [entry, update] = useReducer(
+    () => navigation.currentEntry,
```
robin-drexler, line 13: So this should be the biggest change and I'm not sure why this ever worked. Potentially we got lucky with a race condition in react or a rerender of the hook occured that updated `currentEntry` when now that's not the case anymore.

`navigation.currentEntry` is a value that is being mutated and never triggers a rerender on its own.

So when we ran the reducer, it'd actually set the previous currentEntry value instead of the current one. 

Let's say you start on `/foo`, then you navigate to `/bar`.
The hook would return `/foo` instead of `/bar`.
When you go back to `/bar`, the hook returns `/foo` instead of `/bar`.

Now we're always reading the current version straight from the `navigation` object which guarantees that's it's always the latest

andrewmcgov (approved):
Nice fix! And we 🎩'd together

PR #2932, opened 2025-05-22
Changeset

oliverigor:
### Background

- Adding changeset for release

### Solution

- Added summary of changes that happened in the last days

### Checklist

### DIFF
--- a/.changeset/thirty-paws-rush.md
+++ b/.changeset/thirty-paws-rush.md
@@ -1,0 +1,16 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+- Added customer account migration guide with comparison tables mapping legacy components to new Polaris web components
+- Removed extended form field event properties from the documentation to simplify component APIs
+- Added documentation for Section component with visual examples
+- Improved parent-child component relationship documentation for admin components
+- Added script to exit pre-release mode
+- Updated component descriptions and documentation for Admin UI components
+- Fixed Preact native types for customer account
+- Moved example country selector to top of documentation
+- Fixed form components events and HTML examples
+- Added support for Polaris web components in checkout extensions
+- Added scaffolded with Preact section for checkout extensions
+- Updated documentation for various components with copy edits and improved examples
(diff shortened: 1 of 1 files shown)


.changeset/thirty-paws-rush.md
```diff
@@ -0,0 +1,16 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+- Added customer account migration guide with comparison tables mapping legacy components to new Polaris web components
```
kumar303, line 5: I don't think we need to include docs in the changesets 🤔 It won't hurt, though.

.changeset/thirty-paws-rush.md
```diff
@@ -0,0 +1,16 @@
...
+- Removed extended form field event properties to simplify component APIs
```
kumar303: This was just in the docs, right? I would clarify that because a developer might see _Removed_ and think something was removed from the package.

kumar303 (approved):
I say ship it so we can get the release onto NPM 🚀 

I don't think it needs to include doc changes, though. It should be a description for developers to read about what changed in the RC version of the package. However, I don't know where they would even read the changelog 😅 

I don't see it in either of these places:
* https://www.npmjs.com/package/@shopify/ui-extensions/v/2025.10.0-rc.2
* https://github.com/Shopify/ui-extensions/releases/tag/%40shopify%2Fui-extensions%402025.10.0-rc.2

.changeset/thirty-paws-rush.md
oliverigor: yeah, will modify

.changeset/thirty-paws-rush.md
oliverigor: I'll leave it, just in case haha

.changeset/thirty-paws-rush.md
```diff
@@ -0,0 +1,16 @@
+---
+'@shopify/ui-extensions': minor
```
vividviolet: Hmm shouldn't this be a patch?

oliverigor: @kumar303 suggested as a minor, but I can change to a patch no problem

oliverigor: changed to patch @vividviolet

PR #2944, opened 2025-06-02
remove ui api docs for checkout

lrsterrett:
### Background

Short-term adjustment: Removing UI api docs from 2025-10-rc until the relevant new polaris component (modal) is released. The docs and examples don't make sense and don't work until then

### Solution

image

When accessing the page from a different api version and switching to 2025-10-rc:
image

### 🎩

- ...

### Checklist


oliverigor (approved):
thanks for doing this :D 

Can you run the docs and update the shopify-dev repo as well.

PR #2945, opened 2025-06-03
Convert Refactor StorageError into Interface

Alex-Palad:
### Background

UI-extensions can only export interfaces, not classes. We can't import and use this class in pos app.
https://github.com/Shopify/ui-extensions/blob/ed23ff862497a9a72cc2534411f19ded5adf02f3/config/typescript/tsconfig.base.json#L7

### Solution

Replace class with Interface to define error type

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
@@ -1,12 +1,8 @@
-export class StorageError extends Error {
-  public name = 'StorageError';
-  constructor(
-    public code: 'RecordsCount' | 'RecordSize' | 'KeyType' | 'KeySize',
-    message: string,
-  ) {
-    super(message);
-  }
+export interface StorageError extends Error {
+  name: 'StorageError';
+  code: 'RecordsCount' | 'RecordSize' | 'KeyType' | 'KeySize';
 }
+
 export interface Storage<
   BaseStorageTypes extends Record<string, any> = Record<string, unknown>,
 > {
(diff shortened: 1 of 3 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/types/storage.ts
```diff
@@ -1,12 +1,8 @@
...
-  public name = 'StorageError';
-  constructor(
-    public code: 'RecordsCount' | 'RecordSize' | 'KeyType' | 'KeySize',
-    message: string,
-  ) {
-    super(message);
-  }
+export interface StorageError extends Error {
+  name: 'StorageError';
+  code: 'RecordsCount' | 'RecordSize' | 'KeyType' | 'KeySize' | string;
```
fatbattk: `code` is now also `string`?

Alex-Palad: So that we could add new error codes easier if we need.

fatbattk: think should remove `string` since that makes the other literal values pointless (type safety gone, etc).

Alex-Palad: Done

fatbattk (approved):

vctrchu:
Was this change discovered after testing on POS? I'm just wondering if the original interface was tested before merging.

I'm only asking because I'm seeing a lot of minor unstable patch releases (eg. making the storage api optional) indicating it might be better use yalc or the command `/snapit` to test the dependencies

vctrchu (approved):

Alex-Palad:
> Was this change discovered after testing on POS? I'm just wondering if the original interface was tested before merging.
> 
> I'm only asking because I'm seeing a lot of minor unstable patch releases (eg. making the storage api optional) indicating it might be better use yalc or the command `/snapit` to test the dependencies

Yes, it was tested. The problem is that local build has js files, and import actually worked. 

For optional storage we had another issue - we planned to merge RN pr right after ui-extensions changes, but then we got some feedback that blocked that PR. 
Another team wanted to update ui-extensions version, and mandatory storage field blocked them.

PR #2970, opened 2025-06-10
Allow key props for customer accounts components in extensions

SHAZAM28:
**Description**

Original Issue: https://github.com/shop/issues-checkout/issues/5803

Using the `key` and `ref` props on our elements with preact

Introduce two new type interfaces to standardize prop typing for Preact components:
- `BaseElementProps`: for elements without children, providing `key`, `ref`, and `slot` attributes.
- `BaseElementPropsWithChildren`: extends `BaseElementProps` to include the `children` attribute.

These interfaces replace previously duplicated `BaseProps` interfaces that were defined in individual component files, establishing a consistent and reusable typing approach across all components.

Similar to Checkout: https://github.com/shop/world/pull/64309

## Overview
This PR adds support for key props in customer accounts components within UI extensions, enhancing the ability to uniquely identify and manage components in the customer account interface.

## Changes
- Added support for key props in customer account components
- Implemented changes across multiple customer account components including:
  - CustomerAccountAction
  - Page
  - ImageGroup

## Technical Details
- Modified component interfaces to support key props
- Updated type definitions to ensure proper TypeScript support
- Maintained compatibility with existing customer account extension targets

## Impact
This change improves component identification and management in customer account extensions, which is particularly important for:
- Dynamic component rendering
- Component state management
- Performance optimization through proper key usage
- Better debugging and development experience

## Testing
The changes have been tested with:
- Customer account extension targets
- Various component configurations
- TypeScript type checking

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/components/CustomerAccountAction.d.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/components/CustomerAccountAction.d.ts
@@ -1,3 +1,5 @@
+import {BaseElementPropsWithChildren} from './shared';
+
 export interface CustomerAccountActionProps {
   /**
    * Sets the heading of the Action container.
@@ -16,14 +18,11 @@ declare global {
 }
 
 declare module 'preact' {
-  interface BaseProps {
-    children?: preact.ComponentChildren;
-    slot?: Lowercase<string>;
-  }
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace createElement.JSX {
     interface IntrinsicElements {
-      ['s-customer-account-action']: BaseProps & CustomerAccountActionProps;
+      ['s-customer-account-action']: BaseElementPropsWithChildren<CustomerAccountActionElement> &
+        CustomerAccountActionProps;
     }
   }
 }
--- a/packages/ui-extensions/src/surfaces/customer-account/components/shared.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/components/shared.ts
@@ -12,3 +12,20 @@ export type Size =
   | 'large'
   | 'extraLarge'
   | 'fill';
+
+export interface BaseElementProps<TClass = HTMLElement> {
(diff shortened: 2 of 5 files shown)


packages/ui-extensions/src/surfaces/customer-account/components/shared.ts
```diff
@@ -12,3 +12,20 @@ export type Size =
...
   | 'extraLarge'
   | 'fill';
+
+export interface BaseElementProps<TClass = HTMLElement> {
+  // Assigns a unique key to this element.
+  key?: preact.Key;
+  // Assigns a ref (generally from `useRef()`) to this element.
+  ref?: preact.Ref<TClass>;
+  // Assigns this element to a parent's slot.
+  slot?: Lowercase<string>;
```
oluwatimio, line 22: Slot is not actually available on all components 🤔  so Im wondering if we should remove this from BaseElementProps. We also dont even have any slot on `s-page` itself, just on Its button children instead. Im wondering if we just want this to just be available everywhere even if it may not have any slots. Looks like its the approach checkout is taking so perhaps its fine

oluwatimio (approved):
Tested this out, its fine for me but Im curious what other's think about slot being available all the time even if the component may not have a valid slot (its also what checkout is doing)

packages/ui-extensions/src/surfaces/customer-account/components/shared.ts
oluwatimio: Asked about this in Slack!

packages/ui-extensions/src/surfaces/customer-account/components/CustomerAccountAction.d.ts
```diff
@@ -1,3 +1,5 @@
+import {BaseElementPropsWithChildren} from './shared';
```
robin-drexler, line 1: would be nice to use a shared version with checkout once they have updated theirs here as well.

robin-drexler (approved):

PR #2988, opened 2025-06-17
Add refunds data to reprint receipt input data

Alex-Palad:
### Background

Part 1/2 of https://github.com/shop/issues-retail/issues/2961

### Solution

Add `refunded` fields to LineItems for reprint event

I defined a new OrderLineItems extends LineItems interface to be safe and not to add non-optional field to all usages of LineItems (which is a part of Cart API, and Cart API does not deal with refunds)

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/order.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/order.ts
@@ -1,0 +1,12 @@
+import {LineItem} from './cart';
+
+export interface LineItemRefund {
+  createdAt: string;
+  quantity: number;
+}
+
+export interface OrderLineItem extends LineItem {
+  // Number of remaining goods
+  currentQuantity: number;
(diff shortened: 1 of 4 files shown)


fatbattk (approved):

packages/ui-extensions/src/surfaces/point-of-sale/types/order.ts
```diff
@@ -0,0 +1,12 @@
+import {LineItem} from './cart';
+
+export interface LineItemRefund {
+  createdAt: string;
+  quantity: number;
+}
+
+export interface OrderLineItem extends LineItem {
+  // Number of unreturned goods
```
fatbattk: "unreturned" is not clear. maybe "remaining"?

Alex-Palad: Updated. Thank you

PR #2986, opened 2025-06-16
Add `editable` cart property

MrGVSV:
See https://github.com/shop/issues-retail/issues/7833

PR #3002, opened 2025-06-20
Update docs for `editable` property

MrGVSV:
Resolves https://github.com/shop/issues-retail/issues/8521

### Background

The changes made in #2986 need their documentation to be updated.

### Solution

Adds release notes for:
- Added `editable` property
- Added `useCartEditable` hook
- Added `CartNotEditableError` error class

This PR also changes the API method docs to use `Throws` instead of `@throws` as our internal documentation engine doesn't support those tags yet. Ideally, we'd be able to make use of such tags, but for now I think it makes sense to just do without them.

#### Question

Now `editable` affects all cart methods, throwing an error when the cart is not editable. Should we update all examples to include the `editable` check or is that simply noise?

For example, the react examples would probably look like:

```tsx
  const editable = useCartEditable();
  // ...
  return (
    <Tile 
      // ...
      onPress={() => editable && api.cart.setCustomer({
        id: 1,
      })}
    />
  );
```

And TypeScript examples would look like:

```ts
  const tile = root.createComponent(Tile, {
    // ...
    onPress: () => api.cart.subscribable.initial.editable && api.cart.setCustomer({
      id: 1,
    }),
  });
  // ...
 api.cart.subscribable.subscribe((newCart: Cart) => {
    tile.updateProps({
      onPress: () => editable && api.cart.setCustomer({
        id: 1,
      }),
    });
  });
```

Note that the other way of handling cart editability is to disable the tile when `editable` is false. So we could also update the examples in a similar way but just modifying `enabled` instead of doing `editable && ...`.

How do we feel about including these in the examples? Does it detract from the actual example too much?

### 🎩

https://pos.docs.shopify.io/extensions/contributing/documentation#how-to-update-docs-for-stable-api-versions

Here is the result:

| API Docs | Release Notes | Example |
| -------- | ------------- | ------- |
| Cart API Docs | Release Notes | Example |

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
@@ -44,6 +44,8 @@ const data: LandingTemplateSchema = {
 - Added optional \`taxLines\` property to [ShippingLine](/docs/api/pos-ui-extensions/targets/post-transaction/pos-transaction-complete-event-observe#transactioncompletedata-propertydetail-transaction) interface.
 - Added optional \`onBlur\` handler to [SearchBar](/docs/api/pos-ui-extensions/components/searchbar) component.
 - Added optional \`tone\` property to [Icon](/docs/api/pos-ui-extensions/components/icon) component and expanded \`name\` and \`size\` options.
+- Added optional \`editable\` property to \`Cart\` interface.
+- Added \`useCartEditable\` hook to access the cart's editable state.
 
 ### Developer Preview
 
--- a/packages/ui-extensions/src/surfaces/point-of-sale/render/api/cart-api/cart-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/render/api/cart-api/cart-api.ts
@@ -30,19 +30,17 @@ export interface CartApiContent {
   subscribable: RemoteSubscribable<Cart>;
 
   /** Bulk update the cart
+   *
    * @param cartState the cart state to set
    * @returns the updated cart
-   *
-   * @throws CartNotEditableError if the cart is not currently editable.
    */
   bulkCartUpdate(cartState: CartUpdateInput): Promise<Cart>;
 
   /** Apply a cart level discount
+   *
    * @param type the type of discount applied (example: 'Percentage')
    * @param title the title attributed with the discount
    * @param amount the percentage or fixed monetary amount deducted with the discount. Pass in `undefined` if using discount codes.
-   *
-   * @throws CartNotEditableError if the cart is not currently editable.
    */
   applyCartDiscount(
     type: CartDiscountType,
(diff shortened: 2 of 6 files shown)


MrGVSV:
* **#3002** <a href="https://app.graphite.dev/github/pr/Shopify/ui-extensions/3002?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ui-extensions/3002?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `unstable`


<h2></h2>

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
```diff
@@ -67,6 +67,9 @@ Refer to the [migration guide](/docs/api/pos-ui-extensions/migrating) for more i
 - Added required \`variantId\` property to \`ProductApi\` interface.
 - Added optional \`taxLines\` property to \`ShippingLine\` interface.
 - Added optional \`onBlur\` handler to \`SearchBar\` component.
+- Added optional \`editable\` property to \`Cart\` interface.
+- Added \`useCartEditable\` hook to access the cart's editable state.
+- Added a new \`CartNotEditableError\` error class, thrown when attempting to modify the cart while it is not currently editable.
```
vctrchu: I'm not sure if we want to include `CartNotEditableError ` here, it feels like an implementation detail more than a feature.

vctrchu (commented):
What is the `editable` default value? true?

MrGVSV:
> What is the `editable` default value? true?

POS will default to `true`, yes. But to ease migration, we're also treating `undefined` as `true` as well.

vctrchu:
> POS will default to true, yes. But to ease migration, we're also treating undefined as true as well.

If the default is true I'd say its fine to not update all the examples since it won't affect existing implementations. A single example showcasing the specific editable and error should be sufficient.

packages/ui-extensions/src/surfaces/point-of-sale/render/api/cart-api/cart-api.ts
vctrchu, line 1: @NathanJolly @js-goupil Thoughts on this? Every API call can possibly throw an editable error.

https://camo.githubusercontent.com/a4ef7e38bdc731aa286f4c0f361dee9e5ec95927d22c2e3de32b159c63681914/68747470733a2f2f67726170686974652d757365722d75706c6f616465642d6173736574732d70726f642e73332e616d617a6f6e6177732e636f6d2f4279346d57585852674b664c31646f53344c69662f30313333313262622d396430652d343931382d393534652d3061653365303135646661632e706e67

vctrchu: @aaronschubert0 Do you have any thoughts?

aaronschubert0: @MrGVSV If extensions don't directly handle `CartNotEditableError`, how does it affect the app?

My preference would be that in the case of `CartNotEditableError` we return an error from these methods rather than throwing an error.

lrsterrett:
Was this doc change intended to be ready for 2025-07 stable release planned for today?

packages/ui-extensions/src/surfaces/point-of-sale/render/api/cart-api/cart-api.ts
MrGVSV: > If extensions don't directly handle `CartNotEditableError`, how does it affect the app?

We have a tech doc  that goes into more detail, but failing to handle the error would result in the default UI Extension error behavior (i.e. the extension enters an error state and/or a toast is displayed).

> My preference would be that in the case of `CartNotEditableError` we return an error from these methods rather than throwing an error.

The reason we throw instead of following the error-as-value approach is that we want to reduce the overall changes introduced here. The idea is that this is just a stopgap until we can get certain other features that allow for a cleaner workflow.

vctrchu: When we say reduce the overall changes introduced here, are speaking about the ui-extensions or POS? I think @aaronschubert0 is referring to throwing an error like [this](https://github.com/Shopify/pos-next-react-native/blob/c53dca5e8bdf6d9cbac18b4b065306b535a3e56a/src/extensibility/api/base/useCartApi.ts#L184). Would this be too much work?

MrGVSV: Oh I think I misunderstood then. Yes the plan is to throw the `CartNotEditableError` like that. I was thinking that the suggestion was to change the methods to return either a success or error state. That would be beyond the scope we want to introduce as part of our temporary solution here.

vctrchu: Throwing a typed error is good. But should the error be more general? Using `CartNotEditableError` for every method feels off. A general `CartError` gives us more flexibility. If requirements change, we can add new error types without breaking things or needing a big refactor. What do you think?

MrGVSV: Yeah that works. The benefit of `CartNotEditableError` was that it would allow for more fine-grained and type-safe control over how users handle those errors. But we can loosen the type-safety and use a generic `CartError`.

We actually aren't making use of any of these changes in POS yet (that should come in the next couple months), so we have the freedom to choose whatever type of error we throw.

I guess the one question I have is, if we did want to use `CartError`, would we be able to add that in a patch release for 2025-07? Or is adding it a problem?

If it is a problem and we do want to avoid using `CartNotEditableError`, we can also consider using a basic `Error` for now.

vctrchu: I'm good with using a basic `Error` for now as I think creating an `CartError` type should require a separate PR and reasoning itself. Sorry for the back and forth, this should keep you building for the time being. 

You can just remove the `Throws` comments in the `useCartApi` and we can merge the rest.

MrGVSV: Done! I also marked the `CartNotEditableError` type as deprecated

vctrchu (dismissed):

vctrchu:
Sorry this needs to be merged to https://github.com/Shopify/ui-extensions/tree/2025-07. We don't merge changes to `unstable` anymore.

MrGVSV:
Git's giving me a hard time but I think I managed to rebase onto 2025-07!

vctrchu (approved):

PR #3011, opened 2025-06-25
[Remote-dom]: Migration guide updates

oliverigor:
### Background

- Bring the new updated components and guides for developers

### Solution

- Adding correct files
- Changing route for links on the migration table

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/checkout/build-docs.sh
+++ b/packages/ui-extensions/docs/surfaces/checkout/build-docs.sh
@@ -14,7 +14,10 @@ fail_and_exit() {
 run_sed() {
   if [[ "$OSTYPE" == "darwin"* ]]; then
     # macOS
-    sed -Ei '' "$1" "$2"
+    # The --in-place flag is only needed for sed Gnu 4.9 which is what
+    # checkout-web uses in its nix store. This is NOT compatible with the
+    # ui-extensions repo.
+    sed --in-place='' -E "$1" "$2"
   else
     # Linux and other Unix-like systems
     sed -ri "$1" "$2"
--- a/packages/ui-extensions/docs/surfaces/checkout/build-docs.sh
+++ b/packages/ui-extensions/docs/surfaces/checkout/build-docs.sh
@@ -14,10 +14,7 @@ fail_and_exit() {
 run_sed() {
   if [[ "$OSTYPE" == "darwin"* ]]; then
     # macOS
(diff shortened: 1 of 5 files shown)


packages/ui-extensions/docs/surfaces/checkout/build-docs.sh
```diff
@@ -13,7 +13,10 @@ fail_and_exit() {
 run_sed() {
   if [[ "$OSTYPE" == "darwin"* ]]; then
     # macOS
-    sed -Ei '' "$1" "$2"
+    # The --in-place flag is only needed for sed Gnu 4.9 which is what
+    # checkout-web uses in its nix store. This is NOT compatible with the
+    # ui-extensions repo.
+    sed --in-place='' -E "$1" "$2"
```
kumar303: This breaks `yarn docs:checkout 2025-10-rc` so you can't copy this part into `ui-extensions`. There's a difference in `nix` versions between the two repos so we need to fix that (but not in this PR 😅)

packages/ui-extensions/docs/surfaces/checkout/staticPages/upgrading-to-2025-10.doc.ts
kumar303, line 1: This is missing changes to previous sections like _Migrate API calls_. What if you just removed the component changes temporarily, copied the `checkout-web` changes, then put back the component changes?

kumar303 (commented):
Sorry that this is so annoying 😰 After this change you'll also need to bring your component changes back to `checkout-web` so that both are in sync.

packages/ui-extensions/docs/surfaces/checkout/build-docs.sh
oliverigor: ohh, good to know.

packages/ui-extensions/docs/surfaces/checkout/staticPages/upgrading-to-2025-10.doc.ts
oliverigor: I have it [here](https://github.com/shop/world/pull/68388), bringing all the changes we did for this batch.

will do as you suggested

kumar303:
I filed this to synchronize nix versions https://github.com/shop/issues-checkout/issues/6176

kumar303 (approved):
Thanks, looks good ✨ We'll just need a follow-up PR in `checkout-web` for the component changes.

PR #3018, opened 2025-06-26
Updates for 2025-07 release

NathanJolly:
### Background

Part of https://github.com/shop/issues-retail/issues/1907

### Solution

Update the versions document so it's ready for 2025-07's release on July 3.

I also removed the sections for 1.0 through 1.7 as they are all unsupported.

### 🎩

- Verify the docs generate correctly

### Checklist


packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
```diff
@@ -41,22 +41,14 @@ Refer to the [migration guide](/docs/api/pos-ui-extensions/migrating) for more i
       type: 'Generic',
       anchorLink: '202507',
       title: '2025.07',
-      sectionNotice: [
-        {
-          type: 'critical',
-          title: 'Not yet supported',
-          sectionContent: `This version is not yet compatible with POS. Extensions targeting this version will be rejected by POS and will not run. Please wait for a compatible POS update before deploying extensions targeting this version.`,
-        },
-      ],
```
fatbattk, line 50: can we create the next version here and move this there? or when does it usually get made?

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
```diff
@@ -308,186 +300,6 @@ Refer to the [migration guide](/docs/api/pos-ui-extensions/migrating) for more i
...
-
-### Fixes
-
-- An unremovable scanning icon was removed from \`SearchBar\`.
-- Icon sizes were adjusted for \`SearchBar\` to avoid cropping.
-- \`FormattedTextField\` now doesn't crash on \`currency\` value for \`inputType\`.
-- Removed multiple broken \`inputType\` values for \`FormattedTextField\`.
-- Resolved multiple path issues with the package.
-`,
-    },
```
fatbattk, line 493: have we considered a new section to archive these instead? (e.g.- `Older version`, `Deprecated`, `Legacy version`)
imo these have value for partners and AI still even if deprecated.

NathanJolly: Interesting! Can you explain the value? Also please note that this is not deprecated, but removed permanently. Deprecated means that the feature can still be used, which is not true here.

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
NathanJolly: This notice was specific to 2025-07 at the release of 2025-04, and is no longer valid. There's no need to move it to a new version. In fact, this notice should not have been added in the first place, as the system of releasing the "next" version of `ui-extensions` at the time of a cut is a long time practice.

We have a large body of work to do for the next version, given it's the RemoteDOM version. So I'd like to keep those concerns separate for now.

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
fatbattk: ah, these are removed. 🤔 still think there's archival value but looks like we remove the docs entirely as well so keeping this may include broken links.

fatbattk (approved):

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
```diff
@@ -13,69 +13,40 @@ const data: LandingTemplateSchema = {
...
-- Update pos.draft-order-details.block.render allowed components to \`BlockComponents\`.
-- Added support for a product variant id field in the \`ProductApi\` interface.
+- Removed the deprecated \`FormattedTextField\` component. Use the [TextField](/docs/api/pos-ui-extensions/components/textfield) component instead. It will continue to work in POS 10.6.0, but will no longer function in POS 10.7.0.
 
-### Breaking Changes
+## Deprecations
 
-- Removed the deprecated \`FormattedTextField\` component. Use the \`TextField\` component instead.
+- Deprecated \`'minor'\`, \`'major'\`, \`'spot'\`, \`'caption'\`, \`'badge'\` as values for the \`size\` prop in the [Icon](/docs/api/pos-ui-extensions/components/icon) component. Use \`'s'\`, \`'m'\`, \`'l'\`, \`'xl'\` instead.
+- Deprecated \`'arrow'\`, \`'available-at-other-locations'\`, \`'collections'\`, \`'connectivity-warning'\`, \`'delivery'\`, \`'home'\`, \`'image-placeholder'\`, \`'internet'\`, \`'menu'\`, \`'orders'\`, \`'products'\`, \`'shipment'\` as values for the \`name\` prop in the [Icon](/docs/api/pos-ui-extensions/components/icon) component. See the available icon list names in the \`name\` prop documentation.
```
kangjoa: nit: do you think removing "list" makes this clearer? (Sorry, I wrote this original copy!) Suggestion to change last line to: 
```
See the available icon names in the \`name\` prop documentation.
```

kangjoa (approved):

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
NathanJolly: Great call, yes, I agree!

NathanJolly: I copy-paste this into the changelog post and a technical writer actually caught this, so I will go with the way they worded it:
image

PR #3027, opened 2025-07-02
Add Avatar and update component definitions

sam-b-rose:
Part of https://github.com/Shopify/admin-ui-foundations/issues/357

- Adds the Avatar component to UI Extensions
- Updates UI Extensions with latest Admin UI Components version


vividviolet (commented):
In order to make this component available we need to add it to the Standard component set here: https://github.com/Shopify/ui-extensions/blob/2025-10-rc/packages/ui-extensions/src/surfaces/admin/components/StandardComponents.ts

PR #3028, opened 2025-07-02
Add announcement api to announcement targets

brianshen1990:
### Background

Add announcement api to announcement targets so that
- extensions can call `api.close` to close the extension 
- extension can know when the extension is closed

UI-API-DESIGN side PR https://github.com/Shopify/ui-api-design/pull/1107 

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout/api/announcement/announcement.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/api/announcement/announcement.ts
@@ -1,0 +1,7 @@
+export interface Announcement {
+  announcement: {
+    close(): void;
+    addEventListener(type: 'close', cb: () => void): void;
+    removeEventListener(type: 'close', cb: () => void): void;
+  };
+}
--- a/packages/ui-extensions/src/surfaces/checkout/targets.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/targets.ts
@@ -24,6 +24,7 @@ import type {
   AllowedComponents,
   AnyComponentExcept,
 } from './shared';
+import {Announcement} from './api/announcement/announcement';
 
 /**
  * A UI extension will register for one or more extension targets using `shopify.extend()`.
@@ -652,7 +653,8 @@ export interface RenderExtensionTargets {
    */
   'purchase.thank-you.announcement.render': RenderExtension<
     OrderConfirmationApi &
-      StandardApi<'purchase.thank-you.announcement.render'>,
+      StandardApi<'purchase.thank-you.announcement.render'> &
+      Announcement,
     AnyComponent
   >;
 }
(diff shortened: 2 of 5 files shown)


packages/ui-extensions/src/surfaces/checkout/api/announcement/announcement.ts
```diff
@@ -0,0 +1,5 @@
+export interface AnnouncementApi {
+  close(): void;
```
brianshen1990: extracted this to more general API

brianshen1990: add directly to the target

packages/ui-extensions/src/surfaces/checkout/targets.ts
```diff
@@ -652,7 +652,11 @@ export interface RenderExtensionTargets {
    */
   'purchase.thank-you.announcement.render': RenderExtension<
     OrderConfirmationApi &
-      StandardApi<'purchase.thank-you.announcement.render'>,
+      StandardApi<'purchase.thank-you.announcement.render'> & {
+        close(): void;
```
brianshen1990: we do not want `AnnouncementApi`, so I am just adding the method directly to the target

dnagoda:
Are there doc updates that need to be made to support dev usage of this part of the API? Here's what we added for the target: https://github.com/Shopify/ui-extensions/pull/3028

packages/ui-extensions/src/surfaces/checkout/targets.ts
dnagoda: Does this project support docs for ad hoc APIs like this?

brianshen1990:
> Are there doc updates that need to be made to support dev usage of this part of the API? Here's what we added for the target: #3028

hey @dnagoda  , we are not adding docs at this moment. It will happen when we have editor support, style(color scheme/animation/expand collapse)~ they will be fast follow ups for this month(July). We will have an google doc for partners who want to use this target. 

and i think you added a wrong link~~

dnagoda (approved):

packages/ui-extensions/src/surfaces/checkout/targets.ts
```diff
@@ -652,7 +652,13 @@ export interface RenderExtensionTargets {
    */
   'purchase.thank-you.announcement.render': RenderExtension<
     OrderConfirmationApi &
-      StandardApi<'purchase.thank-you.announcement.render'>,
+      StandardApi<'purchase.thank-you.announcement.render'> & {
+        announcement: {
```
robin-drexler: can we extract this into a type/interface and use it here? Like we do it for others? Like `CartLineItemApi` https://github.com/Shopify/ui-extensions/blob/9c8ecec8c142ed937e768cd40437822183dd1e98/packages/ui-extensions/src/surfaces/checkout/targets.ts#L230

robin-drexler (approved):

PR #3039, opened 2025-07-07
Add Avatar and Thumbnail components with documentation and examples for Admin UI Extensions

sam-b-rose:
Adds the docs `Avatar` and `Thumbnail` components

## 🎩 

```sh
git checkout add-thumbnail-and-avatar-docs
yarn docs:admin 2025-10-rc
```

Then in `shopify-dev`

```sh
dev up && dev s
```

Go to 

- [Avatar page](https://shopify-dev.myshopify.io/docs/api/app-home/polaris-web-components/media/avatar)
- [Thumbnail page](https://shopify-dev.myshopify.io/docs/api/app-home/polaris-web-components/media/thumbnail)

avatar
thumbnail

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components/Avatar/examples/default.html
+++ b/packages/ui-extensions/src/surfaces/admin/components/Avatar/examples/default.html
@@ -1,0 +1,4 @@
+<s-avatar
+  alt="John Doe"
+  initials="JD"
+></s-avatar>
--- a/packages/ui-extensions/src/surfaces/admin/components/Thumbnail/examples/default.html
+++ b/packages/ui-extensions/src/surfaces/admin/components/Thumbnail/examples/default.html
@@ -1,0 +1,4 @@
+<s-thumbnail
+  alt="White sneakers"
+  src="/assets/products/sneakers.jpg">
+</s-thumbnail>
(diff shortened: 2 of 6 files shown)


sam-b-rose:
cc @jffortier since these are adding shared docs for Avatar and Thumbnail

alex-page (approved):
Thanks for the screenshots. Looks great.

packages/ui-extensions/src/surfaces/admin/components/Avatar/examples/default.html
```diff
@@ -0,0 +1,4 @@
+<s-avatar
+  alt="James Larson"
+  initials="JL"
```
jffortier: Unsure about using the name of a real person. Could we go with `John Doe` or something like that?

packages/ui-extensions/src/surfaces/admin/components/Thumbnail/examples/default.html
```diff
@@ -0,0 +1,4 @@
+<s-thumbnail
+  alt="Thumbnail"
```
jffortier: We should lead by example and actually follow our guidelines for the `alt` attribute. 
```suggestion
  alt="White sneakers"
```

jffortier (approved):
Left some feedback, but otherwise good to ship after addressing it.

PR #3044, opened 2025-07-08
Fix Form event handler types

sam-b-rose:
### Background

Fixes https://github.com/Shopify/admin-ui-foundations/issues/2932

> [!IMPORTANT]
> Should this target `2025-10-rc` or an earlier release?

### Solution

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components.d.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components.d.ts
@@ -6451,7 +6451,6 @@ export interface CustomerSegmentTemplateJSXProps
 export interface FormProps {}
 declare class Form extends PreactCustomElement implements FormProps {
   constructor();
-  accessor onsubmit: ((event: ExtendableEvent) => void) | null;
 }
 declare global {
   interface HTMLElementTagNameMap {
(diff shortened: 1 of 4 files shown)


packages/ui-extensions/src/surfaces/admin/components.d.ts
```diff
@@ -6451,7 +6451,6 @@ export interface CustomerSegmentTemplateJSXProps
 export interface FormProps {}
 declare class Form extends PreactCustomElement implements FormProps {
   constructor();
-  accessor onsubmit: ((event: ExtendableEvent) => void) | null;
```
vividviolet, line 6454: Why do we need to delete this?

sam-b-rose: This removal is correct.

There shouldn't be events in the `FormProps` type, but in the `FormEvent` type instead. The `FormJSXProps` should have both (i.e. `onSubmit`).

vividviolet (dismissed):

packages/ui-extensions/src/surfaces/admin/components.d.ts
vividviolet: Actually, we can't delete it. This is the custom element implementation which should have the correct accessor otherwise the JS usage will be broken.

vividviolet (approved):

packages/ui-extensions/src/surfaces/admin/components.d.ts
vividviolet: Oops, this file is for docs. Definitely can delete 👍

PR #3048, opened 2025-07-09
Updating info related to target limits

liamgriffin:
### Background

Currently [the docs for extension targets](https://shopify.dev/docs/api/checkout-ui-extensions/2024-04/configuration#targets) don't inform devs that targets have limitations, leading to confusion and poor dev experience. See slack convo here: https://shopify.slack.com/archives/C01FT69928K/p1744110562588039

See associated issue here: https://github.com/shop/issues-checkout/issues/3475

### Solution

This PR adds info regarding the limitations which are currently not documented:

### 🎩

- ...

### Checklist


jamesvidler (approved):
If the intention is to have this updated on shopfiy.dev for other API versions, you'll need to make the same changes in the relevant stable release branches.

PR #3053, opened 2025-07-09
POS Targets: adds receipt header block target

alexgraffeocohen:
### Background

A partner working on a POS compliance extension requested to target the header of the printed receipt. Therefore, we're adding an extension target so that extension blocks can be added in that spot.

Closes https://github.com/shop/issues-retail/issues/15600

### Solution

We're duplicating the existing extension footer target definition, so this is not really new code.

### Checklist

### DIFF
--- a/.changeset/hip-seas-change.md
+++ b/.changeset/hip-seas-change.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Adds pos.receipt-header.block.render extension target
--- a/packages/ui-extensions/src/surfaces/point-of-sale/extension-targets.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/extension-targets.ts
@@ -187,6 +187,10 @@ export interface RenderExtensionTargets {
     {[key: string]: any} & StorageApi & TransactionCompleteWithReprintData,
     ReceiptComponents
   >;
+  'pos.receipt-header.block.render': RenderExtension<
+    {[key: string]: any} & StorageApi & TransactionCompleteWithReprintData,
+    ReceiptComponents
+  >;
 }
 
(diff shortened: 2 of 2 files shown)


vctrchu:
Can you please merge this to `2025-10-rc`? We no longer use `unstable`

packages/ui-extensions/src/surfaces/point-of-sale/targets.ts
```diff
@@ -185,6 +185,12 @@ export interface ExtensionTargets {
     {[key: string]: any} & StorageApi & TransactionCompleteWithReprintData,
     ReceiptComponents
   >;
+  'pos.receipt-header.block.render': RenderExtension<
+    // NOTE: key/any type is cause of no arg useApi() that includes all target types.
+    //   stop using useApi() with no args, instead specify the target type explicitly.
+    {[key: string]: any} & StorageApi & TransactionCompleteWithReprintData,
```
fatbattk: nit: can probably remove the repeating comment.
also do we still need the `{[key: string]: any}`?

alexgraffeocohen: Agreed on removing the comment; it's actually gone when I re-do this against the right branch anyways. On the typing question, I'm not familiar enough with how its sibling render target is used yet to know for sure. Do you view that investigation as merge blocking, or is it something I can follow up on right after?

fatbattk: yup, non-blocking.

fatbattk (approved):

alexgraffeocohen:
The same change needed to move to `packages/ui-extensions/src/surfaces/point-of-sale/extension-targets.ts` given all of the updates for Preact merged yesterday.

PR #3056, opened 2025-07-10
POS preact components and examples

vctrchu:
### Background

Migrating Point of Sale UI extensions to use preact as the default framework, moving away from React. This follows the modern pattern used in the admin surface where preact web components are used directly.

I have not tested this yet as we are still in progress of adding POS support for Remote DOM.

### Solution

These files can give a us a starting point of how components will look like moving forward. We will need to double check the proper syntax and ui-api-alignment guidelines for the official tag definitions. The names such as `pos-button`, etc are autogenerated from AI for the time being.

### 🎩

- Added .d.ts files defining preact web components with pos- prefix (e.g., <pos-button>)
- Updated 36 component example TSX files to use preact imports and web component syntax
- Enabled point-of-sale surface in loom.config.ts

### Checklist


NathanJolly (approved):
Huge. Thanks :)

PR #3064, opened 2025-07-10
Remove the unsupported section from the versions page

NathanJolly:
### Background

We stopped ending support for older versions in POS.

### Solution

Remove the section suggesting we would end support.

### 🎩

See companion PR: https://github.com/Shopify/shopify-dev/pull/60045

### Checklist


fatbattk (approved):
already mentioned re: title formatting.

PR #3073, opened 2025-07-11
[2025-10]Move analytics api from osp api to standard api in customer account ui extension.

lihaokx:
### Background
Currently, `analytics` api only exists on order status targets. We will make it available on all customer account extension targets in 2025-10 rc.
Resolves: https://github.com/shop/issues-signals/issues/2167

### what have been changed:
1. Move analytics api from osp api to standard api in customer account ui extension.
2. In analytics api, add a new field named `analytics.visitor`

### Solution

Add `analytics` api on customer account standard api, similar as checkout: https://github.com/Shopify/ui-extensions/blob/unstable/packages/ui-extensions/src/surfaces/checkout/api/standard/standard.ts#L492-L495

Checkout dev docs: https://shopify.dev/docs/api/checkout-ui-extensions/2025-07/apis/analytics

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
@@ -437,6 +437,50 @@ export interface SessionToken {
   get(): Promise<string>;
 }
 
+export interface Analytics {
+  /**
+   * Publish method to emit analytics events to [Web Pixels](https://shopify.dev/docs/apps/marketing).
+   */
+  publish(name: string, data: Record<string, unknown>): Promise<boolean>;
+
+  /**
+   * A method for capturing details about a visitor on the online store.
+   */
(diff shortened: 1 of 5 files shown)


packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
```diff
@@ -437,6 +437,50 @@ export interface SessionToken {
   get(): Promise<string>;
 }
 
+export interface Analytics {
```
lihaokx, line 440: Same as checkout: https://github.com/Shopify/ui-extensions/blob/unstable/packages/ui-extensions/src/surfaces/checkout/api/standard/standard.ts#L1485-L1527

emileber (approved):

brianshen1990 (approved):

PR #3077, opened 2025-07-14
Add toast api on standard api

oluwatimio:
### Background

Partially fixes https://github.com/shop/issues-checkout/issues/6460

### Solution

Added a toast api on standard api itself

### 🎩

Im just adding the type and nothing else for now. 

Will update docs and remove toast from ui in a follow up once implementation is complete in the host

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
@@ -394,6 +394,24 @@ export interface CompanyLocation {
 
 export interface Ui {
   /**
+   * The Toast API displays a non-disruptive message that displays at the bottom
+   * of the interface to provide quick, at-a-glance feedback on the outcome
+   * of an action.
+   *
+   * How to use:
+   *
+   * - Use toasts to confirm successful actions.
+   *
+   * - Aim for two words.
+   *
+   * - Use noun + past tense verb format. For example, \`Changes saved\`.
+   *
+   * For errors, or information that needs to persist on the page, use a [banner](/docs/api/checkout-ui-extensions/unstable/components/feedback/banner) component.
+   */
+  toast: {
+    show(content: string): void;
+  };
+  /**
    * Refresh data so the surrounding information on the page is updated. The `content` string will appear in a toast message after refresh, to confirm the action was successful.
    *
    * To request access to this API:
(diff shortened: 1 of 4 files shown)


packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
```diff
@@ -393,24 +393,6 @@ export interface CompanyLocation {
 }
 
 export interface Ui {
-  /**
```
oluwatimio: These are getting moved out into the shopify global. Ui is being removed. Since forceDataRefresh has not been done yet I just removed toast from here but not the whole interface

packages/ui-extensions/src/surfaces/customer-account/api/standard-api/standard-api.ts
```diff
@@ -101,6 +102,23 @@ export interface StandardApi<Target extends ExtensionTarget = ExtensionTarget> {
    */
   ui: Ui;
```
oluwatimio, line 103: Will remove toast type on here once we have completed moving this over + docs shipped but for now leaving it in here as we need the toast api on standard api to implement shopify.toast on the host

packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
```diff
@@ -652,3 +652,10 @@ export interface TrackingConsentChangeResultError {
    */
   message: string;
 }
+
+export interface ToastApiResult {
+  hide: () => void;
```
robin-drexler, line 657: we currently do not support `hide` https://github.com/Shopify/ui-extensions/blob/93815b417d7f33654486d5a4f9e0e5589dbea633/packages/ui-extensions/src/surfaces/point-of-sale/render/api/toast-api/toast-api.ts#L11

Are we going to implement this as part of the 2025-10 version?

oluwatimio: @robin-drexler Yes! we are adding support for it here https://github.com/shop/world/pull/106014

robin-drexler (approved):

packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
oluwatimio: Im updating this type so I can expose that change on standard api :)

lsit (approved):

andrewmcgov (approved):

PR #3080, opened 2025-07-14
Backfill editable changes for CartApi

vctrchu:
### Background

This updates 2025-07 with some changes related to the `editable` property in the POS Cart API. I've aligned with @gino about this so removing the cart error is safe as it is not used on POS yet.

Context on decision https://github.com/Shopify/ui-extensions/pull/3002#discussion_r2183499167

### Solution

### 🎩

- ...

### Checklist


packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
fatbattk, line 1: will the `2025-10` or `unstable` branch need to be updated too?

fatbattk (approved):

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
vctrchu: I'll get a PR up for 2025-10-rc. Going to leave unstable alone

PR #3082, opened 2025-07-14
Backfill editable changes to 2025-10

vctrchu:
### Background

We backfilled 2025-07, so we should also update the next release as well. https://github.com/Shopify/ui-extensions/pull/3080

### Solution

### 🎩

- ...

### Checklist


packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
fatbattk, line 1: oh, weird that it didn't have `2025.07` before.

fatbattk (approved):

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
vctrchu: 2025-10-rc was created in q4 2024 so it it quite outdated

PR #3096, opened 2025-07-18
[UI Components]: Fix missing type error

oliverigor:
### Background

- @simontaisne noticed that the TextField component missing in the types when using it on extensions, I noticed we forgot to add them in the AllowedComponents list.

### Solution

- Adding missing TextField component in the list of component types for extensions.

### 🎩

- ...

### Checklist


simontaisne (approved):
Thank you!

PR #3110, opened 2025-07-25
Update `ResourcePicker` image

billfienberg:
## TLDR

Copy correct `ResourcePicker` image from `shopify-dev`

### Background

1. The ResourcePicker image was updated directly in `shopify-dev` in https://github.com/Shopify/shopify-dev/pull/60302
1. However, the updated image isn't in `ui-extensions`, which means the `docs:sync` script is still using the old image.

### Solution

Update the `ResourcePicker` image in `ui-extensions`, so the `Extensibility docs sync` [workflow](https://github.com/Shopify/shopify-dev/actions/workflows/extensibility_docs_sync.yml) has the correct image. 

### Checklist

## References

- https://github.com/Shopify/shopify-dev/pull/60302
- https://github.com/Shopify/shopify-dev/pull/60343


themarkappleby (approved):
Thank you, LGTM

PR #3114, opened 2025-07-25
Add links for POS UI extensions developer preview

fatbattk:
### Background

Part of, https://github.com/shop/issues-retail/issues/15804
Follow up from https://github.com/Shopify/ui-extensions/pull/3113

### Solution

- Added link for all developer preview note in components/targets.

### 🎩

- ...

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/reference/components/POSReceiptBlock.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/reference/components/POSReceiptBlock.doc.ts
@@ -8,7 +8,7 @@ const data: ReferenceEntityTemplateSchema = {
   name: 'POSReceiptBlock',
   description: `A component used to group other components together for display on POS receipts.
   > Note:
-  > This is part of a POS UI Extensions developer preview. More information to come.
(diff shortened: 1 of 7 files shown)


vctrchu (approved):

packages/ui-extensions/docs/surfaces/point-of-sale/reference/components/POSReceiptBlock.doc.ts
```diff
@@ -8,7 +8,7 @@ const data: ReferenceEntityTemplateSchema = {
   name: 'POSReceiptBlock',
   description: `A component used to group other components together for display on POS receipts.
   > Note:
-  > This is part of a POS UI Extensions developer preview. More information to come.
+  > This is part of a [POS UI Extensions developer preview](/docs/api/developer-previews#pos-ui-extensions-developer-preview). More information to come.
```
aaronschubert0, line 11: I think it makes sense for this, and other repeated text, to be a static string that we import into these template strings to keep consistency and avoid repetition. Thoughts @fatbattk @NathanJolly?

fatbattk: ah, already merged this last week. yeah, can extract the whole "developer preview" block for reuse, although that needs revisiting.

PR #3115, opened 2025-07-28
Add reset function to useScannerDataSubscription hook

CalvinWalzel:
### Background

Implements https://github.com/shop/issues-retail/issues/1976

### Solution

### 🎩

- ...

### Checklist


vctrchu (commented):
Overall looks good to me, we should validate this with an extension before merging to `2025-10-rc`. Also as a follow-up to add the reset functionality to the e2e sandbox extension.

NathanJolly:
Going to close for now: This is something that, going forward, would have to go through TAG approval

PR #3120, opened 2025-07-29
fix: remove unused image reference from index pattern documentation

mpshopify:
### Background

Reference image was not properly removed from the index pattern.

### Solution

Removed the line that added the reference image on index pattern.

### 🎩

- ...

### Checklist


themarkappleby (approved):
👍

PR #3132, opened 2025-08-07
Remove Component docs for POS, update Cart API addLineItem signature

NathanJolly:
### Background

This takes care of some necessary edits to get docs building for POS again. In short, I deleted all of the component docs (as they aren't valid for remote-dom anyway) and edited the build-docs script to stop mentioning unstable. We had also missed edits to the TargetLinks object, which I've included here, too.

The actual content change here is to update the CartApi's addLineItem signature to return a Promise<string | undefined>.

### 🎩

- I committed the generated docs to https://github.com/Shopify/shopify-dev/tree/pos-ui-extensions/2025-10-rc, so you can run this locally by checking out that branch, doing `dev up`, `dev s`, and then ctrl+t to access the server.

### Checklist


katriciab (approved):

awoodall:
Haven't been able to get this running locally, I keep getting errors on `shopify-dev` but the build script is working fine.

In `ui-extensions`
* `git checkout njo/update-addLineItem-sig`
* `yarn docs:point-of-sale 2025-10`
In `shopify-dev`
* `git checkout main`
* `dev up && dev s`

image

packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
```diff
@@ -13,6 +13,21 @@ const data: LandingTemplateSchema = {
...
+      title: '2025.10',
+      sectionContent: `
+- Added in POS version: N/A
+- Release day: N/A
+
+### Important Fixes
+
+- **POS 10.10.0**:
+
+- Fixed an issue where the \`addLineItem\` function in the [Cart API](/docs/api/pos-ui-extensions/apis/cart-api#cartapi-propertydetail-addlineitem) would throw an error if the POS displayed an oversell guard modal. It will now return \`undefined\` if the user dismisses the oversell guard modal without adding anything.
```
henryStelle, line 28: This change in behaviour isn't scoped to only 2025-10, won't it also apply to all currently stable API versions?

NathanJolly: Yes, I'm actually considering it more of a patch than a breaking change, since the behavior before was to throw when an error was not really happening. We don't have a good way to document changes that backfill old versions, and the precedent is set in this file for the format I chose here.

henryStelle: But, in the previous behaviour, we never returned `undefined` - the result was always a `string` or thrown exception. With this change, an undefined might be returned to 3P code that is always expecting a string if no error is thrown and calling string methods like `result.length` will then throw a type error. It might be safer to throw a new error instead of returning undefined if you want this to behave more like a patch.

NathanJolly: Hmm, I'm inclined to agree, much as throwing an error makes no semantic sense whatsoever (no error happened, why is this function throwing)

Also reminder that the whole point of this PR is that partners were not expecting this function to throw at all, as the documentation does not suggest it does. Probably a good idea to take a pass at the docs and actually mark functions that can even throw in the first place.

I'm leaning towards my original implementation of returning an empty string if nothing was added. What do you think?

NathanJolly: @katriciab For review! Tricky stuff.

katriciab: Ah yes good catch, the `result.length` argument in older versions a good point, and they'll likely do this to check if empty. I think we can go back to the original empty string. I think we can also document the empty string case for canceling adding a sold out item (good for partners expectations, and also ours).

katriciab: I still do think undefined a better return type for this case though, but not something we need to do without being able to provide a progressive migration path which is harder without stricter API versioning.

NathanJolly: I do agree that undefined or null would be better, but could break existing extensions, to Henry's point. A bit of trouble either way! I'll go back to just `string or emptystring, or throw` which I'll also document.

NathanJolly:
The lint errors are fixed here: https://github.com/Shopify/ui-extensions/pull/3133

awoodall (approved):

PR #3139, opened 2025-08-12
Update admin-ui-extension types for admin-ui-components 1.8.0

sam-b-rose:
### Background

Updates the type for Admin UI extensions for Admin UI Components `1.8.0`

### 🎩

1. _coming soon_

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components/Menu.d.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components/Menu.d.ts
@@ -6,32 +6,18 @@
 
 // eslint-disable-next-line @typescript-eslint/triple-slash-reference, spaced-comment
 /// <reference lib="DOM" />
-import type {ThumbnailProps$1, ComponentChild} from './shared.d.ts';
+import type {
+  MenuProps$1,
+  ComponentChild,
+  InteractionProps,
+} from './shared.d.ts';
 
-export interface ThumbnailProps
-  extends Required<Pick<ThumbnailProps$1, 'src' | 'alt' | 'size'>> {
-  size: Extract<
-    ThumbnailProps$1['size'],
-    'small-200' | 'small-100' | 'small' | 'base' | 'large' | 'large-100'
-  >;
-}
-
-export type CallbackEvent<T extends keyof HTMLElementTagNameMap> = Event & {
-  currentTarget: HTMLElementTagNameMap[T];
-};
-export type CallbackEventListener<T extends keyof HTMLElementTagNameMap> =
-  | (EventListener & {
-      (event: CallbackEvent<T>): void;
-    })
-  | null;
-/** Used when an element does not have children. */
-export interface PreactBaseElementProps<TClass extends HTMLElement> {
-  /** Assigns a unique key to this element. */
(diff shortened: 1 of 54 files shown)


packages/ui-extensions/src/surfaces/admin/components/Menu.d.ts
```diff
@@ -0,0 +1,374 @@
...
+ * These symbols are used by components that implement overlay behavior
+ * (like Popover, Tooltip, etc.) to communicate with the overlay control system.
+ */
+declare const overlayCommand: unique symbol;
+declare const overlayHidden: unique symbol;
+declare const overlayActivator: unique symbol;
+
+declare class PreactOverlayElement extends PreactCustomElement {
+  constructor(renderImpl: RenderImpl);
+  [overlayHidden]: boolean;
```
sophschneider, line 95: I think we should make these symbols private so they don't affect the docs cc: @kyledurand 

Put up a PR here https://app.graphite.dev/github/pr/shop/world/136458/%5BPreactOverlayElement%5D-Make-symbols-private

kyledurand: Yeah 💯, thanks!

sophschneider: its merged if we wanna rebase and re-run the script!

packages/ui-extensions/src/surfaces/admin/components/Menu.d.ts
```diff
@@ -0,0 +1,133 @@
+/** VERSION: 1.8.0 **/
```
sam-b-rose, line 1: @kyledurand is you be so kind as to review the Menu types 🙏

packages/ui-extensions/src/surfaces/admin/components/Tooltip.d.ts
```diff
@@ -0,0 +1,130 @@
+/** VERSION: 1.8.0 **/
```
sam-b-rose, line 1: @sophschneider would you double-check tooltip types?

sophschneider: you might need to `gt sync` `admin-ui-foundations`, it looks like `[overlayHidden]: boolean;` isn't private still 🤔

kyledurand: Looks like it's [flagged as private where it needs to be](https://github.com/Shopify/ui-extensions/pull/3139/files#diff-4c0c209a565d1ae7b2ba318e6535d085dd682d66752404c6b332016f04e085afR94-R95). We can run the docs admin script and run dev locally to check in the morning

sam-b-rose: 🎩 the changes on shopify-dev locally and can verify the symbol properties are hidden for both Menu and Tooltip (no properties even exist on Tooltip)

sophschneider (approved):
Tooltip looks good!!

PR #3112, opened 2025-07-25
Add selling plan functionality to Cart API and line item types

Bruno125:
Resolves https://github.com/shop/issues-retail/issues/16077

### Background

Add functionality to Cart API and line item types to support Selling Plans on POS.

### Solution

Straightforward changes, as described in our [tech design doc](https://docs.google.com/document/d/1kLy0E9kzY52h_ll5K5DTpoBWr7o88QuaTOh8XyJ3Mzw/edit?tab=t.0#bookmark=id.pfnuj1sm7nxi).

### 🎩

None for now, will be ready for tophatting later when integrated to POS.

### Checklist

- [X] I have :tophat:'d these changes
- [X] I have updated relevant documentation

PR #3141, opened 2025-08-13
[POS]: export SellingPlan type and add extra fields

Bruno125:
Resolves https://github.com/shop/issues-retail/issues/16794

### Background

Follow up PR for: https://github.com/Shopify/ui-extensions/pull/3112, which was missing some fields / export

### Solution

Straightforward. Matches existing POS modals.

### 🎩

No extra tophat on this repo needed. Tested locally against POS, and it already connects nicely with ProductSearch API mappers.

### Checklist

### DIFF
--- a/.changeset/wicked-moons-lay.md
+++ b/.changeset/wicked-moons-lay.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+Expose additional selling plan-related fields in POS UI extensions models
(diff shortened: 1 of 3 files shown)


jiricrispeyn (approved):
LGTM. 

Was about to ask whether this should be on `ProductVariant` instead of `Product`, but compared this to the Admin GQL API which is similar to your PR. 👌 

The only field on `ProductVariant` related to selling plans is this one: https://shopify.dev/docs/api/storefront/latest/connections/SellingPlanAllocationConnection, but not sure if that is needed on the POS UI Extension types.

Bruno125:
@jiricrispeyn indeed. Additionally, `requiresSellingPlan` and `hasSellingPlans` are the only fields we have available both through local and remote search, so I would rather just keep the exposed fields limited to those two.

PR #3134, opened 2025-08-08
Added navigation to POS pages and canNavigate function to Navigation API

thunderfern:
### Background

Resolves https://github.com/shop/issues-retail/issues/1810 and https://github.com/shop/issues-retail/issues/8397. This PR is made together with [this PR](https://github.com/Shopify/pos-next-react-native/pull/64876).

### Solution

We have decided to overload the `navigate` function with our new `navigate` function. This works with the implementation seen in the `pos-next-react-native` repo in the PR linked above. (Two declarations but one implementation). We have also added a `canNavigate` function to allow partners to check if the user has permissions to access the resource and adjust their flows accordingly.

Note: Documentation and e2e tests will be updated in later PRs, referencing [this issue](https://github.com/shop/issues-retail/issues/2018) and [this issue](https://github.com/shop/issues-retail/issues/2015).

### 🎩

- Please check [this PR](https://github.com/Shopify/pos-next-react-native/pull/64876) for tophatting instructions.

### Checklist

- [x] I have :tophat:'d these changes
- [x] I have updated relevant documentation

PR #3146, opened 2025-08-14
Changed the return value of the canNavigate function to a Promise

thunderfern:
### Background

This PR is an update on [this PR](https://github.com/Shopify/ui-extensions/pull/3134).

### Solution

We need to change the update the value of the canNavigate function since we will be throwing an error if the API version is lower than 2025-10 in the function.

### 🎩

- Same thing as [this PR](https://github.com/Shopify/ui-extensions/pull/3134).

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/navigation-api/navigation-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/navigation-api/navigation-api.ts
@@ -31,9 +31,9 @@ export interface NavigationApiContent {
 
   /** Checks if the user has permission to navigate to the specified screen.
    * @param uri the uri to check if the user has permission to navigate to.
-   * @returns false if the uri is a valid uri to navigate to a POS screen and the user does not have permission to navigate to the specified screen, true otherwise.
+   * @returns false if the uri is a valid uri to navigate to a POS screen and the user does not have permission to navigate to the specified screen, true otherwise. The promise rejects if the user is not on the correct API version of POS extensions.
    */
-  canNavigate(uri: string | URL): boolean;
+  canNavigate(uri: string | URL): Promise<boolean>;
 }
 
 /**
--- a/.changeset/odd-walls-brake.md
+++ b/.changeset/odd-walls-brake.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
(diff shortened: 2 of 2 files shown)


thunderfern:
* **#3146** <a href="https://app.graphite.dev/github/pr/Shopify/ui-extensions/3146?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ui-extensions/3146?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `2025-10-rc`


<h2></h2>

katriciab (approved):

packages/ui-extensions/src/surfaces/point-of-sale/api/navigation-api/navigation-api.ts
```diff
@@ -31,9 +31,9 @@ export interface NavigationApiContent {
```
henryStelle, line 31: <p class='graphite__hidden'><i>[Re: line +7]</i></p>
All APIs are asynchronous, even if the POS implementation is synchronous as the extensibility system automatically forces them to be async due to the time is takes for the messages to cross the RPC layer. I know we have several APIs with sync signatures, but I _think_ that was an oversight. Is there any harm with converting these to async as well in 2025-10?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/Shopify/ui-extensions/3146?utm_source=unchanged-line-comment">Graphite</a>.</p>

katriciab: > I know we have several APIs with sync signatures, but I think that was an oversight.

Yeah I agree, and I did think it was odd that all APIs were not async cause they were always printing as promises. Also if the api wants to throw an error and be able to catch on the extension side by doing the following code snippet I think I remember seeing that they had to have Promise<something> :
```
try {
   await api.yourApi()
} catch (error) {
   ... only will go here if you have the await
}
```

> Is there any harm with converting these to async as well in 2025-10?

I think the harm is if we were removing Promise / returning result, cause then older extensions might have `const result = await api.yourApi()` code which they don't need to await for result anymore. But navigate returns void/no result, we should be good.

PR #3169, opened 2025-08-21
add new preact signal like type

lrsterrett:
Adds `RemoteSignalLike` interface for use in all surfaces.

Checkout extends this interface to maintain backwards compatibility with `StatefulRemoteSubscribable`

### 🎩

Tophatted locally with a symlink to local package. `.value` type shows up for expected checkout APIs

### Checklist

### DIFF
--- a/packages/ui-extensions/src/shared.ts
+++ b/packages/ui-extensions/src/shared.ts
@@ -937,3 +937,17 @@ export interface GraphQLError {
     code: string;
   };
 }
+
+/**
+ * Represents a read-only value managed on the main thread that an extension can subscribe to.
+ *
+ * Example: Checkout uses this to manage the state of an object and
+ * communicate state changes to extensions running in a sandboxed web worker.
+ *
+ * This interface is compatible with Preact's ReadonlySignal:
+ * https://github.com/preactjs/signals/blob/a023a132a81bd4ba4a0bebb8cbbeffbd8c8bbafc/packages/core/src/index.ts#L700-L709
+ */
+export interface ReadonlySignalLike<T> {
+  readonly value: T;
+  subscribe(fn: (value: T) => void): () => void;
+}
--- a/packages/ui-extensions/src/surfaces/checkout/api/shared.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/api/shared.ts
@@ -1,4 +1,4 @@
-import type {CountryCode} from '../../../shared';
+import type {CountryCode, ReadonlySignalLike} from '../../../shared';
 
 export interface ValidationError {
   /**
@@ -147,4 +147,15 @@ export interface ShippingAddress extends MailingAddress {
    * Specifies whether the address should be saved to the buyer's account.
    */
   oneTimeUse?: boolean;
+}
+
+/**
+ * A remote-subscribable object exposed to a sandboxed web worker.
+ *
+ * Example: extensions use this to get updates about an object
+ * managed by Checkout on the main thread.
+ */
+export interface StatefulRemoteSubscribable<T> extends ReadonlySignalLike<T> {
+  readonly current: T;
+  destroy(): Promise<void>;
 }
--- a/.changeset/afraid-feet-jog.md
+++ b/.changeset/afraid-feet-jog.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+add signal-like value property to checkout api
--- a/packages/ui-extensions/src/surfaces/checkout.ts
+++ b/packages/ui-extensions/src/surfaces/checkout.ts
@@ -84,6 +84,7 @@ export type {
   ValidationError,
   MailingAddress,
   ShippingAddress,
+  StatefulRemoteSubscribable,
 } from './checkout/api/shared';
 
 export type {
--- a/packages/ui-extensions/src/surfaces/checkout/api/cart-line/cart-line-item.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/api/cart-line/cart-line-item.ts
@@ -1,5 +1,4 @@
-import type {StatefulRemoteSubscribable} from '@remote-ui/async-subscription';
-
+import type {StatefulRemoteSubscribable} from '../shared';
 import type {CartLine} from '../standard/standard';
 
 export interface CartLineItemApi {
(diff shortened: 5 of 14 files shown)


packages/ui-extensions/src/surfaces/checkout/api/shared.ts
```diff
@@ -148,3 +148,34 @@ export interface ShippingAddress extends MailingAddress {
...
+}
+
+/**
+ * A remote-subscribable object exposed to a sandboxed web worker.
+ *
+ * Example: extensions use this to get updates about an object
+ * managed by Checkout on the main thread.
+ *
+ * This interface is compatible with Preact's ReadonlySignal:
+ * https://github.com/preactjs/signals/blob/a023a132a81bd4ba4a0bebb8cbbeffbd8c8bbafc/packages/core/src/index.ts#L700-L709
```
js-goupil: This looks like it's referencing a specific commit, should it be main or something like that? 
```suggestion
 * https://github.com/preactjs/signals/blob/main/packages/core/src/index.ts#L700-L709
```

packages/ui-extensions/src/surfaces/checkout/api/shared.ts
```diff
@@ -148,3 +148,34 @@ export interface ShippingAddress extends MailingAddress {
...
+ */
+export interface StatefulRemoteSubscribable<T> {
+  readonly value: T;
+  readonly current: T;
+  subscribe(subscriber: Subscriber<T>): () => void;
+  destroy(): Promise<void>;
```
js-goupil, line 160: I know this is likely more for backward compatibility, but doesn't the subscribe method return a teardown that would achieve the same thing?

js-goupil (commented):
I was under the impression though that we were moving towards the new terminology established in [this repo](https://github.com/Shopify/ui-api-design/blob/main/ui-api-design/shared/subscribable.ts). ie we'd be moving away from the `RemoteSubscribable`, `StatefulSubscribable` terminology. I don't have that strong of an opinion on the matter, but I thought we'd have to adhere to the new standard

packages/ui-extensions/src/surfaces/checkout/api/shared.ts
lrsterrett: Yes, and this is purely for backward compatibility so any extensions already started in 2025-10-rc don't break.

lrsterrett:
> I was under the impression though that we were moving towards the new terminology established in [this repo](https://github.com/Shopify/ui-api-design/blob/main/ui-api-design/shared/subscribable.ts). ie we'd be moving away from the `RemoteSubscribable`, `StatefulSubscribable` terminology. I don't have that strong of an opinion on the matter, but I thought we'd have to adhere to the new standard

@js-goupil  Yea, I kept these names as they were so that the imports of this type wouldn't have to change everywhere in checkout-web. That was before reading through that [slack thread](https://shopify.slack.com/archives/C04PUQ93D7Y/p1754406082371879) though. 

That being said, I think we're looking to land this type in checkout surface and then come back to use whatever shared type we create later. I don't think this will cause much external confusion in the meantime as they are used to seeing this type name. We just basically add the `.value` property for now. But am open to your and @kumar303 's thoughts

packages/ui-extensions/src/surfaces/checkout/api/shared.ts
lrsterrett: thinking it could change line positions on a specific branch like main, right?

kumar303: Yeah, referencing the commit is safer in case the code gets moved around. It's a pretty stable interface so it probably won't change.

packages/ui-extensions/src/surfaces/checkout/api/shared.ts
```diff
@@ -148,3 +148,34 @@ export interface ShippingAddress extends MailingAddress {
...
+/**
+ * A remote-subscribable object managed on the main thread.
+ *
+ * Example: Checkout uses this to manage the state of an object and
+ * communicate state changes to extensions running in a sandboxed web worker.
+ */
+export interface RemoteSubscribable<T> {
+  readonly initial: T;
+  subscribe(subscriber: Subscriber<T>): MaybePromise<RemoteSubscribeResult<T>>;
+}
```
kumar303: This expresses a subscribable on the main thread. It's irrelevant to the public `ui-extensions` library and I think it was a mistake to move it into the library in https://github.com/shop/world/pull/127597 (private link).  We should delete it and move it back to the checkout codebase which is the only place it's needed.

packages/ui-extensions/src/surfaces/checkout/api/shared.ts
kumar303: Checkout is the only surface that needs a backwards compatible subscribable. Let's keep it in Checkout but expose the shared type in the new aligned-on shape ([private alignment spec](https://github.com/Shopify/ui-api-design/blob/main/ui-api-design/shared/subscribable.ts)). I agree with the earlier comments that we should pick a new name 🚲🚪 I suggest a name that communicates how it's closely related to a signal:

```ts
// In src/shared.ts
export interface ReadonlySignalLike<T> {
  readonly value: T;
  subscribe(subscriber: Subscriber<T>): () => void;
}

// In src/surfaces/checkout/api/shared.ts
export interface StatefulRemoteSubscribable<T> extends ReadonlySignalLike<T> {
  readonly current: T;
  destroy(): Promise<void>;
}
```

kumar303 (commented):
Some thoughts...

packages/ui-extensions/src/shared.ts
```diff
@@ -937,3 +937,17 @@ export interface GraphQLError {
...
+/**
+ * Represents a read-only value managed on the main thread that an extension can subscribe to.
+ *
+ * Example: Checkout uses this to manage the state of an object and
+ * communicate state changes to extensions running in a sandboxed web worker.
+ *
+ * This interface is compatible with Preact's ReadonlySignal:
+ * https://github.com/preactjs/signals/blob/a023a132a81bd4ba4a0bebb8cbbeffbd8c8bbafc/packages/core/src/index.ts#L700-L709
+ */
+export interface ReadonlySignalLike<T> {
```
lrsterrett, line 950: The interface we all (surfaces) actually want 😄

js-goupil (approved):

PR #3172, opened 2025-08-21
Njo/reorganize pos targets docs

NathanJolly:
### Background

https://github.com/shop/issues-retail/issues/17212

### Solution

This adds the targets that were missing in 2025-10-rc but extant in unstable, and it also creates documentation for the receipt block header target which was missing. it adds screenshots for a lot of target locations and reorganizes our target pages to match admin's.

### 🎩

See companion PR: https://github.com/Shopify/shopify-dev/pull/61839

### Checklist


katriciab (approved):
Counting 31, all targets covered from targetDefinitions.ts 👍

PR #3174, opened 2025-08-22
Enhance Cart API for Selling Plan Support

Bruno125:
### Background

The `addLineItemSellingPlan` only required `sellingPlanId`. While that's technically enough for the operation to complete, it results in less than ideal UX given that POS has to wait for the server to sync in order to display the selected selling plan name in the cart, which can take a few seconds.

### Solution

Require the 1P/3P app to send the selling plan `id` AND `name`. That way, POS can persist both values have them at hand without having to wait for server, providing immediate feedback to user. Fetching the `name` is straightforward so it should not be complicated for apps to add this.

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/cart-api/cart-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/cart-api/cart-api.ts
@@ -7,6 +7,7 @@ import type {
   CustomSale,
   SetLineItemDiscountInput,
   SetLineItemPropertiesInput,
+  SetLineItemSellingPlanInput,
 } from '../../types/cart';
 
 /**
@@ -226,10 +227,9 @@ export interface CartApiContent {
   /**
    * Add a selling plan to a line item in the cart.
    *
-   * @param uuid the uuid of the line item that should receive the selling plan
-   * @param sellingPlanId the ID of the selling plan to add to the line item
+   * @param input required data to add a selling plan to the line item
    */
-  addLineItemSellingPlan(uuid: string, sellingPlanId: number): Promise<void>;
+  addLineItemSellingPlan(input: SetLineItemSellingPlanInput): Promise<void>;
 
   /**
    * Remove the selling plan from a line item in the cart.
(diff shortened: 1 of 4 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/api/cart-api/cart-api.ts
```diff
@@ -226,10 +227,9 @@ export interface CartApiContent {
   /**
    * Add a selling plan to a line item in the cart.
    *
-   * @param uuid the uuid of the line item that should receive the selling plan
-   * @param sellingPlanId the ID of the selling plan to add to the line item
+   * @param input required data to add a selling plan to the line item
    */
-  addLineItemSellingPlan(uuid: string, sellingPlanId: number): Promise<void>;
+  addLineItemSellingPlan(input: SetLineItemSellingPlanInput): Promise<void>;
```
Bruno125, line 232: This pattern of `SetLineItem[   ]Input` is used in a few other methods in this same file.

packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
```diff
@@ -69,14 +69,15 @@ export interface SellingPlan {
...
-  digest: string;
+  digest?: string;
   /**
-   * The interval of the selling plan. i.e. "every 2 weeks"
+   * The delivery frequency, it can be either: day, week, month or year.
    */
   deliveryInterval?: string;
   /**
-   * The interval count of the selling plan. i.e. "2"
+   * The number of intervals between deliveries.
```
Bruno125, line 80: Updating the descriptions to match the description from Checkout GQL schema

aaronschubert0 (approved):
Looks good!

Bruno125:
@jiricrispeyn merging to start testing out the integration with the updated API. Happy to revisit this PR is you have feedback when you're back

jiricrispeyn:
I am not 100% sold on expanded the extensibility methods params purely for UX needs to be completely honest.

As a partner, I'd find it odd (and prone to errors) to require both the id AND name.

Isn't there an alternative approach where we can match the `uuid` with state we have locally available on POS to get the name before the server roundtrip?

PR #3175, opened 2025-08-25
[Customer accounts] Update navigation api

oluwatimio:
### Background

Fixes https://github.com/shop/issues-checkout/issues/6646

Also review https://github.com/Shopify/shopify-dev/pull/61983

### Solution

1. Updated the types aligning with the TAG (except committed and intercept since these are not implemented yet, we will do so in a follow up)
2. Updated the global, adding navigation api to it. No more full page specific api
3. Removed comments from a lot of the full page docs
4. Removed globals we are not using as recommended by @vividviolet 

### 🎩

1. Try this ui-extension version for your extension 
2. ```"@shopify/ui-extensions": "https://registry.npmjs.org/@shopify/ui-extensions/-/ui-extensions-0.0.0-snapshot-20250826195219.tgz"```
3. Now do `navigation.` in your extension the global navigation types should be available

Screenshot 2025-08-26 at 1 59 25 PM

4. Also try the shopify global navigation type (it should not be there anymore although it still works for now)

Screenshot 2025-08-26 at 2 27 33 PM

Screenshot 2025-08-26 at 2 38 17 PM

### Checklist


packages/ui-extensions/src/surfaces/customer-account/globals.ts
vividviolet, line 1: This file is out of date - the shopify global is typed by target

packages/ui-extensions/src/surfaces/customer-account/globals.ts
```diff
@@ -1,3 +1,4 @@
+import {FullExtensionNavigation} from './api';
 import type {ExtensionTargets} from './extension-targets';
 
 export interface ShopifyGlobal {
```
vividviolet, line 4: There's should be no `reload` any more unless you're keeping it around

packages/ui-extensions/src/surfaces/customer-account/globals.ts
```diff
@@ -13,4 +14,5 @@ declare global {
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
   const shopify: ShopifyGlobal;
```
vividviolet: Let's remove this

packages/ui-extensions/buildTargetDts.ts
```diff
@@ -81,6 +81,24 @@ export type GlobalThis = typeof globalThis & {
...
 
+  if (surface === 'customer-account') {
+    template = `import type {ExtensionTargets} from '../extension-targets';
+import {ExtensionNavigation} from '../api';
+type Target = ExtensionTargets[${name}];
+export type Api = Target['api'];
+export type Output = Target['output'];
+export type Navigation = ExtensionNavigation<${name}>;
+
+export type GlobalThis = typeof globalThis & {
```
vividviolet: Can you remove this from the templates? It's not used because it doesn't work properly

packages/ui-extensions/buildTargetDts.ts
```diff
@@ -81,6 +81,24 @@ export type GlobalThis = typeof globalThis & {
...
+  shopify: Api;
+}\n
+
+declare global {
```
vividviolet: Instead of re-declaring can you try adding this at the top?
```
import('../globals');
```

oluwatimio: So I noticed that in globals.ts I dont have the target to be able to do something like `ExtensionNavigation<${name}>`. 

ExtensionNavigation is different based on the target.

packages/ui-extensions/buildTargetDts.ts
oluwatimio: Will do!

packages/ui-extensions/src/surfaces/customer-account/globals.ts
oluwatimio: Hmm unsure if we are using this, cc @robin-drexler do you know if we are using reload?

packages/ui-extensions/buildTargetDts.ts
oluwatimio: Talked to @vividviolet about this looks like guidance for globals is for them to not be target specific (e.g navigation). For functions or getters that dont apply to specific targets we could decide to not do anything on the host for those targets.

packages/ui-extensions/src/surfaces/customer-account/globals.ts
robin-drexler: we never had `reload` in the api, right? 🤔

vividviolet: I think this was a super old api from when each surface had its own package and we were using a different implementation for hot reload. It likely was accidentally copied over. For example: https://github.com/Shopify/admin-ui-extensions/blob/1072577bbc658e0fbf43156a0d066b3a275210e9/packages/admin-ui-extensions-run/src/server/hot-client.ts#L16

packages/ui-extensions/src/surfaces/customer-account/api/standard-api/standard-api.ts
```diff
@@ -236,29 +240,26 @@ export interface NavigationCurrentEntryChangeEvent {
...
    */
   from: NavigationHistoryEntry;
 }
 
-export interface StandardExtensionNavigation {
+export interface Navigation {
   /**
    * The navigate() method navigates to a specific URL, updating any provided state in the history entries list.
    */
   navigate: NavigateFunction;
```
oluwatimio, line 252: We still return => void in NavigateFuncton for customer accounts. We havent implemented commited or intercept yet, this should come later

packages/ui-extensions/docs/surfaces/customer-account/reference/apis/navigation.doc.ts
```diff
@@ -20,19 +19,8 @@ const data: ReferenceEntityTemplateSchema = {
       title: 'Navigation',
       description:
         'Navigation API for all extensions. [Refer to supported protocols](/docs/api/customer-account-ui-extensions/unstable#custom-protocols)',
-      type: 'StandardExtensionNavigation',
+      type: 'Navigation',
     },
-    {
```
oluwatimio, line 25: Its a global now, no more standard & full page navigation

packages/ui-extensions/src/surfaces/customer-account/globals.ts
```diff
@@ -9,8 +10,5 @@ export interface ShopifyGlobal {
 }
 
 declare global {
-  // conflicts with build/ts/globals.d.ts
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
-  const shopify: ShopifyGlobal;
```
oluwatimio, line 15: Was told this was not being used by @vividviolet so removed it on our end

oluwatimio: The shopify global is typed per target, this is added by the CLI not through here

packages/ui-extensions/docs/surfaces/customer-account/reference/apis/navigation.doc.ts
```diff
@@ -47,55 +29,7 @@ const data: ReferenceEntityTemplateSchema = {
...
-  //             title: 'React',
-  //           },
-  //         ],
-  //       },
-  //     },
-  //   ],
-  // },
-  related: [
-    {
-      name: 'StandardApi',
```
oluwatimio, line 93: There's no more standard api docs, this goes to 404 in production so I am just removing this from here

packages/ui-extensions/buildTargetDts.ts
```diff
@@ -72,14 +72,20 @@ function createInitialTargetDefinition({
   const directory = join(buildPath, `ts/surfaces/${surface}/targets`);
   const targetPath = join(directory, fileName);
 
-  const template = `import type {ExtensionTargets} from '../extension-targets';
+  let template = `import type {ExtensionTargets} from '../extension-targets';
```
vividviolet: Does this work? In the future I think each surface should have its own globals and we could always import it but for now we can tie it to customer-account only.

```ts
  const template = `import type {ExtensionTargets} from '../extension-targets';
type Target = ExtensionTargets[${name}];
export type Api = Target['api'];
export type Output = Target['output'];
${
  surface === 'customer-account'
    ? `import '../globals';\n`
    : ''
}
\n`;
```

oluwatimio: Let me try that! 🤞🏽 that should be much nicer

oluwatimio: Updated!

vividviolet (commented):
Looks good. Could you create a snapshot and test it out to make sure things are working as expected? Once it's published you could use the tarball url directly to avoid any local npm caches. Like this
```json
"@shopify/ui-extensions": "https://registry.npmjs.org/@shopify/ui-extensions/-/ui-extensions-0.0.0-snapshot-20250826175324.tgz"
```

oluwatimio:
> ```json
> 0.0.0-snapshot-20250826195219
> ```

Thanks @vividviolet I did not realize we can bypass the cache! It works perfectly

Screenshot 2025-08-26 at 4 03 06 PM

vividviolet (approved):

packages/ui-extensions/buildTargetDts.ts
```diff
@@ -73,13 +73,11 @@ function createInitialTargetDefinition({
   const targetPath = join(directory, fileName);
 
   const template = `import type {ExtensionTargets} from '../extension-targets';
+  ${surface === 'customer-account' ? `import '../globals';\n` : ''}
```
robin-drexler, line 76: are we the only surface with global APIs?

robin-drexler (approved):

andrewmcgov (approved):
Working in my local extension! No errors when using navigation directly: 

Screenshot 2025-08-27 at 9 29 08 AM


And now seeing the error when using `shopify.navigation`: 

Screenshot 2025-08-27 at 9 29 34 AM

packages/ui-extensions/buildTargetDts.ts
vividviolet: We could update to do a check before attempting to import the global. I think most surfaces will have globals eventually

oluwatimio: yeah we are the only ones for now, once other surfaces implement their globals they can remove this check and import `../globals` too

PR #3161, opened 2025-08-19
Updated ReturnTransactionData and ExchangeTransactionData interfaces

IP1102:
### Background

Closes https://github.com/issues/assigned?issue=shop%7Cissues-retail%7C17098

### Solution

Updated the interface fields to add optional `returnId` and `refundId`. The change has already been merged in `unstable` branch with this [PR](https://github.com/Shopify/ui-extensions/pull/3148).

### Checklist

- [ ] I have :tophat:'d these changes
- [ ] I have updated relevant documentation

PR #3148, opened 2025-08-14
Updated ReturnTransactionData and ExchangeTransactionData interfaces

IP1102:
### Background

Closes https://github.com/shop/issues-retail/issues/16805?issue=shop%7Cissues-retail%7C17025 

### Solution

Updated the interface fields to add returnId and refundId

### 🎩

- This PR is difficult to tophat standalone.
- This PR serves as a proposed approach to eventually close this [ticket](https://github.com/shop/issues-retail/issues/16805). Once the approach is finalized, we can update the POS with the changes and tophat locally.

### Checklist

- [ ] I have :tophat:'d these changes
- [ ] I have updated relevant documentation

PR #3178, opened 2025-08-26
Adds return and refund ids to ReturnTransactionData and ExchangeTransationData

alexgraffeocohen:
### Background

We need to ensure that `Return` resource ids from Shopify Admin API are included in the transaction data passed to specific extension targets used for compliance workflows. This will allow partners to properly track orders that have returns. Currently `ExchangeTransactionData` has no return id and `ReturnTransactionData` is passing a `Refund` resource ID to the `returnId` field.

Closes https://github.com/shop/issues-retail/issues/17098. Re-creating https://github.com/Shopify/ui-extensions/pull/3161 because the author left Shopify.

### Solution

Update `ExchangeTransactionData` with optional `returnId` and `ReturnTransactionData` with optional `refundId`. This change has already been merged in unstable branch in https://github.com/Shopify/ui-extensions/pull/3148. Documentation will be updated after merge.

### DIFF
--- a/.changeset/tough-panthers-impress.md
+++ b/.changeset/tough-panthers-impress.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Modified transaction data interface fields to include returnId and refundId to Return and Exchange TransactionData.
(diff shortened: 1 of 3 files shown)


caroldarski (approved):
LGTM!

PR #3182, opened 2025-08-26
Revert "Enhance Cart API for Selling Plan Support"

Bruno125:
> [!NOTE] 
> The previous PR was missing changelog so it was never released as a new version. It's safe to rollback this change.

Reverts Shopify/ui-extensions#3174

We're still discussing if name is truly required. We're able to make some improvements in POS to enhance the loading time. The change is still not immediate, but perhaps it's good enough for now. Rolling back while we're discussing further. We're still open to allowing apps to send `name` for immediate display if partners request it.

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/cart-api/add-line-item-selling-plan.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/cart-api/add-line-item-selling-plan.ts
@@ -5,11 +5,7 @@ export default extension(
   (root, api) => {
     const button = root.createComponent(Button, {
       onPress: () => {
-        api.cart.addLineItemSellingPlan({
-          lineItemUuid: api.cartLineItem.uuid,
-          sellingPlanId: 123,
-          sellingPlanName: 'My Selling Plan',
-        });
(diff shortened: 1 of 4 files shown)


janezhu918 (commented):
> We're able to make some improvements in POS to enhance the loading time. 

Can you please provide some more details on how?

Bruno125:
> Can you please provide some more details on how?

@janezhu918 we're fetching extra fields during C1 negotiation in a PR that is in review. There's still a delay, but from internal testing, not too bad. I'll be sharing videos internally to discuss with Product / UX how much of an issue it is. Note that there is a chance we bring this change back - i'm reverting temporarily until we get alignment.

janezhu918:
> > Can you please provide some more details on how?
> 
> @janezhu918 we're fetching extra fields during C1 negotiation in a PR that is in review. There's still a delay, but from internal testing, not too bad. I'll be sharing videos internally to discuss with Product / UX how much of an issue it is. Note that there is a chance we bring this change back - i'm reverting temporarily until we get alignment.

Gotcha, yes. That's what we were noticing when testing on POS as well and decided to pass the selling plan name. I think it's better to have quicker feedback on the selling plan being successfully added considering slower network for some merchants, but let's see what UX decide on.

janezhu918 (approved):

PR #3203, opened 2025-08-29
Add Preact signal-like type to Customer account

lsit:
### Background

Part of https://github.com/shop/issues-checkout/issues/7540
Related Checkout issue https://github.com/shop/issues-checkout/issues/5129
Related to https://github.com/Shopify/shopify-dev/pull/62111

### Solution

Export `StatefulRemoteSubscribable` from Customer account

| Before | After |
|-------|-------|
| Screenshot 2025-08-29 at 12 43 05 PM | Screenshot 2025-08-29 at 12 43 31 PM |
| Screenshot 2025-09-03 at 11 07 23 AM | Screenshot 2025-09-03 at 10 55 10 AM |

### 🎩

1. See https://github.com/Shopify/shopify-dev/pull/62111 for details
2. Try using the latest snapshot

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout.ts
+++ b/packages/ui-extensions/src/surfaces/checkout.ts
@@ -84,7 +84,6 @@ export type {
   ValidationError,
   MailingAddress,
   ShippingAddress,
-  StatefulRemoteSubscribable,
 } from './checkout/api/shared';
 
 export type {
--- a/packages/ui-extensions/src/surfaces/customer-account/api.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api.ts
@@ -78,6 +78,7 @@ export type {
   VisitorResult,
   VisitorSuccess,
   VisitorError,
+  StatefulRemoteSubscribable,
 } from './api/shared';
 
 export type {CartLineItemApi} from './api/cart-line/cart-line-item';
--- a/packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
@@ -1,5 +1,4 @@
 import type {ExtensionTarget} from '../extension-targets';
-import type {StatefulRemoteSubscribable} from '@remote-ui/async-subscription';
 
 import type {
   ApiVersion,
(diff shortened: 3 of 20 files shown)


packages/ui-extensions/src/surfaces/customer-account/api.ts
```diff
@@ -78,6 +78,7 @@ export type {
   VisitorResult,
   VisitorSuccess,
   VisitorError,
+  StatefulRemoteSubscribable,
```
robin-drexler, line 81: What if we just used the `ReadonlySignalLike` interface instead? Wouldn't that also work?

robin-drexler (approved):
I like it, but I think we should align this with c1 to make sure we have the same types for APIs as we're sharing so many with them.

andrewmcgov (approved):

lsit:
Changing back to `StatefulRemoteSubscribable` because we'll have more [type errors](https://github.com/shop/world/pull/164914) than we can handle right now when we go to update the dependency in `customer-account-web`

packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
```diff
@@ -23,6 +23,17 @@ export {
   StorefrontApiVersion,
 };
 
+/**
+ * A remote-subscribable object exposed to a sandboxed web worker.
+ *
+ * Example: extensions use this to get updates about an object
+ * managed by Customer account on the main thread.
+ */
+export interface StatefulRemoteSubscribable<T> extends ReadonlySignalLike<T> {
```
robin-drexler: Wonder if we should move this here and then use that in c1 and customer accounts: https://github.com/Shopify/ui-extensions/blob/2025-10-rc/packages/ui-extensions/src/shared.ts

robin-drexler (approved):

andrewmcgov (approved):

packages/ui-extensions/src/surfaces/checkout.ts
```diff
@@ -84,7 +84,6 @@ export type {
   ValidationError,
   MailingAddress,
   ShippingAddress,
-  StatefulRemoteSubscribable,
```
lsit, line 87: It'll be exported through the `export * from '../shared';` at the end of the file

PR #3209, opened 2025-09-03
update target reference for syntax and linking

NathanJolly:
### Background

Resolves https://github.com/shop/issues-retail/issues/17208

### Solution

Updates the docs files for POS UI Extensions' target usage references to be compatible with RemoteDOM
Fixes linking so that the sidebar nav is maintained when visiting related pages
Adds examples where they were missing

### 🎩

- See companion PR: https://github.com/Shopify/shopify-dev/pull/62223

### Checklist


packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/targets/pos-cart-update-event-observe.jsx
```diff
@@ -0,0 +1,9 @@
+export default async () => {
+  console.log('Cart update event received');
+  console.log('Cart:', shopify.cart);
```
katriciab, line 3: 🎩 Should I be seeing this example here?
Screenshot 2025-09-03 at 2 40 15 PM

katriciab (approved):

packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/targets/pos-cart-update-event-observe.jsx
NathanJolly: We talked about it in DM's and this is because links from the build docs auto-navigate to latest, which is still 2025-07 for now, as 2025-10 is an RC. The actual docs for 2025-10 include the examples now.

PR #3210, opened 2025-09-04
Update Customer account examples to use signals instead of hooks

lsit:
### Background

Part of https://github.com/shop/issues-checkout/issues/7487
Related to https://github.com/Shopify/shopify-dev/pull/62266

We're moving away from hooks and using signals instead. The examples in the docs should reflect that.

### Solution

- Updates all examples to support and reflect the usage of signals
- Updates all examples to use the same base code

### 🎩

See https://github.com/Shopify/shopify-dev/pull/62266 for details

### Checklist

- [x]  I have updated relevant documentation

### DIFF
--- a/packages/ui-extensions/docs/surfaces/customer-account/reference/examples/apis/analytics-publish.example.tsx
+++ b/packages/ui-extensions/docs/surfaces/customer-account/reference/examples/apis/analytics-publish.example.tsx
@@ -1,9 +1,10 @@
+import '@shopify/ui-extensions/preact';
 import {render} from 'preact';
 import {useEffect} from 'preact/hooks';
 
-export default function extension() {
+export default async () => {
   render(<Extension />, document.body);
-}
+};
 
 function Extension() {
   useEffect(() => {
--- a/packages/ui-extensions/docs/surfaces/customer-account/reference/examples/apis/attribute-values.example.tsx
+++ b/packages/ui-extensions/docs/surfaces/customer-account/reference/examples/apis/attribute-values.example.tsx
@@ -1,34 +1,16 @@
+import '@shopify/ui-extensions/preact';
 import {render} from 'preact';
-import {useEffect, useState} from 'preact/hooks';
 
 export default async () => {
-  render(<App />, document.body);
-}
+  render(<Extension />, document.body);
+};
 
-function App() {
-  const [
-    buyerSelectedFreeTShirt,
-    setBuyerSelectedFreeTShirt,
-  ] = useState(
-    shopify.attributes.current
-      ?.buyerSelectedFreeTShirt || false,
-  );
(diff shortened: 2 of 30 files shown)


packages/ui-extensions/docs/surfaces/customer-account/reference/examples/apis/analytics-publish.example.tsx
```diff
@@ -1,9 +1,10 @@
+import '@shopify/ui-extensions/preact';
```
lsit, line 1: Added this line to all examples for the same reason as checkout ([link](https://github.com/shop/world/pull/158811/files#r2322206296))

packages/ui-extensions/docs/surfaces/customer-account/reference/examples/apis/analytics-publish.example.tsx
```diff
@@ -1,9 +1,10 @@
...
 import {render} from 'preact';
 import {useEffect} from 'preact/hooks';
 
-export default function extension() {
+export default async () => {
```
lsit, line 5: Updated all examples to use async anon function since that's what we want to update our CLI template to ([issue](https://github.com/shop/issues-checkout/issues/5915))

packages/ui-extensions/docs/surfaces/customer-account/reference/examples/apis/attribute-values.example.tsx
```diff
@@ -1,34 +1,16 @@
...
-    buyerSelectedFreeTShirt,
-    setBuyerSelectedFreeTShirt,
-  ] = useState(
-    shopify.attributes.current
-      ?.buyerSelectedFreeTShirt || false,
-  );
-  const [tshirtSize, setTshirtSize] = useState(
-    shopify.attributes.current?.tshirtSize || '',
-  );
+  render(<Extension />, document.body);
```
lsit, line 5: Updated all examples to use `<Extension />` instead of `<App />`, since the CLI template currently uses `<Extension />`

packages/ui-extensions/docs/surfaces/customer-account/reference/examples/apis/settings-access.example.tsx
```diff
@@ -1,24 +1,14 @@
...
-      (updatedSettings) => {
-        setBannerTitle(
-          updatedSettings.banner_title,
-        );
-      },
-    );
-  }, []);
+function Extension() {
+  const bannerTitle =
+    shopify.settings.value?.banner_title || '';
```
lsit, line 11: I've tested this change and it works. So I assume similar changes to other examples also work.

robin-drexler (approved):
🧹

PR #3212, opened 2025-09-04
Fix target generation script for single-component targets

NathanJolly:
### Background

`pos.home.tile.render`, which only accepts a Tile component, was generating an empty target type definition

### Solution

Update the build script to consider the case where the type is a single string literal

### 🎩

- yarn build and verify that packages/ui-extensions/build/ts/surfaces/point-of-sale/targets/pos.home.tile.render.d.ts includes the import for "Tile.d.ts"

### Checklist


andy-chhuon (approved):
Looks great!

vctrchu (approved):

PR #3213, opened 2025-09-04
POS Pin Pad API

steved-shopify:
### Background

https://github.com/shop/issues-retail/issues/17432
Historically, POS has allowed extension developers to show a Pin Pad to their users by use of the `<PinPad />` component. For 2025-10 we are unifying POS extensions with Polaris Web Components. Because PinPad is a niche component used only for POS, getting alignment with the TAG team to add it to the PWC spec has proved very difficult. 

Instead, in 2025-10 we will be offering PinPad via a non-component API, exposing a `showPinPad` method that shows a modal with a PinPad when called.

### Solution

Here we add a new `PinPad` API to the POS surface which exposes a single method, `showPinPad`. The method, when called, shows a modal dialog to host the PinPad. It has two parameters:

1) A required callback for the extension to validate the PIN when it is submitted on the `PinPad`, allowing extensions to accept or reject entered PINs
2) An optional `options` parameter which allows developers to specify options for the PinPad. These options provide feature parity with the [original PinPad component](https://shopify.dev/docs/api/pos-ui-extensions/latest/components/pinpad). 

Where the old PinPad component and the new non-component API share common types, they have been moved to the `types` folder rather than being duplicated. Some common types are modified to reflect the fact we are using a modal and to more closely follow PWC standards (e.g. replacing `onPress` with `onClick`).

### 🎩

There is a [draft PR for POS](https://app.graphite.dev/github/pr/Shopify/pos-next-react-native/67134/Prototype-Extensions-Pin-Pad-API#file-areas/clients/pos-mobile/src/extensibility/api/base/usePinPadApi.ts) that implements these changes for tophatting.

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/pin-pad.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/pin-pad.ts
@@ -1,0 +1,54 @@
+export interface PinPadResult {
+  /**
+   * Whether the PIN entry was completed (not cancelled)
+   */
+  completed: boolean;
+  /**
+   * The entered PIN (only present if completed is true)
+   */
+  pin?: number[];
+}
+
+/**
+ * Represents the result of the pin pad onSubmit function.
+ * @typedef {('accept'|'reject')} PinValidationResult
+ */
+export type PinValidationResult = 'accept' | 'reject';
+
+export type PinLength = 4 | 5 | 6 | 7 | 8 | 9 | 10;
+
+export interface PinPadActionType {
+  label: string;
+  onClick: () => Promise<number[]>;
+}
+
+export interface PinPadOptions {
+  /**
+   * The function to be called when the pin pad modal is dismissed
+   */
+  onDismissed?: (result: PinPadResult) => void;
+  /**
+   * The content for the prompt on the pin pad
+   */
+  label?: string;
+  /**
+   * Whether the entered PIN should be masked
+   */
+  masked?: boolean;
+  /**
+   * The minimum length of the PIN
+   */
+  minPinLength?: PinLength;
+  /**
+   * The maximum length of the PIN
+   */
+  maxPinLength?: PinLength;
+  /**
+   * The call to action between the entry view and the keypad
+   */
+  pinPadAction?: PinPadActionType;
+  /**
+   * Title shown in the modal header
+   */
+  title?: string;
+}
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api.ts
@@ -121,3 +121,11 @@ export type {CountryCode} from './types/country-code';
 export type {Session} from './types/session';
 export type {Storage} from './types/storage';
 export {StorageError} from './types/storage';
+
+export type {PinPadApiContent, PinPadApi} from './api/pin-pad-api';
+export type {
+  PinPadOptions,
+  PinPadResult,
+  PinLength,
+  PinPadActionType,
+} from './types/pin-pad';
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/pin-pad-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/pin-pad-api.ts
@@ -1,0 +1,21 @@
+import {PinPadOptions, PinValidationResult} from '../types/pin-pad';
+
+export interface PinPadApiContent {
+  /** Shows a pin pad to the user in a modal dialog.
+   *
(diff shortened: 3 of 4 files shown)


steved-shopify:
* **#3213** <a href="https://app.graphite.dev/github/pr/Shopify/ui-extensions/3213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/ui-extensions/3213?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `2025-10-rc`


<h2></h2>

tsdevshop (approved):
Steve and I paired on this, generally I think this looks good.

I had a query about whether we should split the presentation of the PIN pad from the submission of the PIN for verification, similar to the way that 2025-07 splits CameraScanner from the Scanner API. This would allow us to (potentially) support other mechanisms for PIN entry, e.g. hardware badges, external PIN pads, etc.

But on reflection, this might be too much YAGNI, so for now - this looks good.

packages/ui-extensions/src/surfaces/point-of-sale/api/action-target-api/action-target-api.ts
```diff
@@ -1,9 +1,11 @@
...
 import {StandardApi} from '../standard/standard-api';
+import {PinPadApi} from '../pin-pad-api';
 
 export type ActionTargetApi<T> = {[key: string]: any} & {
   extensionPoint: T;
 } & StandardApi<T> &
   ScannerApi &
-  NavigationApi;
+  NavigationApi &
+  PinPadApi;
```
NathanJolly: If the API is supposed to open a modal with a PinPad, wouldn't it be appropriate to add it to the StandardApi since then you'd be able to call it directly from a Tile or Menu item instead of having to go Tile -> Modal -> pinPad? There's really no reason for that intermediate Modal in most use cases since PinPad is no longer a component that goes on a screen.

tsdevshop:
Something else to consider - we should ensure that the new PinPad API is added to the documentation, and as part of the migration guide.

packages/ui-extensions/src/surfaces/point-of-sale/api/action-target-api/action-target-api.ts
steved-shopify: Thanks @NathanJolly - absolutely right. Updated. 👍

NathanJolly (approved):

steved-shopify:
### Merge activity

* **Sep 5, 3:57 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/ui-extensions/3213).
* **Sep 5, 3:58 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/Shopify/ui-extensions/3213) rebased this pull request as part of a merge.
* **Sep 5, 3:59 PM UTC**: @steved-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/ui-extensions/3213).

packages/ui-extensions/src/surfaces/point-of-sale/types/pin-pad.ts
```diff
@@ -0,0 +1,54 @@
+export interface PinPadResult {
+  /**
+   * Whether the PIN entry was completed (not cancelled)
+   */
+  completed: boolean;
+  /**
+   * The entered PIN (only present if completed is true)
+   */
+  pin?: number[];
```
aaronschubert0, line 9: @tsdevshop @steved-shopify does the current PinPad expose the pin that's entered? Would this not mean that an extension can harvest pins, which would be a security breach?

tsdevshop: Yes, this is correct - although this is also an issue in the legacy `<PinPad>` component from 2025-07 and below, which also exposes an 

Perhaps this is a separate discussion we need to have. Are we comfortable with this concern? Or should we hold off on adding support for PinPad for now?

cc/ @awoodall

tsdevshop (commented):
Looking again at the 2025-07 implementation, this exposes an `onPinEntry` callback, fired whenever a PIN digit is pressed. Should we consider adding the same here?

packages/ui-extensions/src/surfaces/point-of-sale/types/pin-pad.ts
awoodall: Yeah this has been a concern and one of the main concerns the TAG review folks had with PinPad.

Long-term we should drop this but let's wait for that convo between @runmad and easyteam. Some insight from them might help tip the scales on way or another.

PR #3214, opened 2025-09-04
Update badge docs

cappslock:
### Background

Partial resolution of https://github.com/shop/issues-retail/issues/17897

### Solution

Updates Badge docs for PWC. This is split into two commits, one which updates the generated types from pos-next-react-native (for several components), one which modifies the Badge docs for accuracy.

### 🎩

Mainly I'd like eyes on the example. screenshots - I was not able to get a higher resolution screenshot from the device

### Checklist


awoodall (commented):
Generally LGTM but we may want Patrick to get us some assets as these do look a tad bit blurry. on my high resolution display.

Image

PR #3252, opened 2025-09-05
Migrate Customer account target examples to Preact

lsit:
### Background

Part of https://github.com/shop/issues-checkout/issues/7668
https://github.com/Shopify/shopify-dev/pull/62347

The target examples hadn't been migrated to Preact and were commented out in the docs

### Solution

Migrate them to Preact and uncomment

### 🎩

See https://github.com/Shopify/shopify-dev/pull/62347 for details

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/customer-account/reference/examples/targets/customer-account.footer.render-after/default.example.tsx
+++ b/packages/ui-extensions/docs/surfaces/customer-account/reference/examples/targets/customer-account.footer.render-after/default.example.tsx
@@ -1,24 +1,16 @@
-import {
-  BlockStack,
-  reactExtension,
-  Text,
-  useApi,
-} from '@shopify/ui-extensions-react/customer-account';
+import '@shopify/ui-extensions/preact';
+import {render} from 'preact';
 
-// 1. Choose an extension target
-export default reactExtension(
-  'customer-account.footer.render-after',
-  () => <Extension />,
-);
+export default async () => {
+  render(<Extension />, document.body);
+};
 
 function Extension() {
-  // 2. Use the extension API to gather context from extension
-  const {extension} = useApi();
-
-  // 3. Render a UI
   return (
-    <BlockStack>
-      <Text>target: {extension.target}</Text>
-    </BlockStack>
+    <s-stack direction="block">
+      <s-text>
+        target: {shopify.extension.target}
+      </s-text>
+    </s-stack>
   );
 }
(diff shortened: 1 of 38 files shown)


packages/ui-extensions/docs/surfaces/customer-account/reference/examples/targets/customer-account.footer.render-after/default.example.tsx
```diff
@@ -1,24 +1,16 @@
-import {
-  BlockStack,
-  reactExtension,
-  Text,
-  useApi,
-} from '@shopify/ui-extensions-react/customer-account';
+import '@shopify/ui-extensions/preact';
```
brianshen1990, line 1: do we have to import from `import '@shopify/ui-extensions/preact';` ? I think the `shopify.d.ts` already included it by default right?

brianshen1990 (approved):
left a comment related to `import '@shopify/ui-extensions/preact';` , otherwise the PR looks good~

packages/ui-extensions/docs/surfaces/customer-account/reference/examples/targets/customer-account.footer.render-after/default.example.tsx
lsit: That's the side effect import to which turns our subscribables into signal-like objects so that they work with Preact signals. I've added it to all examples so that if someone starts with any example and then starts using calls to `.value`, they'll just work

andrewmcgov (approved):
~I think @brianshen1990 is probably right we should not need the `@shopify/ui-extensions/preact'` imports, but aside from that~ it looks good!

packages/ui-extensions/docs/surfaces/customer-account/reference/examples/targets/customer-account.footer.render-after/default.example.tsx
lsit: @andrewmcgov ^ fyi

andrewmcgov: Understood, thanks @lsit !

brianshen1990: interesting, I feel like developers may miss this import easily...

lsit: Yes, I think we'll have to update the docs to make it clearer. I'm just not sure where yet

PR #3260, opened 2025-09-08
Adds DiscountApplication and DiscountAllocation to correct interfaces

alexgraffeocohen:
### Background

Partners need access to discount application data at the order level and discount allocation data at the line item level within the pos.receipt-footer.block.render extension target. This will allow them to correctly calculate taxes for all line items.

This PR makes the correct updates to the relevant interfaces. That includes the line item interface shared with the cart API.

Closes https://github.com/shop/issues-retail/issues/18108
Closes https://github.com/shop/issues-retail/issues/18109

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/discount-allocation.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/discount-allocation.ts
@@ -1,0 +1,12 @@
+import type {MoneyV2} from './money-v2';
+import type {DiscountApplication} from './discount-application';
+
+export interface DiscountAllocation {
+  allocatedAmountSet: MoneyBag;
+  discountApplication: DiscountApplication;
+}
+
+export interface MoneyBag {
+  shopMoney: MoneyV2;
+  presentmentMoney: MoneyV2;
+}
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/discount-application.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/discount-application.ts
@@ -1,0 +1,59 @@
+import type {MoneyV2} from './money-v2';
+
+export enum DiscountApplicationAllocationMethod {
+  ACROSS = 'ACROSS',
+  EACH = 'EACH',
+  ONE = 'ONE',
+}
+
+export enum DiscountApplicationTargetSelection {
+  ALL = 'ALL',
+  ENTITLED = 'ENTITLED',
+  EXPLICIT = 'EXPLICIT',
+}
+
+export enum DiscountApplicationTargetType {
+  LINE_ITEM = 'LINE_ITEM',
+  SHIPPING_LINE = 'SHIPPING_LINE',
+}
+
+export type DiscountApplication =
+  | AutomaticDiscountApplication
+  | DiscountCodeApplication
+  | ManualDiscountApplication
+  | ScriptDiscountApplication;
+
+export interface DiscountApplicationBase {
+  allocationMethod: DiscountApplicationAllocationMethod;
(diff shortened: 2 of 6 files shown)


alexgraffeocohen:
One consideration is whether we should have a separate line item interface for transaction complete data, if we don't want to add to all cart APIs right now.

packages/ui-extensions/src/surfaces/point-of-sale/types/discount-allocation.ts
```diff
@@ -0,0 +1,12 @@
+import type {Money} from './money';
+import type {DiscountApplication} from './discount-application';
+
+export interface DiscountAllocation {
+  allocatedAmountSet: MoneyBag;
+  discountApplication: DiscountApplication;
+}
+
+export interface MoneyBag {
+  shopMoney: Money;
```
chadcromwell: Just double checking, online docs show MoneyV2. Just want to make sure Money references the latest version (assuming it does).

Edit: I see we added MoneyV2 below. How does this conflict?

packages/ui-extensions/src/surfaces/point-of-sale/types/discount-application.ts
```diff
@@ -0,0 +1,45 @@
...
+  | DiscountCodeApplication
+  | ManualDiscountApplication
+  | ScriptDiscountApplication;
+
+export interface DiscountApplicationBase {
+  allocationMethod: 'ACROSS' | 'EACH' | 'ONE';
+  index: number;
+  targetSelection: 'ALL' | 'ENTITLED' | 'EXPLICIT';
+  targetType: 'LINE_ITEM' | 'SHIPPING_LINE';
+  value: PricingValue;
```
chadcromwell: Should we define some of these as types since we're in here?

allocationMethod: [DiscountApplicationAllocationMethod](https://shopify.dev/docs/api/admin-graphql/latest/enums/DiscountApplicationAllocationMethod)
targetSelection: [DiscountApplicationTargetSelection](https://shopify.dev/docs/api/admin-graphql/latest/enums/DiscountApplicationTargetSelection)
targetType: [DiscountApplicationTargetType](https://shopify.dev/docs/api/admin-graphql/latest/enums/DiscountApplicationTargetType)

chadcromwell (commented):
Thanks for this! I added a few comments/questions. I don't want to block as I'm going to be busy all week at Startup, but wanted to take a peek and help keep things moving along.

packages/ui-extensions/src/surfaces/point-of-sale/types/discount-application.ts
alexgraffeocohen: Probably!

packages/ui-extensions/src/surfaces/point-of-sale/types/discount-allocation.ts
alexgraffeocohen: Hm interesting, I will investigate. I'm hoping there's no need to actually define `MoneyV2`.

alexgraffeocohen: Money is

`export interface Money {
  amount: number;
  currency: string;
}
`

But `MoneyV2` has `currencyCode` instead, which incidentally is what necessary here as you say. This is a long running migration across Shopify that is pretty annoying. We should still introduce `MoneyV2` and use it where needed/possible instead of `Money`.

alexgraffeocohen:
Closing in favor of https://github.com/Shopify/ui-extensions/pull/3283 and a second PR to introduce `DiscountApplication` with a more restricted attribute set.

PR #3261, opened 2025-09-09
Enhance Cart API for Selling Plan Support

Bruno125:
### Background

The `addLineItemSellingPlan` only required `sellingPlanId`. While that's technically enough for the operation to complete, it results in less than ideal UX given that POS has to wait for the server to sync in order to display the selected selling plan name in the cart, which can take a few seconds.

This change was previously merged: https://github.com/Shopify/ui-extensions/pull/3174 and reverted: https://github.com/Shopify/ui-extensions/pull/3182 as part of our design phase. Ultimately, we decided it will enable better extensibility and better UX, so we're bringing it back.

### Solution

Require the 1P/3P app to send the selling plan `id` AND `name`. That way, POS can persist both values have them at hand without having to wait for server, providing immediate feedback to user. Fetching the `name` is straightforward so it should not be complicated for apps to add this.

### Checklist

### DIFF
--- a/.changeset/clean-keys-float.md
+++ b/.changeset/clean-keys-float.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Update POS Cart API selling plan method to support more fields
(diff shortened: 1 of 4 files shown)


janezhu918 (approved):
Code changes LGTM.

coderBackPacker (approved):
change LGTM

aaronschubert0 (approved):
I know it's already been merged but approving for completeness sake since I approved the original PR.

PR #3272, opened 2025-09-10
Remove force data refresh and UI typings

oluwatimio:
### Background

Looks like only the docs were removed. Also removing the typings

### Solution

Remove typings

### 🎩

To try this out

```
"@shopify/ui-extensions": "https://registry.npmjs.org/@shopify/ui-extensions/-/ui-extensions-0.0.0-snapshot-20250910160331.tgz"
```

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
@@ -393,23 +393,6 @@ export interface CompanyLocation {
   id: string;
 }
 
-export interface Ui {
-  /**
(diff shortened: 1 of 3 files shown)


packages/ui-extensions/src/surfaces/customer-account/api/shared.ts
```diff
@@ -393,23 +393,6 @@ export interface CompanyLocation {
   id: string;
 }
 
-export interface Ui {
```
oluwatimio, line 396: Looks like only docs were removed. Removing typings too from 2025-10-rc

lsit (approved):

andrewmcgov (approved):

PR #3281, opened 2025-09-11
Add Admin Modal docs and improve preview size

sam-b-rose:
### Modal docs

Screenshot 2025-09-12 at 10 26 55 AM

### Custom preview sizes

https://github.com/user-attachments/assets/8d30bd12-c421-4d03-92bb-ee8cd3a1f04f


packages/ui-extensions/src/surfaces/admin/components/Modal/Modal.doc.ts
```diff
@@ -0,0 +1,40 @@
...
+    },
+  ],
+  defaultExample: {
+    codeblock: {
+      title: 'Code',
+      tabs: [
+        {
+          code: './examples/default.html',
+          language: 'preview',
+          layout: 'alignStart',
```
kyledurand: I think default layout would be better here

packages/ui-extensions/src/surfaces/admin/components/Modal/examples/default.html
```diff
@@ -0,0 +1,15 @@
+<s-button commandFor="modal">Open</s-button>
```
kyledurand: Let's fix this example up. Pare down the content so it can all fit in the window and let's not mention Instagram :)

kyledurand (commented):
This is looking good to me. We need to fix some other things in docs to day so hold off on shipping this until it's in the cdn then we can prioritize getting this into docs

packages/ui-extensions/src/surfaces/admin/components/Modal/Modal.doc.ts
```diff
@@ -0,0 +1,39 @@
+import {AdminReferenceEntityTemplateSchema} from '../../docs-types';
```
kyledurand: Let's rename this to AB.doc.ts or whatever the naming convention is until we know for sure if this is going to ui-extensions

packages/ui-extensions/src/surfaces/admin/components/StandardComponents.ts
```diff
@@ -25,6 +25,7 @@ export type StandardComponents =
   | 'Link'
   | 'ListItem'
   | 'Menu'
+  | 'Modal'
```
kyledurand: We don't want this in StandardComponents I don't think

packages/ui-extensions/src/surfaces/admin/components/Modal/Modal.ab.doc.ts
```diff
@@ -0,0 +1,40 @@
...
+    },
+  ],
+  defaultExample: {
+    codeblock: {
+      title: 'Code',
+      tabs: [
+        {
+          code: './examples/default.html',
+          language: 'preview',
+          customStyles: 'min-height: 300px;',
```
kyledurand: How do you pass in multiple styles here? Should this be an array?

packages/ui-extensions/src/surfaces/admin/components/StandardComponents.ts
davejcameron: Yeah, we want to remove it from StandardComponents.

packages/ui-extensions/src/surfaces/admin/components/Modal/Modal.ab.doc.ts
sam-b-rose: It is just a string. Inline styles like you'd add to a style attribute (e.g. `style="color: red; background: blue;"`)

sam-b-rose: I've refined this to support a CSS Style Object. Types included

packages/ui-extensions/src/surfaces/admin/components/Modal/Modal.ab.doc.ts
```diff
@@ -0,0 +1,42 @@
...
+  ],
+  defaultExample: {
+    codeblock: {
+      title: 'Code',
+      tabs: [
+        {
+          code: './examples/default.html',
+          language: 'preview',
+          customStyles: {
+            minHeight: '300px',
```
kyledurand, line 34: Slightly worried about the can of worms this opens up. Will we end up with a bunch of different sized examples? I'm kinda leaning toward removing customStyles and adding another template

kyledurand (approved):

packages/ui-extensions/src/surfaces/admin/components/Modal/Modal.ab.doc.ts
sam-b-rose: Limited the custom style values to `minHeight`. This makes this more controlled but also extendible for specific cases.

packages/ui-extensions/src/surfaces/admin/components/StandardComponents.ts
sam-b-rose: Removed 💥

PR #3283, opened 2025-09-11
Adds DiscountAllocation to CartLineItem API

alexgraffeocohen:
### Background

Added optional `discountAllocations` field to Cart Line Item API so that POS extensions can see the precise proportion of a discount applied to a particular line item to accurately calculate a line item's total amount. Only the `allocatedAmount` field is included for this purpose. See https://shopify.dev/docs/api/storefront/latest/objects/DiscountAllocation.

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/discount-allocation.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/discount-allocation.ts
@@ -1,0 +1,5 @@
+import type {MoneyV2} from './money';
+
+export interface DiscountAllocation {
+  allocatedAmount: MoneyV2;
+}
--- a/.changeset/discount-allocations.md
+++ b/.changeset/discount-allocations.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Added optional `discountAllocations` field to Cart Line Item API so that POS extensions can see the precise proportion of a discount applied to a particular line item. Only the `allocatedAmount` field is included for this purpose, matching the Storefront API structure. See https://shopify.dev/docs/api/storefront/latest/objects/DiscountAllocation.
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
@@ -1,5 +1,6 @@
 import {CountryCode} from './country-code';
 import {TaxLine} from './tax-line';
+import {DiscountAllocation} from './discount-allocation';
 
 export interface Cart {
   /**
@@ -40,6 +41,7 @@ export interface LineItem {
   variantId?: number;
   productId?: number;
   discounts: Discount[];
+  discountAllocations?: DiscountAllocation[];
   taxable: boolean;
   taxLines: TaxLine[];
   sku?: string;
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/money.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/money.ts
@@ -2,3 +2,8 @@ export interface Money {
   amount: number;
   currency: string;
 }
+
+export interface MoneyV2 {
+  amount: string;
+  currencyCode: string;
+}


packages/ui-extensions/src/surfaces/point-of-sale/types/money-v2.ts
fatbattk, line 1: should this be combined with [money.ts](https://github.com/Shopify/ui-extensions/blob/unstable/packages/ui-extensions/src/surfaces/point-of-sale/types/money.ts)?

packages/ui-extensions/src/surfaces/point-of-sale/types/discount-allocation.ts
```diff
@@ -0,0 +1,10 @@
+import type {MoneyV2} from './money-v2';
+
+export interface DiscountAllocation {
+  allocatedAmountSet: MoneyBag;
+}
+
+export interface MoneyBag {
+  shopMoney: MoneyV2;
+  presentmentMoney: MoneyV2;
+}
```
fatbattk: wdyt about combining this with `Money` type as well.

packages/ui-extensions/src/surfaces/point-of-sale/types/money-v2.ts
alexgraffeocohen: As far as I understand, the `Money` and `MoneyV2` types are intentionally different things across shopify, where `MoneyV2` has the `currencyCode`. I wasn't sure if we could mess with `Money` as a result, as in maybe other APIs depend on the old version of money. But if not, that would def be better.

packages/ui-extensions/src/surfaces/point-of-sale/types/money-v2.ts
```diff
@@ -0,0 +1,4 @@
+export interface MoneyV2 {
+  amount: string;
+  currencyCode: string;
+}
```
jolenehayes: Is this type not available in this code base somewhere already?

jolenehayes: I was looking at the Money interface in [standard.ts](https://github.com/Shopify/ui-extensions/blob/2025-10-rc/packages/ui-extensions/src/surfaces/checkout/api/standard/standard.ts#L1056) and it has the currency code as a part of it. Is there a reason we need a separate type?

jolenehayes: If we want to keep this should probably change CurrencyCode to reflect a similar type:
```
export interface MoneyV2 {
  amount: number;
  currencyCode: CurrencyCode;
}
```

alexgraffeocohen: Good find! I don't think there is a meaningful difference to have `MoneyV2` declared specifically; it feels more important to just have the attributes correct. So I like the idea of using the standard API here. My only hesitation is it being confusing to use it intermingled with the pre-existing `point-of-sale` surface `Money`.

packages/ui-extensions/src/surfaces/point-of-sale/types/discount-allocation.ts
jolenehayes: Is this a type that other things in the api use? Might make sense to move it to "standard.ts" as well.

alexgraffeocohen: Ah this should go away sorry, one sec.

fatbattk (approved):

packages/ui-extensions/src/surfaces/point-of-sale/types/money-v2.ts
alexgraffeocohen: Ah actually, let's follow up on this in a different way: to see if we can maybe create a centralized type for Money that multiple surfaces could use, rather than importing across surface.

fatbattk: it's probably better to separate typings between surfaces.

PR #3291, opened 2025-09-12
Add a link to the useNavigate docs about 'shopify:navigate'

jas7457:
Fixes https://github.com/shop/issues-polaris/issues/1183

image

### Background

The Polaris `s-link` component will prevent links from working while inside of an embedded app due to [this code](https://github.com/shop/world/blob/main/libraries/javascript/polaris/admin-ui-components/lib/componentBundleGenerationPlugin.js#L66). It looks like we are firing off a `shopify:navigate` event instead.

### Solution

This is not very clear/discoverable in the place devs would likely look when their links aren't navigating anywhere. This simply provides a small note and a link to where we discuss it further in our [current docs](https://shopify.dev/docs/api/app-home?accordionItem=getting-started-existing-remix-application).

[Video](https://screenshot.click/12-49-1rg0c-13l10.mp4)

### 🎩

- Ensure that the link works

### Checklist

### DIFF
--- a/packages/ui-extensions/src/docs/shared/components/Link.ts
+++ b/packages/ui-extensions/src/docs/shared/components/Link.ts
@@ -3,7 +3,7 @@ import type {SharedReferenceEntityTemplateSchema} from '../docs-type';
 const data: SharedReferenceEntityTemplateSchema = {
   name: 'Link',
   description:
-    'Makes text interactive, allowing users to navigate to other pages or perform specific actions. Supports standard URLs, custom protocols, and navigation within Shopify or app pages.',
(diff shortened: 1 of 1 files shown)


kyledurand (approved):

jas7457:
I've added [automatic handling](https://github.com/Shopify/shopify-app-js/pull/2794) in our Remix template. This should mainly negate the need for doc updates.

I'm going to close this PR. If we find that people are struggling with this (meaning they didn't update their `@shopify/shopify-app-remix`, we can revisit updating the s-link docs to point them to the solution.

PR #3296, opened 2025-09-13
Fix popover example

davejcameron:
### Background

`accessibilityLabel` isn't available on Popover, direction is `block` and no gap or padding should be set.

### Solution

### 🎩

- ...

### Checklist


kyledurand (commented):
> accessibilityLabel isn't available on menu.

It is a valid prop:

https://github.com/shop/world/blob/remove-children-from-jsx-types/libraries/javascript/polaris/admin-ui-components/src/components/Menu/shadowRoot.tsx#L113

Are you maybe seeing a types issue?

kyledurand (approved):
Looks good, thanks for fixing this

PR #3297, opened 2025-09-15
Simplify selling plans APIs for POS

Bruno125:
### Solution

The final API for selling plans for POS will look like this:
- `lineItemUuid`: **required**, to know to which line item to apply the change
- `sellingPlanId`: **required**, to know which plan to apply
- `sellingPlanName`: **optional**, to display the name in POS right away. If not provided, POS will try to get it from the server

As part of this PR, we're removing `sellingPlanDeliveryInterval` and `sellingPlanDeliveryIntervalCount` from the input, to simplify the API. POS will fetch those from the server instead.

### Checklist


jiricrispeyn (approved):
LGTM! 👌 

Let's make sure we also ship the updated public dev docs and the POS changes to these methods as soon as possible.

coderBackPacker (approved):
Changes LGTM

PR #3313, opened 2025-09-16
[chore]: Bump generate docs library to 0.19.3

oliverigor:
### Background

- Bumping generate-docs library to fix several issues when generating docs.

### Solution

- bumped the library
- removed duplicate of the library from the root package
  - this was causing a weird "race condition" when generating docs locally causing the APIs/Components to not be parsed correctly.

### 🎩

- `yarn docs:checkout 2025-10-rc` or your surface :D

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "@shopify/babel-preset": "^24.1.0",
     "@shopify/browserslist-config": "^3.0.0",
     "@shopify/eslint-plugin": "^42.0.0",
-    "@shopify/generate-docs": "0.17.1",
     "@shopify/loom": "^1.0.0",
     "@shopify/loom-cli": "^1.0.0",
     "@shopify/loom-plugin-build-library": "^1.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -39,7 +39,6 @@
     "@shopify/babel-preset": "^24.1.0",
     "@shopify/browserslist-config": "^3.0.0",
     "@shopify/eslint-plugin": "^42.0.0",
-    "@shopify/generate-docs": "0.17.1",
```
oliverigor, line 42: I wonder if we need this to generate the docs through the github action ??

cc: @jamesvidler

yarn.lock
```diff
@@ -1913,10 +1913,10 @@
     pkg-dir "^5.0.0"
     pluralize "^8.0.0"
 
-"@shopify/generate-docs@0.17.1":
-  version "0.17.1"
-  resolved "https://registry.npmjs.org/@shopify/generate-docs/-/generate-docs-0.17.1.tgz"
```
oliverigor, line 1918: hum ... not good, we should still be getting this through the registry, will check

oliverigor:
closing in favor of https://github.com/Shopify/ui-extensions/pull/3323

PR #3314, opened 2025-09-17
Update signal-like type in Customer account to have deprecated fields

lsit:
### Background

Part of https://github.com/shop/issues-checkout/issues/7883
Related to https://github.com/Shopify/shopify-dev/pull/62708

We want to keep the fields that should no longer be used on the signal-like type and just mark them as deprecated so that we can remove them later.

### Solution

Update the signal-like type to have deprecated fields

### 🎩

See https://github.com/Shopify/shopify-dev/pull/62708 for details

### Checklist

### DIFF
--- a/packages/ui-extensions/src/shared.ts
+++ b/packages/ui-extensions/src/shared.ts
@@ -941,11 +941,10 @@ export interface GraphQLError {
 /**
  * Represents a read-only value managed on the main thread that an extension can subscribe to.
  *
- * Example: Checkout uses this to manage the state of an object and
+ * Example: extension surfaces use this to manage the state of an object and
  * communicate state changes to extensions running in a sandboxed web worker.
  *
- * This interface is compatible with Preact's ReadonlySignal:
- * https://github.com/preactjs/signals/blob/a023a132a81bd4ba4a0bebb8cbbeffbd8c8bbafc/packages/core/src/index.ts#L700-L709
+ * This interface is compatible with [Preact's ReadonlySigna](https://github.com/preactjs/signals/blob/a023a132a81bd4ba4a0bebb8cbbeffbd8c8bbafc/packages/core/src/index.ts#L700-L709).
  */
 export interface ReadonlySignalLike<T> {
   readonly value: T;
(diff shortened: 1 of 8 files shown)


packages/ui-extensions/src/shared.ts
```diff
@@ -962,3 +962,26 @@ export interface StatefulRemoteSubscribable<T> extends ReadonlySignalLike<T> {
...
+ * Example: Checkout uses this to manage the state of an object and
+ * communicate state changes to extensions running in a sandboxed web worker.
+ *
+ * This interface is compatible with [Preact's ReadonlySignal](https://github.com/preactjs/signals/blob/a023a132a81bd4ba4a0bebb8cbbeffbd8c8bbafc/packages/core/src/index.ts#L700-L709).
+ *
+ * Some fields are deprecated but still supported for backwards compatibility.
+ * In version 2025-10, [`StatefulRemoteSubscribable`](https://github.com/Shopify/remote-dom/blob/03929aa8418a89d41d294005f219837582718df8/packages/async-subscription/src/types.ts#L17) was replaced with `ReadonlySignalLike`. The old fields will be removed some time in the future.
+ *
+ */
+export interface SubscribableSignalLike<T> extends ReadonlySignalLike<T> {
```
lsit, line 976: Copied from Checkout [here](https://github.com/shop/world/blob/main/areas/clients/checkout-web/packages/ui-extensions/src/surfaces/checkout/shared.ts#L92)

packages/ui-extensions/src/shared.ts
```diff
@@ -962,3 +962,26 @@ export interface StatefulRemoteSubscribable<T> extends ReadonlySignalLike<T> {
   readonly current: T;
   destroy(): Promise<void>;
 }
+
+/**
+ * Represents a read-only value managed on the main thread that an extension can subscribe to.
+ *
+ * Example: Checkout uses this to manage the state of an object and
```
robin-drexler: hmmm should we keep checkout here now that it's shared?

robin-drexler (approved):

packages/ui-extensions/src/shared.ts
lsit: Is it reasonable to say "Surfaces use this to manage the state of an object..."?

andrewmcgov (approved):

packages/ui-extensions/src/shared.ts
robin-drexler: extension surface?

PR #3317, opened 2025-09-17
Docs: Updated discounts API docs to reflect case-insensitive discount codes (2025-07)

skylarrrji:
### Background
**Closes:** https://github.com/shop/issues-checkout/issues/597
**Issue:** The documentation for the checkout UI extensions discounts API states that the discount code types for add/remove are case-sensitive for the 2025-07 version as shown in the image below, when they are actually case-insensitive as associated with https://github.com/shop/issues-checkout/issues/597

17-41-m4p2z-vriwx

### Solution
Updated documentation to state discount code types for add/remove are case-insensitive for the 2025-07 version as shown below:

17-36-0kq5r-ynesi

### Checklist


dnagoda (approved):

lrsterrett (approved):
👍

PR #3323, opened 2025-09-18
[chore]: Bump generate docs library to 0.19.3

oliverigor:
### Background

- Bumping generate-docs library to fix several issues when generating docs.

### Solution

- bumped the library
- removed duplicate of the library from the root package
  - this was causing a weird "race condition" when generating docs locally causing the APIs/Components to not be parsed correctly.

### 🎩

- `yarn docs:checkout 2025-10-rc` or your surface :D

### Checklist


kumar303 (approved):
We weren't sure about using the `yarnpkg` registry but it seems fine

PR #3326, opened 2025-09-18
Align POS Navigation in Global

js-goupil:
### Background
- This PR aligns POS' navigation system to be on the global, using the aligned interfaces.

Note that we are still missing `updateCurrentEntry` as it's not a requirement currently for us, and we needed to cut some scope. However, we do have the `goBack()` function as a requirement for our surface, which does not currently exist in customer-account's implementation.

### Solution
While most of the Navigation type is aligned, there are some small differences. As such, we have our own globals.ts file. Because we also provide the navigation type, I had to add a comment indicating an expected TS error. Let me know if this is problematic, as I'm not a type script expert when it comes to bundling.

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/navigation-api/navigation-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/navigation-api/navigation-api.ts
@@ -1,45 +1,63 @@
-export interface NavigationApiContent {
+export interface NavigationNavigateOptions {
   /**
-   * Using `(screenName: string, params?: any)` navigates to a route in current navigation tree.
-   * Pushes the specified screen if it isn't present in the navigation tree, goes back to a created screen otherwise.
-   * @param screenName the name of the screen you want to navigate to.
-   * @param params the parameters you want to pass to that screen.
+   * Developer-defined information to be stored in the associated NavigationHistoryEntry once the navigation is complete, retrievable via getState().
    */
-  navigate(screenName: string, params?: any): Promise<void>;
-
-  /**
-   * Using `(uri: string | URL)` opens a POS Native screen modally.
-   * If the uri starts with `shopify:point-of-sale/` it will be treated as a POS Native screen. Otherwise it will try to navigate to an extension screen.
-   * Available POS Native screens:
-   * - `shopify:point-of-sale/products/{product_id}` to present product details.
-   * - `shopify:point-of-sale/products/{product_id}/variants/{variant_id}` to present product variant details.
-   * - `shopify:point-of-sale/customers/{customer_id}` to present customer details.
-   * - `shopify:point-of-sale/orders/{order_id}` to present order details.
-   * - `shopify:point-of-sale/draft_orders/{draft_order_id}` to present draft order details.
-   * - `shopify:point-of-sale/staff/{staff_id}` to present staff details.
-   * @param uri the uri of the POS Native screen to present modally.
-   * @returns A promise that resolves when the POS screen is presented, rejects when an error occurs.
-   * @example Open product variant details screen for product id 123 and variant id 456
-   * navigate('shopify:point-of-sale/products/123/variants/456');
-   */
-  navigate(uri: string | URL): Promise<void>;
-
-  /** Pops the currently shown screen */
-  pop(): void;
-
-  /** Dismisses the extension. */
-  dismiss(): void;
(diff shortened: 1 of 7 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/api/navigation-api/navigation-api.ts
```diff
@@ -1,45 +1,68 @@
...
+  navigate: NavigateFunction;
+  /**
+   * The currentEntry read-only property of the Navigation interface returns a NavigationHistoryEntry object representing the location the user is currently navigated to right now.
+   */
+  currentEntry: NavigationHistoryEntry;
+  /**
+   * The updateCurrentEntry() method of the Navigation interface updates the state of the currentEntry; used in cases where the state change will be independent of a navigation or reload.
+   */
+  updateCurrentEntry(options: NavigationUpdateCurrentEntryOptions): void;
+  back(): void;
```
robin-drexler, line 46: does this need to be added to ui-api-design first?

packages/ui-extensions/src/surfaces/point-of-sale/api/navigation-api/navigation-api.ts
```diff
@@ -1,45 +1,68 @@
...
-  canNavigate(uri: string | URL): Promise<boolean>;
+  navigate: NavigateFunction;
+  /**
+   * The currentEntry read-only property of the Navigation interface returns a NavigationHistoryEntry object representing the location the user is currently navigated to right now.
+   */
+  currentEntry: NavigationHistoryEntry;
+  /**
+   * The updateCurrentEntry() method of the Navigation interface updates the state of the currentEntry; used in cases where the state change will be independent of a navigation or reload.
+   */
+  updateCurrentEntry(options: NavigationUpdateCurrentEntryOptions): void;
```
robin-drexler: > Note that we are still missing updateCurrentEntry as it's not a requirement currently for us

why add it to the types if you're not implementing it?

Or are you at least going to provide a noop?

js-goupil: Ah - I was trying to see if I could get it to work since we are exposing the signal inside extensibility host. Alas, it doesn't work after all since it's read only. I intended to remove this.

packages/ui-extensions/src/surfaces/point-of-sale/api/navigation-api/navigation-api.ts
js-goupil: I will open a PR there tomorrow - but I would prefer to ship this assuming [it won't be an issue](https://developer.mozilla.org/en-US/docs/Web/API/Navigation/back) as is and it's definitely a requirement for mobile navigation in 2025-10. Without it, mobile navigation is regressed

packages/ui-extensions/src/surfaces/point-of-sale/api/navigation-api/navigation-api.ts
js-goupil: Removed

packages/ui-extensions/src/surfaces/point-of-sale/api/navigation-api/navigation-api.ts
```diff
@@ -1,45 +1,63 @@
...
-   * - `shopify:point-of-sale/products/{product_id}` to present product details.
-   * - `shopify:point-of-sale/products/{product_id}/variants/{variant_id}` to present product variant details.
-   * - `shopify:point-of-sale/customers/{customer_id}` to present customer details.
-   * - `shopify:point-of-sale/orders/{order_id}` to present order details.
-   * - `shopify:point-of-sale/draft_orders/{draft_order_id}` to present draft order details.
-   * - `shopify:point-of-sale/staff/{staff_id}` to present staff details.
-   * @param uri the uri of the POS Native screen to present modally.
-   * @returns A promise that resolves when the POS screen is presented, rejects when an error occurs.
-   * @example Open product variant details screen for product id 123 and variant id 456
-   * navigate('shopify:point-of-sale/products/123/variants/456');
```
js-goupil, line 23: Let's make sure to just document this stuff separately for POS docs

henrytao-me (approved):

robin-drexler (approved):

PR #3327, opened 2025-09-18
UI extensions updates

oliverigor:
### Background

Preparing checkout soft release

(Add list of changes)

>[!IMPORTANT]
> - For components I removed the QueryContainer, DatePicker, DateField, Choice and ChoiceList that are being generated on the d.ts creation.
> - Removed Announcement docs since it was causing a failure due to it's types not being included in the d.ts generation yet

### Solution

- ran prepare-package-release on the monorepo.

### 🎩

- snapshot below (🙏🏽 )
- Will post docs in a bit.

### Checklist


packages/ui-extensions/src/docs/shared/components/index.ts
```diff
@@ -7,8 +8,8 @@ export {default as Checkbox} from './Checkbox';
 export {default as ChoiceList} from './ChoiceList';
 export {default as Clickable} from './Clickable';
 export {default as ClipboardItem} from './ClipboardItem';
-export {default as DateField} from './DateField';
```
jun-shop, line 10: did we also want to remove DatePicker? currently that's for R2

oliverigor: Yes, good catch, I'll remove with a commit fixing the lint issues. thanks

oliverigor: Actually, this is the shared docs with Admin, so they might be using it already, so no need to remove it from here a least

packages/ui-extensions/src/surfaces/checkout/components/Badge.d.ts
```diff
@@ -25,14 +25,14 @@ export interface BaseElementPropsWithChildren<TClass = HTMLElement> extends Base
     children?: preact.ComponentChildren;
 }
 /**
- * The list of available icon types.
- *
- * @see https://github.com/Shopify/ui-api-design/blob/main/packages/ui-api-design/src/components/Icon/Icon.ts#L10
+ * The list of available icon types
+ * Ideally we would infer the type from CHECKOUT_AVAILABLE_ICONS by doing (typeof CHECKOUT_AVAILABLE_ICONS)[number]
+ * but the docs tool does not support this syntax.
```
jun-shop: ⚰️ lol oh no i don't like that this comment got pulled in

oliverigor: :yikes: yeah, I can manually remove it... but this will be fixed next week, since the generate-docs tool was "fixed" and we will be able to bump the package and add our previous fancy fix for this.

packages/ui-extensions/src/surfaces/checkout/components/MapMarker.d.ts
```diff
@@ -40,14 +45,6 @@ export interface MapMarkerElementEvents {
      */
     click?: ((event: CallbackEventListener<typeof tagName>) => void) | null;
 }
-export interface MapMarkerSlots {
-    /**
-     * The graphic to use as the marker.
-     *
-     * If unset, it will default to the provider’s default marker.
-     */
-    graphic?: ReactNode;
```
jun-shop, line 49: is this right? i get confused with slots. i thought MapMarker had a slot for graphic

oliverigor: it is right, and will keep having support despite this change, we are just adding the types internally, the `graphic`/`slots` support as children's are still in place, this is just removing it as a preact property on the component

packages/ui-extensions/docs/surfaces/checkout/reference/examples/purchase.checkout.chat.render/default.example.tsx
```diff
@@ -1,3 +1,4 @@
+import '@shopify/ui-extensions/preact';
```
jun-shop: ah shoot - since we're not planning on shipping `Chat` for 2025-10 should we hide these extension targets? I don't think they can be used without the accompanying chat component. cc @cpeddecord

packages/ui-extensions/docs/surfaces/checkout/reference/examples/purchase.thank-you.chat.render/default.example.tsx
```diff
@@ -1,3 +1,4 @@
+import '@shopify/ui-extensions/preact';
```
jun-shop: same question here re chat

packages/ui-extensions/docs/surfaces/checkout/reference/examples/purchase.checkout.chat.render/default.example.tsx
jun-shop: never mind. just saw the note in `packages/ui-extensions/src/surfaces/checkout/shared.ts` - i wonder if it makes sense though to remove the example code since it references an `s-chat` component that doesn't exist?

packages/ui-extensions/src/surfaces/checkout/components/components.d.ts
```diff
@@ -13,6 +13,7 @@
  * https://github.com/Shopify/ui-api-design/issues/139
  */
 export type ComponentChildren = any;
+export type StringChildren = string;
```
jun-shop, line 16: just for my own benefit - what's the difference between this file and `packages/ui-extensions/src/surfaces/checkout/components/components-shared.d.ts`? it seems to be repeated content

oliverigor: The `d.ts` build script does this split and we should be using only the `components-shared.d.ts` (meaning that anything that our components share, are on that file) so `components.d.ts` is ui-api-design library all inlined, and `components-shared.d.ts` is extracted from `components.d.ts` and it is what is shared from our components. 

I'll look into using only the shared file for the final release, I just had problems with that approach (back in June) but I think we can remove that dependency now :D

packages/ui-extensions/docs/surfaces/checkout/reference/examples/purchase.checkout.chat.render/default.example.tsx
oliverigor: Hum, yeah, we might want to remove them

packages/ui-extensions/src/surfaces/checkout/components/components-shared.d.ts
```diff
@@ -1788,16 +2143,17 @@ interface DividerProps$1 extends GlobalProps {
 	 */
 	color?: ColorKeyword;
```
jun-shop, line 2144: hey can we remove this somehow? i don't think we want to expose color

oliverigor: this is just the props that comes from ui-api-design, inlined by the d.ts file, this is not being exported in the component type (since we're not picking it anymore) and consequently not avail on the docs either.

jun-shop: got it thanks

packages/ui-extensions/src/surfaces/checkout/components/Divider.d.ts
```diff
@@ -11,7 +11,7 @@
 import type {DividerProps$1} from './components-shared.d.ts';
 
 declare const tagName = "s-divider";
-export interface DividerProps extends DividerProps$1 {
+export interface DividerProps extends Pick<DividerProps$1, 'direction' | 'id'> {
```
jun-shop, line 29: is line 26 a problem? it's still referencing `DividerProps$1`?

oliverigor: wow, that was a good catch, this is also wrong on shop/world, need to change there and re-run the scripts :dead: 

good catch though.

packages/ui-extensions/src/surfaces/checkout/components/Sheet.d.ts
```diff
@@ -72,11 +72,17 @@ export interface SheetElementSlots {
      */
     'secondary-actions'?: HTMLElement;
 }
-export interface SheetElement extends SheetElementProps, SheetSlots, Omit<HTMLElement, 'id'> {
+export interface SheetElementMethods extends Pick<SheetProps$1, 'hideOverlay'> {
+}
+export interface SheetElement extends SheetElementProps, SheetElementMethods, Omit<HTMLElement, 'id'> {
+    afterhide: SheetEvents['onAfterHide'];
+    aftershow: SheetEvents['onAfterShow'];
```
jun-shop, line 79: just curious why we need this?

oliverigor: iirc, so something like this 

`sheetElement.addEventListener('afterhide', ()=> console.log) ` 

works without any typing issue

packages/ui-extensions/src/surfaces/checkout/components/Divider.d.ts
oliverigor: fixed :D

jun-shop (approved):
LGTM! I've mostly focused my testing on **components only**. Thank you!

PR #3330, opened 2025-09-19
Fix customer account docs build command

andrewmcgov:
## Summary

Fixed the `yarn docs:customer-account` command which was failing with a "Cannot find module" error since the recent update to `@shopify/generate-docs`

The customer-account documentation build script had some differences from the checkout version that were causing the errors. I made two changes to the arguments we pass to `generate-docs`:

1. Updated the `--input` from `--input ./${docsRelativePath}/reference ./${docsRelativePath} ./${docsRelativePath}/../../../src/surfaces/customer-account/components` to `--input ./${docsRelativePath}/reference ./${srcRelativePath}`
2. Added `--overridePath ./${docsRelativePath}/typeOverride.json`

## Testing

- Make some test changes to component docs and api docs in customer accounts
- `yarn docs:customer-account 2025-10-rc` should run successfully and you should see your changes


oliverigor (approved):
🎩 and now build runs with no errors, Thanks Andrew

lsit (approved):

PR #3335, opened 2025-09-19
removing chat docs from rc

cpeddecord:
removing chat docs from the rc


oliverigor (approved):
lint: fine
type-check: fine

igor: fine

PR #3349, opened 2025-09-23
18625: fix api docs

andy-chhuon:
Closes [#18625](https://github.com/shop/issues-retail/issues/18625)

### Background
Brought navigation docs back in (type is Navigation)
image

Fixed s-navigator/s-screen references
Fixed navigation api docs to use global object
Removed scanner.scanQRCode reference

### Checklist


NathanJolly (approved):

awoodall:
@andy-chhuon is it possible to get typing for the `navigate` call? To pass params you have to put it in the `state` object:

```tsx
 navigation.navigate("my-route", {
              state: {
                isAuthenticated: true,
              },
            });
```

Maybe a code example for sharing params as well?

awoodall:
It'd also be nice if we had a tutorial for passing params between screens.

awoodall (approved):
Approving for now but let's get a follow-up issue going for adding an additional example that showcases passing state between screens.

PR #3365, opened 2025-09-25
[chore]: Bump gen docs lib to 0.19.6

oliverigor:
### Background

- We need to bump the generate-docs library so when testing we get the correct output from the library and avoid discrepancies.
- This also brings the default values to show on extracted properties.

### Solution

- Bumping the lib

### 🎩

- `yarn docs:checkout 2025-10-rc` and you should see Default values for properties showing when they're defined

### Checklist


ncardeli (approved):
LGTM

PR #3377, opened 2025-09-26
18241: update starting guide

andy-chhuon:
### Background

image

### Checklist


packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/examples/getting-started/generate-extension.bash
```diff
@@ -1,2 +1,2 @@
 shopify upgrade
-shopify app generate extension
+POLARIS_UNIFIED=true shopify app generate extension
```
NathanJolly, line 2: Do you know what the plan is for this env var? In other words, should we be publicly documenting it as an option, or is it there for internal use for now and will be the default behavior after Oct 1?

NathanJolly (approved):
Env var documentation is expected 👍

PR #3409, opened 2025-09-30
Remove pos observe targets docs

mauriciosierra:
### Background

Resolves https://github.com/Shopify/shopify-dev/issues/62978

Observe targets are in dev preview mode and they will change significantly before being released (see linked issue). To avoid developer confusion in the meantime, we are removing the docs for these targets while we finalize the new form of these targets.

### Solution

Remove existing `observe` dev preview docs.

### 🎩

- Ensure no `observe` targets appear in docs for 2025-10

### Checklist


packages/ui-extensions/docs/surfaces/point-of-sale/staticPages/pages/versions.doc.ts
```diff
@@ -0,0 +1,280 @@
+import type {LandingTemplateSchema} from '@shopify/generate-docs';
```
mauriciosierra: I'm not sure why this file showed up here. Should this be a generated file?

aaronschubert0 (approved):
Tested and can't find any references to observe targets left in the docs. Thanks again, this should reduce partner confusion about these target types 🙂

PR #3412, opened 2025-09-30
Fix typos and missing component in StandardComponents list

awoodall:
### Background

This PR adds missing components to the Point of Sale UI extensions: `Divider` and `QrCode` (case-sensitive alternative to the existing `QRCode`).

### Solution

Added `Divider` and `QrCode` to the `StandardComponents` type in the Point of Sale components. The `QrCode` variant is marked as case-sensitive and important for the 2025-10 release, similar to the existing pattern with `POSBlock` and `PosBlock`.

### 🎩

- Verify that both new components are properly recognized in the type system
- Test that extensions using these components render correctly

### Checklist


tsdevshop (approved):
@tsdevshop approved this PR from Slack with [Graphite](https://app.graphite.dev/github/pr/Shopify/ui-extensions/3412)

NathanJolly (approved):

PR #3427, opened 2025-09-30
Remove preview observe targets from developer docs

mauriciosierra:
### Background

Porting PR #3409 over to `2025-10` since we missed the `2025-10-rc` cutoff

### Solution

Cherry picking already merged commit from RC branch

### 🎩

- Ensure all `observe` targets and references have been removed (this was already tophatted and approved in original PR)

### Checklist


packages/ui-extensions/docs/surfaces/point-of-sale/mdxExamples/loyalty-example/shopify.extension.toml
```diff
@@ -15,7 +15,4 @@ description = "Loyalty Points Extension"
 module = "./src/CustomerDetailsBlock.tsx"
 target = "pos.customer-details.block.render"
 
-[[extensions.targeting]]
```
NathanJolly, line 18: I'm deleting the loyalty example altogether so this is fine

NathanJolly (approved):
I'm a bit weary of this change as it removes docs for targets that are being used in production apps, but I understand the tradeoff that having the documentation public is causing developer confusion as to why they can't use them.

awoodall (approved):

PR #3428, opened 2025-09-30
make remote-dom/polaris updates to readme that is displayed on npm package for public package

lrsterrett:
Before:

image

After:

image

### DIFF
--- a/.changeset/pretty-yaks-attend.md
+++ b/.changeset/pretty-yaks-attend.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+update readme to reflect 2025-10 related api changes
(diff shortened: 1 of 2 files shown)


.changeset/pretty-yaks-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+udpate readme to reflect 2025-10 related api changes
```
oliverigor:
```suggestion
Update readme to reflect 2025-10 related api changes
```

oliverigor (approved):
I wonder do we need a patch at all ?

.changeset/pretty-yaks-attend.md
lrsterrett: well I think without a changset it won't trigger a release so the readme will remain out of date, right? or do you think the readme would update regardless?

PR #3442, opened 2025-10-01
Update choice list example to include onChange event

kyledurand:
### Background

This is a weird API and we should document how to get change values here.


jas7457 (approved):
Thanks for adding this!

sam-b-rose (approved):

PR #3459, opened 2025-10-01
make ts steps more obvious in migration guide

lrsterrett:
Updated migration guides for checkout, pos, and customer accounts. Admin doesn't have one as of now.

Need was based on [slack thread](https://shopify.slack.com/archives/C08H644FPSQ/p1759336938977509) to make each step more obvious and not hidden in a collapsible accordion.

Before:

image

After:

image


shayarnett (approved):
🎩 on staging looked good!

PR #3460, opened 2025-10-02
Add JSX support for component examples in admin docs

davebcn87:
### Background

Admin component examples were previously written in HTML format using custom elements (e.g., `<s-number-field>`). This PR adds support for writing examples in JSX format, which allow us to showcase interaction with the components, and javascript-only attributes. This will also enable us to do type checking on examples in the future.

### Solution

Added JSX preview support to the documentation build system:

**Build System Changes (`build-docs.mjs`):**
- Added new `jsxWrapper()` function that generates HTML with Preact runtime, Babel standalone transpiler, and necessary polyfills
- Auto-wraps simple JSX expressions with `return` statement, while supporting complete function bodies with hooks
- Extended `createTemplate()` to accept a `jsx` flag that determines whether to use JSX or HTML wrapper
- Added `preview-jsx` language type alongside existing `preview` type
- Updated tab processing logic to handle both HTML and JSX preview formats

**Component Example Updates:**
- Converted `NumberField` default example from HTML to JSX format
- Changed file extension from `.html` to `.jsx`
- Updated numeric attributes to use JSX syntax (e.g., `step={5}` instead of `step="5"`)

### Demos

https://github.com/user-attachments/assets/f8a58cbc-7a43-423e-9cc9-cd10b2bed9cc

https://github.com/user-attachments/assets/66afa7aa-52c8-4fe2-bfc9-ccc1ce40e4fd

### 🎩

- [x] JSX examples render correctly in documentation preview
- [x] HTML examples continue to work as before
- [x] NumberField example displays properly with JSX format
- [x] Build process successfully generates documentation with JSX examples

### Checklist


kyledurand (approved):
Approving to unblock this. The example isn't super broken but we should fix in a follow up. 

**Avatar: Error handling example**
```bash
parse-error.ts:95 Uncaught SyntaxError: /Inline Babel script: Unexpected token, expected "," (5:3)

  3 |   initials="CS"
  4 |   alt="Customer support"
> 5 |  />;
    |    ^
  6 | {
  7 |   /* Will display "CS" initials when image fails */

```

PR #3479, opened 2025-10-08
Update Admin components and remove connectedCallback prop from Page

sam-b-rose:
### Background

Slack 🧵 
https://shopify.slack.com/archives/C06UT7UK3R7/p1759936444085199

### Solution

Screenshot 2025-10-08 at 12 01 53 PM

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components.d.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components.d.ts
@@ -1,4 +1,4 @@
-/** VERSION: 1.21.2 **/
+/** VERSION: 1.22.1 **/
 
 /* eslint-disable @typescript-eslint/ban-types */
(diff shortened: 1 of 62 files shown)


packages/ui-extensions/src/surfaces/admin/components.d.ts
```diff
@@ -6633,7 +6639,9 @@ declare class Page extends PreactCustomElement implements PageProps {
   accessor inlineSize: PageProps['inlineSize'];
   accessor heading: PageProps['heading'];
   constructor();
+  /** @private */
   connectedCallback(): void;
+  /** @private */
```
sam-b-rose, line 6644: 👆 here are the key lines that fix the Page docs

davejcameron (approved):

PR #3500, opened 2025-10-14
Fix JSDoc description truncation by pinning TypeScript to 4.9.x in generate-docs

davebcn87:
## Problem

JSDoc descriptions containing `@property` (and other JSDoc tags like `@param`, `@returns`) in the middle of sentences are being truncated when generating documentation.

**Example:**
**Before:** 
`Queue a run of the render function. You shouldn't need to call this manually - it should be handled by changes to @property values.`

**After (broken):** 
`Queue a run of the render function. You shouldn't need to call this manually - it should be handled by changes to`

## Root Cause

TypeScript 5.1+ introduced a breaking change in JSDoc parsing where it now treats `@tagname` as a JSDoc tag **even when it appears in the middle of descriptive text** (not just at line starts).

When `@shopify/generate-docs@0.19.8` was published on October 8, 2025, we changed the resolutions to use typescript 5+ as well in this repo. See [this commit](https://github.com/Shopify/ui-extensions/pull/3485/commits/42e445d62ccff1ae83cb3373f91a905574b1b075) from @gwyneplaine.

## Solution

Use Yarn resolutions to force `@shopify/generate-docs` and `@shopify/type-harvester` to use TypeScript 4.9.x instead of 5.9.3:

```json
"resolutions": {
  "**/typescript": "^4.9.5",
}
```

This ensures the bundled TypeScript version uses the working parser behavior from 4.9.x.

## Testing

- Verified TypeScript 4.9.5 preserves full JSDoc text including `@property` mentions
- Verified TypeScript 5.0.4 also works (but 5.1.0+ is broken)
- This fix allows any 4.9.x version while blocking problematic 5.x versions

## Alternative Considered

Escaping `@property` as ``@property`` in the actual source code is the best long term fix, so whenever we update the parsing script to be fully in Typescript 5+ we don't get the issue. We'll solve that in another PR, but we will at least unblock the docs generation.

### 🎩

- Verified documentation generation no longer truncates JSDoc descriptions at `@property` mentions

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "./packages/*"
   ],
   "resolutions": {
-    "**/typescript": "4.9.5"
+    "**/typescript": "^4.9.5"
   },
   "scripts": {
     "build": "loom build",
--- a/package.json
+++ b/package.json
@@ -6,6 +6,9 @@
   "workspaces": [
     "./packages/*"
   ],
+  "resolutions": {
+    "**/typescript": "4.9.5"
+  },
   "scripts": {
     "build": "loom build",
     "build-consumer": "loom build && ./scripts/build-consumer.sh",
--- a/yarn.lock
+++ b/yarn.lock
@@ -6552,7 +6552,7 @@ typedarray-to-buffer@^3.1.5:
   dependencies:
     is-typedarray "^1.0.0"
 
-typescript@4.9.5, typescript@^4.3.5, "typescript@^4.8.3 || ^5.0.0", typescript@^4.9.0, typescript@^5.6.0:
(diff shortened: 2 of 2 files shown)


jas7457 (commented):
As discussed on slack, I'm a bit hesitant with the resolutions approach. It can force-downgrade (or upgrade) to breaking versions that a package does not supports.

I think, instead, we should try to find a way to escape the characters or remove them or something else.

majd-shopify:
@davebcn87 has there been a follow up to this?

Just asking to know whether I can close https://github.com/Shopify/temp-project-mover-Archetypically-20260312105649/issues/208 or not

PR #3506, opened 2025-10-15
Fix outdated references to checkout-web-ui in comment links

andrewmcgov:
### Background

Closes https://github.com/shop/issues-checkout/issues/8516

### Solution

Updated incorrect references to checkout-web-ui when they should link to customer account docs instead.

Some references remain, but that is because we are intentionally linking to checkout docs. For example we link to the button component which is only documented on the checkout side.

### 🎩

- Copy the changed links into your browser to make sure they work

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
@@ -190,7 +190,7 @@ export interface OrderStatusApi<Target extends ExtensionTarget> {
 
   /**
    * The metafields requested in the
-   * [`shopify.ui.extension.toml`](https://shopify.dev/docs/api/checkout-ui-extensions/configuration)
+   * [`shopify.ui.extension.toml`](https://shopify.dev/docs/api/customer-account-ui-extensions/configuration)
    * file. These metafields are updated when there's a change in the merchandise items
    * being purchased by the customer.
    *
@@ -241,7 +241,7 @@ export interface OrderStatusApi<Target extends ExtensionTarget> {
    * extension’s configuration file.
    *
    * @example 'customer-account.order-status.block.render'
-   * @see https://shopify.dev/docs/api/checkout-ui-extensions/unstable/extension-targets-overview
+   * @see https://shopify.dev/docs/api/customer-account-ui-extensions/extension-targets-overview
    * @see https://shopify.dev/docs/apps/app-extensions/configuration#targets
    *
    * @deprecated Deprecated as of version `2023-07`, use `extension.target` instead.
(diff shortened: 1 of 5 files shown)


packages/ui-extensions/src/surfaces/customer-account/preact/api.ts
```diff
@@ -8,12 +8,6 @@ import {CustomerAccountUIExtensionError} from './errors';
 /**
  * Returns the full API object that was passed in to your extension when it was created.
  * Depending on the extension target, this object can contain different properties.
- *
- * For example, the `customer-account.order-status.cart-line-item.render-after` extension target will return the [CartLineDetailsApi](https://shopify.dev/docs/api/checkout-ui-extensions/apis/cartlinedetailsapi) object.
- * Other targets may only have access to the [StandardApi](https://shopify.dev/docs/api/checkout-ui-extensions/apis/standardapi) object,
- * which contains a basic set of properties about the order.
- *
- * For a full list of the API available to each extension target, see the [ExtensionTargets type](https://shopify.dev/docs/api/checkout-ui-extensions/apis/extensiontargets).
```
andrewmcgov, line 16: We don't have 1:1 matches for these links in the new docs because we don't have pages for StandardApi or ExtensionTargets.

Since we are not documenting hooks anymore, I decided to just remove these links. The remaining comments still provide enough info.

packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
```diff
@@ -241,7 +241,7 @@ export interface OrderStatusApi<Target extends ExtensionTarget> {
    * extension’s configuration file.
    *
    * @example 'customer-account.order-status.block.render'
-   * @see https://shopify.dev/docs/api/checkout-ui-extensions/unstable/extension-targets-overview
+   * @see https://shopify.dev/docs/api/customer-account-ui-extensions/latest/extension-targets-overview
```
brianshen1990: I don;t think we need to add `latest`, without it we will just go to the latest one automatically

packages/ui-extensions/src/surfaces/customer-account/preact/api.ts
brianshen1990: that's very interesting one, we use StandardAPI everywhere in the target API page, but we don't have a page for it to link. I am not sure whether do we want to aggregate all standardAPIs into a page just like https://shopify.dev/docs/api/customer-account-ui-extensions/latest/apis/order-status-api/ or not. ( BTW, we don't have the page, i think we can customize the category page like https://shopify.dev/docs/api/customer-account-ui-extensions/latest/polaris-web-components  ). maybe Deanna can help

brianshen1990 (approved):

packages/ui-extensions/src/surfaces/customer-account/api/order-status/order-status.ts
andrewmcgov: Good call, updated!

packages/ui-extensions/src/surfaces/customer-account/preact/api.ts
andrewmcgov: Yeah that could be something to look at. Since these links were only part of the package, and for hooks that we don't document anymore, I think they are safe to remove for now.

PR #3509, opened 2025-10-17
Remove pos static pages that are migrating

NathanJolly:
### Background

https://github.com/shop/issues-retail/issues/19120

### Solution

Remove the static pages that are moving to the build docs and will now live in shopify-dev

### 🎩

- See companion docs pr: https://github.com/Shopify/shopify-dev/pull/63889

### Checklist


mcvinci (approved):
Thanks so much for putting this together so quickly, @NathanJolly! Is it possible to remove these pages from all versions just so we have alignment across versions? (Not sure if that would need to be done in separate PRs per version).

PR #3523, opened 2025-10-27
Footer Help and Resource List Patterns

mpshopify:
### Background

We did not have a Footer Help or Resource List pattern. A few aria-labels should have used accessibilityLabel instead. There was a typo with SetupGuide. s-image on EmptyState and calloutCard resulted in odd LLM rendering due to aspect ratio.

### Solution

Add patterns for Footer Help and Resource List. Change aria-labels for accessibilityLabel on button and clickable where applicable. Fix typo to Setup guide. Added comment to guide LLM to match image's aspect ratio to actual width and height.

### 🎩

- ...

### Checklist


packages/ui-extensions/src/surfaces/admin/components/patterns/footerHelp.ab.doc.ts
```diff
@@ -0,0 +1,48 @@
...
+  category: 'Patterns',
+  subCategory: 'Compositions',
+  thumbnail: '/assets/templated-apis-screenshots/admin/patterns/footerHelp.png',
+  defaultExample: {
+    codeblock: {
+      title: 'Footer help',
+      tabs: [
+        {
+          title: 'Preview',
+          code: 'examples/footerHelp.html',
```
jas7457, line 31: Do we also need the jsx tab?

packages/ui-extensions/src/surfaces/admin/components/patterns/resourceList.ab.doc.ts
```diff
@@ -0,0 +1,51 @@
...
+  subCategory: 'Compositions',
+  thumbnail:
+    '/assets/templated-apis-screenshots/admin/patterns/resourceList.png',
+  defaultExample: {
+    codeblock: {
+      title: 'Resource list',
+      tabs: [
+        {
+          title: 'Preview',
+          code: 'examples/resourceList.html',
```
jas7457, line 34: Do we also need the jsx tab?

packages/ui-extensions/src/surfaces/admin/components/patterns/examples/resourceList.html
```diff
@@ -0,0 +1,58 @@
...
+          <s-stack gap="small-200" padding="small-200">
+            <s-text-field value="VIP" placeholder="Add tag"></s-text-field>
+            <s-link href="">Clear</s-link>
+          </s-stack>
+        </s-popover>
+      </s-grid>
+      <s-button variant="secondary">Save</s-button>
+    </s-grid>
+
+    <s-stack paddingInline="base">
```
jas7457: Nit: could this just be an `s-box` instead? stacks normally should have more than 1 item in it, or else it doesn't really stack anything.

packages/ui-extensions/src/surfaces/admin/components/patterns/examples/resourceList.jsx
```diff
@@ -0,0 +1,58 @@
...
+    <s-grid gridTemplateColumns="1fr auto" gap="base" alignItems="center" paddingInline="base">
+      <s-checkbox label="Showing 2 customers"></s-checkbox>
+      <s-select>
+        <s-option value="newest">Newest update</s-option>
+        <s-option value="oldest">Oldest update</s-option>
+      </s-select>
+    </s-grid>
+    
+    <s-stack>
+      <s-clickable borderstyle="solid none none none" border="base" paddingInline="base" paddingBlock="small">
```
jas7457: I think this needs to be `borderStyle` - same for the one below and in the html file.

packages/ui-extensions/src/surfaces/admin/components/patterns/examples/resourceList.jsx
```diff
@@ -0,0 +1,58 @@
+<s-section padding="none">
+  <s-stack gap="small-200">
+    <s-grid gridTemplateColumns="1fr auto" gap="base" alignItems="center" paddingInline="base" paddingBlockStart="base">
+      <s-grid gridTemplateColumns="1fr auto" gap="small-200" alignItems="center">
+        <s-text-field icon="search" placeholder="Filter customers"></s-text-field>
+        <s-button commandFor="tagged-with">Tagged with</s-button>
+        <s-popover placement="bottom-start" id="tagged-with">
```
jas7457: Is `placement` a thing? I may just have old ts files, but I don't think this is part of our API. Check the html file too.

jas7457 (changes requested):
Mainly looking good, just caught a few things. Nice!

packages/ui-extensions/src/surfaces/admin/components/patterns/footerHelp.ab.doc.ts
mpshopify: Good catch, also updated 'empty state' which was missing it.

packages/ui-extensions/src/surfaces/admin/components/patterns/examples/resourceList.html
mpshopify: I think there may be situations where there are more than one tags, so they would stack, although inline. So i've added `direction="inline"  gap="small-400"`

jas7457 (approved):
Changes look good, nice job!

PR #3524, opened 2025-10-27
Add cashtracking observe target

merkoyep:
### Background

This PR adds the cash-tracking.event.observe target to the UI-Extensions repo. This new target will replace the existing cash-tracking-session-complete.event.observe and cash-tracking-session-start.event.observe targets, which we will be deprecating in a follow-up PR.

### Solution

This PR follows implementation patterns for .observe targets in the ui-extensions repo. Noting that documentation for observe targets have recently all been removed.

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/event/data/CashTrackingSessionData.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/event/data/CashTrackingSessionData.ts
@@ -1,6 +1,28 @@
 import {BaseData} from './BaseData';
 import {BaseApi} from './BaseApi';
 
+export type CashTrackingEventType =
+  | 'opening'
+  | 'closing'
+  | 'mid-session-count'
+  | 'adjustment'
+  | 'sale'
+  | 'refund';
+
+export interface CashTrackingEvent {
+  id: number;
+  type: CashTrackingEventType;
+  countExpectedAmount?: number;
+  countActualAmount?: number;
+  adjustmentAmount?: number;
+  note?: string;
+  orderId?: number;
+}
+
+export interface CashTrackingData extends BaseData, BaseApi {
+  cashTracking: CashTrackingEvent;
+}
+
 export interface CashTrackingSessionStartData extends BaseData, BaseApi {
   cashTrackingSessionStart: {
     id: number;
--- a/.changeset/famous-windows-unite.md
+++ b/.changeset/famous-windows-unite.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Add cash tracking observe target
--- a/packages/ui-extensions/src/surfaces/point-of-sale/extension-targets.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/extension-targets.ts
@@ -2,6 +2,7 @@ import {BaseOutput} from './event/output';
 import {
   CashTrackingSessionStartData,
   CashTrackingSessionCompleteData,
+  CashTrackingData,
 } from './event/data/CashTrackingSessionData';
 import {TransactionCompleteData} from './event/data/TransactionCompleteData';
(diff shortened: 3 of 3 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/event/data/CashTrackingSessionData.ts
```diff
@@ -1,6 +1,28 @@
...
+  countExpectedAmount?: number;
+  countActualAmount?: number;
+  adjustmentAmount?: number;
+  note?: string;
+  orderId?: number;
+}
+
+export interface CashTrackingData extends BaseData, BaseApi {
+  cashTracking: CashTrackingEvent;
+}
```
alexkung1, line 24: Is it possible to type this so that the interface is inferred by the type? e.g.,



```suggestion
export type CashTrackingEvent =
  | CashTrackingOpeningEvent
  | CashTrackingClosingEvent
  | CashTrackingMidSessionCountEvent
  | CashTrackingAdjustmentEvent
  | CashTrackingSaleEvent
  | CashTrackingRefundEvent;

interface CashTrackingCountEvent {
  id: number;
  countExpectedAmount: number;
  countActualAmount: number;
  adjustmentAmount: number;
  note: string;
}

export interface CashTrackingOpeningEvent extends CashTrackingCountEvent {
  type: 'opening';
}
export interface CashTrackingClosingEvent extends CashTrackingCountEvent {
  type: 'closing';
}

export interface CashTrackingMidSessionCountEvent
  extends CashTrackingCountEvent {
  type: 'mid-session-count';
}

export interface CashTrackingAdjustmentEvent {
  id: number;
  type: 'adjustment';
  adjustmentAmount: number;
  note: string;
}

interface CashTrackingOrderEvent {
  id: number;
  orderTransactionId: number;
  orderId: number;
  note: string;
}

export interface CashTrackingSaleEvent extends CashTrackingOrderEvent {
  type: 'sale';
  saleAmount: number;
}

export interface CashTrackingRefundEvent extends CashTrackingOrderEvent {
  type: 'refund';
  refundAmount: number;
}
```

packages/ui-extensions/src/surfaces/point-of-sale/event/data/CashTrackingSessionData.ts
```diff
@@ -1,6 +1,28 @@
...
+  | 'adjustment'
+  | 'sale'
+  | 'refund';
+
+export interface CashTrackingEvent {
+  id: number;
+  type: CashTrackingEventType;
+  countExpectedAmount?: number;
+  countActualAmount?: number;
+  adjustmentAmount?: number;
```
alexkung1, line 17: Couple questions about the money amounts:
- Is there an opinion on whether to represent money amounts as a `string` vs `number`? I see that we [sometimes do strings](https://github.com/Shopify/ui-extensions/blob/10-27-add_cashtrackingsession_observe_target/packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts#L12-L14) and [sometimes do numbers](https://github.com/Shopify/ui-extensions/blob/095bd61c063844e7e0d5fd43189ccba3968cae25/packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts#L38), wondering if there's a preference here
- Do we need a currency associated with the amounts? Again, seems inconsistent but[ I see that `Discount` has a currency attached](https://github.com/Shopify/ui-extensions/blob/095bd61c063844e7e0d5fd43189ccba3968cae25/packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts#L87-L92) whereas other places we don't. Can we check in with the extensions team to see what the preference is? seems like attaching the currency would be prudent in this scenario

merkoyep: Good point, I'll add currency back in for safety, and since these events pretty much directly correlate with out cash management operations, I'll follow that data structure directly, which is currency + amount as an int/number :)

PR #3536, opened 2025-10-30
Adds components to POS cart line item interface

alexgraffeocohen:
Adds components to Cart interface for point of sale so that we can interact with product bundle components.

- New LineItemComponent interface with a subset of LineItem fields
- Optional components field on LineItem for bundle line items
- Components will only be populated for product bundles

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
@@ -55,6 +55,33 @@ export interface LineItem {
    * The currently selected selling plan for this line item.
    */
   sellingPlan?: SellingPlan;
+  /**
+   * Bundle components for this line item. Only present for product bundles.
+   * Each component represents an individual item within the bundle with its own tax information.
+   */
+  components?: LineItemComponent[];
+}
+
+/**
+ * Represents a component of a product bundle line item.
+ * Bundle components contain the individual items that make up a bundle,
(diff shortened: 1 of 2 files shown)


mauriciosierra (approved):
🚢

alexgraffeocohen:
@mauriciosierra Made slight changes since you reviewed to make sure `LineItemComponent` more closely mirrored `LineItem` interface.

fatbattk (approved):

packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
```diff
@@ -55,6 +55,33 @@ export interface LineItem {
    * The currently selected selling plan for this line item.
    */
   sellingPlan?: SellingPlan;
+  /**
+   * Bundle components for this line item. Only present for product bundles.
+   * Each component represents an individual item within the bundle with its own tax information.
+   */
+  components?: LineItemComponent[];
```
fatbattk, line 62: nit: clearer naming `bundleComponents`?

alexgraffeocohen: I thought about it, but hesitated because [Admin API](https://shopify.dev/docs/api/admin-graphql/latest/objects/DraftOrderLineItem#field-DraftOrderLineItem.fields.components) tends to call this just `components` and I wanted to be consistent.

PR #3554, opened 2025-11-07
Admin: expose picker and resource API for all rendering targets

vividviolet:
# 🎩 
1. Generate a order routing rule settings extension `shopify app generate extension --template order_routing_location_rule_ui`.
2. Update your package.json to consume the snapshot
```json
"@shopify/ui-extensions": "https://registry.npmjs.org/@shopify/ui-extensions/-/ui-extensions-0.0.0-snapshot-20251107182616.tgz"
```
3. Verify that you can access `shopify.picker` and `shopify.resourcePicker`

### DIFF
--- a/.changeset/wicked-singers-float.md
+++ b/.changeset/wicked-singers-float.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+Admin: Expose picker and resource picker APIs in the standard API for rendering extensions
--- a/packages/ui-extensions/src/surfaces/admin/api.ts
+++ b/packages/ui-extensions/src/surfaces/admin/api.ts
@@ -1,5 +1,6 @@
 export type {I18n, I18nTranslate} from '../../api';
 export type {StandardApi, Intents} from './api/standard/standard';
+export type {StandardRenderingExtensionApi} from './api/standard/standard-rendering';
 export type {Navigation} from './api/block/block';
 export type {
   CustomerSegmentTemplateApi,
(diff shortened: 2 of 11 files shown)


billfienberg:
Do we also need to implement this in `2026-01` and/or `2026-04-rc`?

billfienberg:
## 🎩 Tophat

<details>

<summary>Sample Code</summary>

```jsx
// extensions/order-routing-location-rule-ui/src/OrderRoutingLocationRule.jsx
import {render} from 'preact';

export default async () => {
  render(<Extension />, document.body);
};

function Extension() {
  async function openPicker() {
    shopify.picker({
      heading: 'Select an item',
      headers: [],
      items: [],
    });
  }
  async function openResourcePicker() {
    await shopify.resourcePicker({
      type: 'product',
      multiple: true,
    });
  }

  return (
    <s-function-settings>
      <s-button onClick={openPicker}>Open Picker</s-button>
      <s-button onClick={openResourcePicker}>Open Picker</s-button>
    </s-function-settings>
  );
}
```

</details>

### Before

❌ Type errors are **present**

```
Property 'picker' does not exist on type 'OrderRoutingRuleApi<"admin.settings.order-routing-rule.render">'.ts(2339)
Property 'resourcePicker' does not exist on type 'OrderRoutingRuleApi<"admin.settings.order-routing-rule.render">'.ts(2339)
```

image

### After

✅ Type errors are **absent**

image

billfienberg:
Question about the API design: the PR adds `resourcePicker` and `picker` to both `StandardApi` and `StandardRenderingExtensionApi`. This means non-rendering targets like `admin.app.tools.data` would also get these APIs even though they can't use them (no UI to trigger pickers).

Was the intent to only add pickers to `StandardRenderingExtensionApi`? That would keep `StandardApi` minimal for data-only targets:

- `StandardApi` - base (no pickers)
- `StandardRenderingExtensionApi` extends `StandardApi` + pickers
- Rendering targets use `StandardRenderingExtensionApi`
- Data targets use `StandardApi`

---
🤖 This comment was generated with Claude Code

PR #3568, opened 2025-11-14
[POS UI extensions]: Improved API, component, and prop descriptions for 2024-04

mcvinci:
## This PR: 
- Improves all API, component, and prop descriptions in the POS UI extensions docs for version 2024-04
- Relates to the [POS UI extensions reference prototype](https://pr-62503.shopify-dev.shopify.io/docs/api/templated-refs-v2/latest)
- Partially fixes https://github.com/Shopify/shopify-dev/issues/64745

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
 
       - id: typescript-cache
         name: Restore TypeScript cache
-        uses: actions/cache@v1
+        uses: actions/cache@v4
         with:
           path: |
             packages/*/build/ts
@@ -31,7 +31,7 @@ jobs:
 
       - id: jest-cache
         name: Restore jest cache
-        uses: actions/cache@v1
+        uses: actions/cache@v4
         with:
           path: .loom/cache/jest/
           key: ${{ runner.os }}-jest-v1-${{ github.sha }}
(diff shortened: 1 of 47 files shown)


NathanJolly:
Unfortunately, 2024-04 docs aren't actually generated from the 2024-04 branch, as this branch was not in a good state and does not include the deployment scripts. In the past when I backfill 2024-04 docs, I've just been manually changing the generated JSON in shopify-dev. It may be fine to simply delete the 2024-04 version docs at this point as they're really old, but otherwise this is a manual process. I'd recommend closing this and leaving 2024-04 alone as the path of least resistance. Sorry about this!

mcvinci:
> Unfortunately, 2024-04 docs aren't actually generated from the 2024-04 branch, as this branch was not in a good state and does not include the deployment scripts. In the past when I backfill 2024-04 docs, I've just been manually changing the generated JSON in shopify-dev. It may be fine to simply delete the 2024-04 version docs at this point as they're really old, but otherwise this is a manual process. I'd recommend closing this and leaving 2024-04 alone as the path of least resistance. Sorry about this!

@NathanJolly Thanks so much for this context! I'll close down this branch. I can easily backfill the 2024-04 docs in the generated JSON file in `shopify-dev`, but if we're planning to remove that version from the docs entirely soon, that works as well. I'll spin up a quick PR in `shopify-dev` in the meantime to cover things for the time being!

PR #3579, opened 2025-11-18
[App Gen] Updated Docs For Modal Bug in App Gen

Tumph:
### Background

Resolves [shop/issues-sidekick/issues/#2055](https://github.com/shop/issues-sidekick/issues/2055)

Improves the TypeScript definitions for the Modal component's overlay control methods.

### Problem
The documentation for modal overlay methods (`hideOverlay`, `showOverlay`, and `toggleOverlay`) lacks clarity on how these methods should be used programmatically versus through user interactions..

### Solution

This PR updates the TypeScript definitions for the `Modal` component by moving the overlay control methods (`hideOverlay`, `showOverlay`, and `toggleOverlay`) from the excluded props list to be properly defined as methods with detailed JSDoc comments. 

The added documentation clarifies that these methods are meant to be called imperatively on the modal element instance (via refs) rather than being passed as props. It also provides guidance on the proper usage patterns for controlling modal visibility through user interactions.

### 🎩

- Verify that TypeScript correctly recognizes these methods when using refs with Modal components
- Check that the documentation appears correctly in IDE tooltips

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/components/Modal.d.ts
+++ b/packages/ui-extensions/src/surfaces/admin/components/Modal.d.ts
@@ -39,13 +39,7 @@ export type RequiredAlignedModalProps = Required<ModalProps$1>;
 export interface ModalProps
   extends Pick<
     RequiredAlignedModalProps,
-    | 'accessibilityLabel'
-    | 'heading'
-    | 'padding'
-    | 'size'
-    | 'hideOverlay'
-    | 'showOverlay'
-    | 'toggleOverlay'
+    'accessibilityLabel' | 'heading' | 'padding' | 'size'
   > {
   /**
    * Adjust the size of the Modal.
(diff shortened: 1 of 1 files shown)


Tumph:
* **#3579** <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/3579?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/3579?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `2025-10`


<h2></h2>

hughnguy (approved):

davebcn87 (changes requested):
I don't think that will work, that typescript definition only affects the docs, but the real component won't have those methods. 

Let's discuss possible other solutions.

spiderbites:
> I don't think that will work, that typescript definition only affects the docs, but the real component won't have those methods.
> 
> Let's discuss possible other solutions.

Correct me if I'm wrong but the real component _does_ have these methods. They weren't added here, just documented

Tumph:
Yeah we just want it to use the methods properly, rather than incorrectly. Here is the diff before and after this change: https://docs.google.com/document/d/10NzJe2FEFfL0g4xrWIE8sMDRmZq111RWD6OFYCfqecc/edit?tab=t.0

spiderbites:
stale

PR #3570, opened 2025-11-14
20644: Implement Camera API Standard

andy-chhuon:
### Background

(Provide a link to any relevant issues AND provide a TLDR of the issue in this pull request)

### Solution

(Describe your solution, why this approach was chosen, and what the alternatives/impacts may be)

### 🎩

- ...

### Checklist

- [ ] I have :tophat:'d these changes
- [ ] I have updated relevant documentation

PR #3597, opened 2025-11-25
Add Camera API type definitions

fatbattk:
Closes [#21032](https://github.com/shop/issues-retail/issues/21032)

### Background

Add type definitions for Camera API

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
@@ -1,0 +1,56 @@
+export interface CameraMediaOptions {
+  /**
+   * The camera that will be active when the camera interface first opens.
+   * - `'user'`: The front-facing camera
+   * - `'environment'`: The rear-facing camera
+   * @defaultValue 'environment'
+   */
+  facingMode?: 'user' | 'environment';
+  /**
+   * The maximum width (1 to 1920) of the image in pixels. Resizes the image to this width if it is larger.
+   * @defaultValue 1080
+   */
+  maxWidth?: number;
+  /**
+   * The maximum height (1 to 1920) of the image in pixels. Resizes the image to this height if it is larger.
+   * @defaultValue 1080
+   */
+  maxHeight?: number;
+  /**
+   * The quality of the image returned.
+   * Percentile value between 0 (lowest quality/highest compression) and 1 (highest quality/lowest compression).
+   * @defaultValue 0.9
+   */
+  quality?: number;
+}
+
+export interface CameraMediaResponse {
+  /** The image data as base64 string. */
+  base64: string;
+  /** The width of the image in pixels. */
+  width: number;
+  /** The height of the image in pixels. */
+  height: number;
+  /** The file size of the image in bytes. */
+  fileSize: number;
+  /** The MIME type of the image. */
+  type: string;
+}
+
+export interface CameraApiContent {
+  /**
+   * Launch the device's camera to take a photo.
+   *
+   * @param options the options for the camera media
+   * @returns a promise that resolves with the captured image and metadata
+   * @throws {Error} when camera permission is denied by the user
+   * @throws {Error} when the user cancels the photo capture
+   * @throws {Error} when the device has no available camera
+   * @throws {Error} when the camera API is already in use
+   */
+  takePhoto: (options?: CameraMediaOptions) => Promise<CameraMediaResponse>;
+}
+
+export interface CameraApi {
+  camera: CameraApiContent;
+}
--- a/.changeset/slimy-foxes-hide.md
+++ b/.changeset/slimy-foxes-hide.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Add Camera API types
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api.ts
@@ -15,6 +15,13 @@ export type {StandardApi} from './api/standard/standard-api';
 export type {ActionTargetApi} from './api/action-target-api/action-target-api';
 
 export type {
+  CameraApi,
+  CameraApiContent,
+  CameraMediaOptions,
+  CameraMediaResponse,
+} from './api/camera-api/camera-api';
+
+export type {
   ConnectivityStateSeverity,
   ConnectivityState,
   ConnectivityApiContent,
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/standard/standard-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/standard/standard-api.ts
@@ -1,3 +1,4 @@
+import {CameraApi} from '../camera-api/camera-api';
 import {ConnectivityApi} from '../connectivity-api/connectivity-api';
 import {DeviceApi} from '../device-api/device-api';
 import {LocaleApi} from '../locale-api/locale-api';
@@ -20,4 +21,5 @@ export type StandardApi<T> = {[key: string]: any} & {
   DeviceApi &
   ConnectivityApi &
   StorageApi &
-  PinPadApi;
+  PinPadApi &
+  CameraApi;


andy-chhuon:
Test the snapshot by updating your `package.json` with the newly published version:
```json
"@shopify/ui-extensions": "0.0.0-snapshot-20251125223311"
```

packages/ui-extensions/src/surfaces/point-of-sale/api/action-target-api/action-target-api.ts
```diff
@@ -1,7 +1,9 @@
 import {ScannerApi} from '../scanner-api/scanner-api';
 import {StandardApi} from '../standard/standard-api';
+import {CameraApi} from '../camera-api/camera-api';
 
 export type ActionTargetApi<T> = {[key: string]: any} & {
```
vctrchu, line 4: Maybe I missed it but we only want to allow the camera to be triggered from action targets? Is there a reason why we don't want the api accessible from tiles or blocks?

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
```diff
@@ -0,0 +1,51 @@
+export interface CameraMediaOptions {
+  /**
+   * The camera that will be active when the camera interface first opens.
+   * - `'user'`: The front-facing camera
+   * - `'environment'`: The rear-facing camera
+   * @defaultValue 'environment'
+   */
+  facingMode?: 'user' | 'environment';
+  /**
+   * The maximum width (0 to 1080) of the image in pixels. Resizes the image to this width if it is larger.
```
vctrchu: Should it be `(1 to 1080)`

What happens when users put a value of 0?

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
```diff
@@ -0,0 +1,51 @@
...
+   * @defaultValue 1080
+   */
+  maxWidth?: number;
+  /**
+   * The maximum height (0 to 1080) of the image in pixels. Resizes the image to this height if it is larger.
```
vctrchu: Same here

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
```diff
@@ -0,0 +1,51 @@
...
+   * @defaultValue 1080
+   */
+  maxHeight?: number;
+  /**
+   * The quality of the image returned. Percentile value between 0 and 1.
```
vctrchu: Same here, whats 0 quality do?

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
```diff
@@ -0,0 +1,51 @@
...
+  type: string;
+}
+
+export interface CameraApiContent {
+  /**
+   * Launch the device's camera to take a photo.
+   *
+   * @param options the options for the camera media.
+   * @returns Promise<CameraMediaResponse> that resolves when the POS has necessary permissions to access the camera and the media is captured.
+   */
```
vctrchu, line 50: Should we be clear on the promise reject reasons? Maybe something like:
```
 * @throws {Error} when camera permission is denied by the user
 * @throws {Error} when the user cancels the photo capture
 * @throws {Error} when the device has no available camera
```

packages/ui-extensions/src/surfaces/point-of-sale/api/action-target-api/action-target-api.ts
andy-chhuon: Good catch! We cherry picked the wrong PR, meant to be [this](https://github.com/Shopify/ui-extensions/pull/3570/files) one. Fixed

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
andy-chhuon: Yes, we set the minimum dimensions to 1 in [POS](https://github.com/shop/world/pull/289063/files#diff-8019fcea783bc1c45ca8180b59a4bf6fef97e660e5f7acd910dab071fdadfc82R24). The camera library errors if we set it to 0. This needs to be updated here too

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
andy-chhuon: Quality 0 outputs a low quality picture and is supported, so we do expose it as an option

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export interface CameraApiContent {
+  /**
+   * Launch the device's camera to take a photo.
+   *
+   * @param options the options for the camera media.
+   * @returns Promise<CameraMediaResponse> that resolves when the POS has necessary permissions to access the camera and the media is captured.
+   * @throws {Error} when camera permission is denied by the user
+   * @throws {Error} when the user cancels the photo capture
+   * @throws {Error} when the device has no available camera (e.g. POS Go)
```
vctrchu:
```suggestion
   * @throws {Error} when the device has no available camera
```

vctrchu: I think your first part is clear enough and we don't sell POS Gos anymore

vctrchu (approved):
LGTM, just the one comment about removing the POS GO reference

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
fatbattk: updated

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
```diff
@@ -0,0 +1,55 @@
...
+   * @defaultValue 1080
+   */
+  maxWidth?: number;
+  /**
+   * The maximum height (1 to 1920) of the image in pixels. Resizes the image to this height if it is larger.
+   * @defaultValue 1080
+   */
+  maxHeight?: number;
+  /**
+   * The quality of the image returned. Percentile value between 0 and 1.
```
BrittaEvansFenton: Is the 0-1 based on some standards or is it made up for this? 
Piggybacking off what Victor was asking I think we need to give some more context of that 0 and what 1 gives.

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export interface CameraMediaResponse {
+  /** The image data as base64 string. */
+  base64: string;
+  /** The width of the image in pixels. */
+  width: number;
+  /** The height of the image in pixels. */
+  height: number;
+  /** The file size of the image in bytes. */
+  fileSize: number;
```
BrittaEvansFenton, line 35: So this fileSize would change depending on the quality number correct?

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
```diff
@@ -0,0 +1,55 @@
...
+  /** The MIME type of the image. */
+  type: string;
```
BrittaEvansFenton, line 37: Should we have this be an enum if it is MIME type? 
And what is the value of this?

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
fatbattk: correct, as well as image dimensions.

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
fatbattk: Value for type here would be, `image/jpeg`, `image/png`, etc.
Using `string` is preferable to a hardcoded enum since MIME types can evolve and the OS/libraries may support additional types without us needing to update an `enum`.

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
fatbattk: 0-1 range (percentile/percentage) is common way to represent compression quality levels. we can add (0 = lowest quality, 1 = highest quality) here so it's clearer.

andy-chhuon: Yes, this is the standard for image compression APIs on Native surfaces like IOS (see [here](https://developer.apple.com/documentation/uikit/uiimage/jpegdata(compressionquality:)))

packages/ui-extensions/src/surfaces/point-of-sale/api/camera-api/camera-api.ts
BrittaEvansFenton: unless we are testing all use cases I think this is not a great idea. 
It is easier to add types than to try to remove cases that don't work later. 
We have a similar MIME type on embed.

fatbattk: this is the MIME type value for the response and since the device took the photo, it'll always be the correct MIME value. maintaining `enum` would be toil here.

andy-chhuon (approved):

PR #3601, opened 2025-11-28
Remove deprecated `ActionItem` component doc

mcvinci:
## This PR: 
- Removes `ActionItem` from the 2024-10 POS UI extensions docs, as this is a deprecated component (and only exists in 2024-07 and 2024-04)
- Relates to https://github.com/Shopify/shopify-dev/pull/65462


johndcollett (approved):
LGTM. Are we good with this change coming into effect next time they generate their docs? Or did we want to generate them now?

mcvinci:
> LGTM. Are we good with this change coming into effect next time they generate their docs? Or did we want to generate them now?

Oh, I just generated the docs in https://github.com/Shopify/shopify-dev/pull/65462, so we can have it come into effect. 😄

PR #3632, opened 2025-12-05
Remove POS camera scanner examples

vctrchu:
Part of https://github.com/shop/issues-retail/issues/21393
### Background

This component is not available in 2025-10. We shouldn't have example for it.


aaronschubert0 (approved):
:blush:

PR #3641, opened 2025-12-10
Update API and add observables for discount function settings

jonathanhamel4:
### Background

Related to https://github.com/Shopify/ui-api-design/pull/1332

Adds a subscribable API for Discounts Functions Settings extension target. 

- PurchaseType
- DiscountMethod (readonly)
- DiscountClasses
- RecurringCycleLimit

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
+++ b/packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
@@ -1,3 +1,8 @@
+import type {
+  ReadonlySignalLike,
+  UpdateSignalFunction,
+} from '../../../../shared';
+
 interface Metafield {
   description?: string;
   id: string;
@@ -7,23 +12,56 @@ interface Metafield {
   type: string;
 }
 
-export enum DiscountClass {
-  Product = 'PRODUCT',
-  Order = 'ORDER',
-  Shipping = 'SHIPPING',
-}
+type DiscountClass = 'product' | 'order' | 'shipping';
 
-interface Discount {
-  /**
-   * the discount's gid
-   */
-  id: string;
-}
+type DiscountMethod = 'automatic' | 'code';
+
+type PurchaseType = 'one_time_purchase' | 'subscription' | 'both';
 
 /**
  * The object that exposes the validation with its settings.
  */
 export interface DiscountFunctionSettingsData {
-  id: Discount;
+  /**
+   * The unique identifier for the discount.
+   */
+  id: string;
+  /**
+   * The discount metafields.
+   */
   metafields: Metafield[];
+}
+
+/**
+ * Reactive Api for managing discount function configuration.
+ */
+export interface DiscountsApi {
+  /**
+   * A signal that contains the discount classes.
+   */
+  discountClasses: ReadonlySignalLike<DiscountClass[]>;
+  /**
+   * A function that updates the discount classes.
+   */
+  updateDiscountClasses: UpdateSignalFunction<DiscountClass[]>;
+  /**
+   * A signal that contains the discount method.
+   */
+  discountMethod: ReadonlySignalLike<DiscountMethod>;
+  /**
+   * A signal that contains the purchase type.
+   */
+  purchaseType: ReadonlySignalLike<PurchaseType>;
+  /**
+   * A function that updates the purchase type.
+   */
+  updatePurchaseType: UpdateSignalFunction<PurchaseType>;
+  /**
+   * A signal that contains the recurring cycle limit for subscriptions purchases.
+   */
+  recurringCycleLimit: ReadonlySignalLike<number | null | undefined>;
+  /**
+   * A function that updates the recurring cycle limit for subscriptions purchases.
+   */
+  updateRecurringCycleLimit: UpdateSignalFunction<number | null | undefined>;
 }
--- a/.changeset/wet-cycles-punch.md
+++ b/.changeset/wet-cycles-punch.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Added subscribable discounts api. Update the type for `data.id` to string to fix a previously incorrect type.
--- a/packages/ui-extensions/src/shared.ts
+++ b/packages/ui-extensions/src/shared.ts
@@ -952,3 +952,38 @@ export interface ReadonlySignalLike<T> {
    */
   subscribe(fn: (value: T) => void): () => void;
 }
+
+/**
+ * A result type that indicates the success or failure of an operation.
+ */
+type Result<T> =
+  | {success: true; value: T}
+  | {success: false; errors: ValidationError[]};
+
+/**
+ * A validation error object that is returned when an operation fails.
+ */
+interface ValidationError {
+  type: 'error';
+  /**
+   * A message describing the error.
+   */
+  message: string;
+  /**
+   * A code identifier for the error.
+   */
+  code: string;
+  /**
+   * Field-level validation issues
+   */
+  issues?: {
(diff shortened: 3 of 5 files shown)


devisscher (approved):

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
```diff
@@ -7,23 +7,69 @@ interface Metafield {
...
-  id: Discount;
+  id: string;
+  discountClasses: DiscountClass[];
   metafields: Metafield[];
 }
+
+/**
+ * Represents a reactive signal interface that provides both immediate value access and subscription-based updates. Enables real-time synchronization with changing data through the observer pattern.
+ */
+export interface ReadonlySignalLike<T> {
```
vividviolet: Let's pull this in from here https://github.com/Shopify/ui-extensions/blob/7b5b4c7e8f8e545c97d889d4e1ad41478c463ec0/packages/ui-extensions/src/shared.ts#L945.

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
devisscher, line 1: Leaving a couple comments from the ui-api-design team:
- Let's use unions with lower case values: `'product'|'order'|'shipping'`, same goes for method and purchaseType.
- For the result we'll return errors (plural) and `ValidationError[]`
- discountClasses will become accessible in the subscribables, but not initial data

jonathanhamel4: `type PurchaseType = 'one_time_purchase' | 'subscription' | 'both';`

Camel case? or PascalCase?

jonathanhamel4: Updated 👍

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
```diff
@@ -7,23 +7,57 @@ interface Metafield {
...
+   * The current value of the field.
+   */
+  readonly value: T;
+  /**
+   * Subscribes to field changes and calls the provided function whenever the field updates. Returns an unsubscribe function to clean up the subscription.
+   */
+  subscribe(fn: (value: T) => void): () => void;
+}
+
+type UpdateFunction<T> = (value: T) => Result<T>;
```
vividviolet: Can you put this into ui-extensions/src/shared.ts as well and name it `UpdateSignalFunction`?

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
```diff
@@ -7,23 +7,57 @@ interface Metafield {
   type: string;
 }
 
-export enum DiscountClass {
-  Product = 'PRODUCT',
-  Order = 'ORDER',
-  Shipping = 'SHIPPING',
-}
+type DiscountClass = 'product' | 'order' | 'shipping';
```
vividviolet, line 15: We need to call this out in the changset as a breaking change. We will also need to ensure backwards compatibility so that only 2026-01 and up will receive the lowercased enum.

jonathanhamel4: Do we need to document this as a breaking change? This type was defined but never used. Like the input never referenced that type 🤔 It wasn't flagged as unused in CI because it is exported I assume.

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
jonathanhamel4: Done

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
```diff
@@ -7,23 +12,26 @@ interface Metafield {
...
 export interface DiscountFunctionSettingsData {
-  id: Discount;
+  id: string;
   metafields: Metafield[];
 }
+
+export interface SubscribableDiscountFields {
+  discountClasses: ReadonlySignalLike<DiscountClass[]>;
+  updateDiscountClasses: UpdateSignalFunction<DiscountClass[]>;
+  discountMethod: ReadonlySignalLike<DiscountMethod>;
```
jonathanhamel4, line 50: discountMethod is readonly.

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
```diff
@@ -7,23 +12,26 @@ interface Metafield {
...
-}
+type DiscountMethod = 'automatic' | 'code';
+
+type PurchaseType = 'one_time_purchase' | 'subscription' | 'both';
 
 /**
  * The object that exposes the validation with its settings.
  */
 export interface DiscountFunctionSettingsData {
-  id: Discount;
```
jonathanhamel4, line 27: This was actually wrong in the past. We never sent the `id` as an object with a single `id` property.

jonathanhamel4:
@vividviolet ready for another review. Added a comment about the discountClasses type.

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
vividviolet: Ah ok, we've never exposed this in the api, just added the enum so it's ok to not be a breaking change

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
vividviolet: Can we please patch the previous release too?

vividviolet (approved):

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
jonathanhamel4: It has been wrong for more than one release. I can patch all of them, do I just have to merge the branch to the non-rc one? Like 2025-10, etc. Anything else I need to do?

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/launch-options.ts
```diff
@@ -7,23 +12,56 @@ interface Metafield {
...
+   */
+  updatePurchaseType: UpdateSignalFunction<PurchaseType>;
+  /**
+   * A signal that contains the recurring cycle limit for subscriptions purchases.
+   */
+  recurringCycleLimit: ReadonlySignalLike<number | null | undefined>;
+  /**
+   * A function that updates the recurring cycle limit for subscriptions purchases.
+   */
+  updateRecurringCycleLimit: UpdateSignalFunction<number | null | undefined>;
```
loic-d, line 66: we might need to pull these out from `2026-01`:

- outstanding bugs
- BED changes required to honour the provided purchase type. more specifically, we need to rollout https://experiments.shopify.com/flags/f_discount_purchase_type_filtering and figure out a backfill strategy to preserve the existing behaviour

cc/ @devisscher @jonathanhamel4 @dpwolf

jonathanhamel4: 👍 The change has been merged. https://github.com/Shopify/ui-extensions/pull/3695

PR #3649, opened 2025-12-15
Add intents to customer account extension api

andrewmcgov:
### Background

Closes https://github.com/shop/issues-extensions-api/issues/18

### Solution

Add type support for intents on the customer account extension Standard API.

### 🎩

See https://github.com/Shopify/shopify-dev/pull/67158 for testing instructions.

### DIFF
--- a/packages/ui-extensions/docs/surfaces/customer-account/reference/apis/intents.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/customer-account/reference/apis/intents.doc.ts
@@ -4,21 +4,14 @@ const data: ReferenceEntityTemplateSchema = {
   name: 'Intents',
   overviewPreviewDescription:
     'The API for invoking Shopify intents to request workflows.',
-  description: `Entry point for Shopify intents. Intents pair an \`action\` (verb) with a resource \`type\` and optional \`value\` and \`data\` to request a workflow.`,
+  description: `The Intents API provides a way to invoke existing customer account workflows for managing buyer information.`,
   isVisualComponent: false,
   category: 'APIs',
   type: 'API',
-  definitions: [
-    {
-      title: 'Intents',
-      description: 'Intents API for invoking Shopify workflows.',
-      type: 'Intents',
-    },
-  ],
   defaultExample: {
     description: '',
     codeblock: {
-      title: 'Extension.jsx',
+      title: 'Replace payment method',
       tabs: [
         {
           code: '../examples/apis/intents.example.jsx',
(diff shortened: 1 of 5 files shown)


packages/ui-extensions/docs/surfaces/customer-account/reference/apis/intents.doc.ts
```diff
@@ -0,0 +1,100 @@
...
+|--------|------|-------|------|
+| \`open\` | \`shopify/SubscriptionContract\` | \`gid://shopify/SubscriptionContract/{id}\` | \`{ field: 'paymentMethod' }\` |`,
+      type: 'Intents',
+    },
+    {
+      title: 'IntentAction',
+      description: `Supported actions that can be performed on resources.
+- \`create\`: Opens a creation workflow for a new resource
+- \`edit\`: Opens an editing workflow for an existing resource (requires \`value\` parameter)
+- \`open\`: Opens a workflow for an existing resource (requires \`value\` parameter)`,
```
andrewmcgov: I think we should only have `open` here - right now we do not support edit

packages/ui-extensions/docs/surfaces/customer-account/reference/apis/intents.doc.ts
```diff
@@ -0,0 +1,100 @@
...
+    },
+    {
+      title: 'IntentResponse',
+      description: `Response object returned when the intent workflow completes.`,
+      type: 'IntentResponse',
+    },
+  ],
+  examples: {
+    description: 'Intents for each Shopify resource type',
+    exampleGroups: [
```
andrewmcgov: We might not want to use `exampleGroups` here, its creating a group dropdown that we probably dont want: 

Image

packages/ui-extensions/docs/surfaces/customer-account/reference/apis/intents.doc.ts
```diff
@@ -0,0 +1,100 @@
...
+      {
+        title: 'Subscription Contract',
```
andrewmcgov: I wonder if we should add another example that shows the string url syntax?

```js
const activity = await shopify.intents.invoke(
  'open:shopify/SubscriptionContract,gid://shopify/SubscriptionContract/123?field=paymentMethod',
);
```

packages/ui-extensions/docs/surfaces/customer-account/reference/examples/apis/intents.example.jsx
```diff
@@ -0,0 +1,35 @@
+import '@shopify/ui-extensions/preact';
+import {render} from 'preact';
+
+export default async () => {
+  render(<Extension />, document.body);
+};
+
+function Extension() {
+  async function handleReplacePaymentMethod() {
+    const activity = await intents.invoke({
```
andrewmcgov: Need to get this from the shopify global:


```suggestion
    const activity = await shopify.intents.invoke({
```

graeme-shopify (approved):

lrsterrett (approved):

PR #3650, opened 2025-12-15
Add OIDC to rc branch

andy-chhuon:
### Background

Migrate from npm to OIDC according to [docs](https://vault.shopify.io/page/Migrating-to-npm-OIDC-Trusted-Publishers~sl6d.md#step-2) to avoid 404 error

Also see, https://github.com/Shopify/ui-extensions/pull/3650#issuecomment-3658183889

image


vctrchu (approved):

fatbattk:
## Update
- Modified to centralize to `deploy.yml` because npm OIDC only allows one publish flow file.
  - Works for `deploy-rc` consolidation but may need different strategy with the snap shots and other flows.
- Safe to merge and test as publishing to npm does not work now.

andy-chhuon:
Thanks @fatbattk, makes sense! 

context on pr:
[slack](https://shopify.slack.com/archives/C0683EBQ55M/p1765829696670009)

context on why only one publish file:
[Slack](https://shopify.slack.com/archives/C01MXHNTT4Z/p1762204640029009?thread_ts=1762202727.817039&cid=C01MXHNTT4Z) & OIDC [docs](https://vault.shopify.io/page/Migrating-to-npm-OIDC-Trusted-Publishers~sl6d.md#error-quotpackage-not-foundquot-404---for-reusable-workflows-only) (control-F for '404' - vault doesn't link correctly)

Snapit won't work as it currently does not support [OIDC](https://github.com/Shopify/snapit/issues/50) (currently doesn't work either with NPM token)

.github/workflows/deploy.yml
```diff
@@ -11,40 +12,44 @@ on:
...
 
       - uses: ./.github/workflows/actions/prepare
 
+      - name: Update npm to latest
+        run: npm install -g npm@latest
+
       - id: changesets
         name: Create release Pull Request or publish to NPM
-        uses: changesets/action@06245a4e0a36c064a573d4150030f5ec548e4fcc # v1.4.10
+        uses: changesets/action@v1 # Must use latest version for OIDC
```
aaronschubert0: @andy-chhuon do you know what version added support for OIDC or is pinning it to a commit hash not compatible with OIDC? It would be good to keep pinned to a specific version so we don't automatically opt-into new versions.

andy-chhuon: Good point, this was what is recommended in the docs but pinning it to a version makes sense. I pinned it to the [latest](https://github.com/changesets/action/commit/e0145edc7d9d8679003495b11f87bd8ef63c0cba) version just in case and it's a version used throughout [shopify](https://github.com/search?q=org%3AShopify%20e0145edc7d9d8679003495b11f87bd8ef63c0cba&type=code) too

image

aaronschubert0: Great, thanks @andy-chhuon ! Surprisingly the intro image does have a pinned version. Could you ping the team that owns the page about the discrepancy and ask them to update it?

aaronschubert0 (approved):

.github/workflows/deploy.yml
andy-chhuon: Ah interesting, will do!

PR #3661, opened 2025-12-16
Add EmptyState docs

cappslock:
Adds EmptyState docs. Partially resolves https://github.com/shop/issues-retail/issues/21717.

This includes some unrelated changes to tab due to upstream changes in the definition file. It's safer to commit these than ignore them.


BrittaEvansFenton:
Did you mean to add the Tab stuff in here too? There is just no mention of it in the description. Please either remove or add into Descriptions

cappslock:
@BrittaEvansFenton it's generated output that's true to the source, so no, not intentional, but would also require just spinning up a second PR with those changes, which would block this PR. I'd honestly rather just get them in at once given the timeframe we're dealing with.

BrittaEvansFenton:
Absolutely, just edit the description so at least when we go back to figure out what we did later we have some context 😅

BrittaEvansFenton (approved):

PR #3676, opened 2025-12-17
Update Toast API docs

cappslock:
Remove the `duration` option, as it is no longer in use by POS. Partially resolves https://github.com/shop/issues-retail/issues/21718

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/reference/apis/toast-api.doc.ts
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/reference/apis/toast-api.doc.ts
@@ -7,7 +7,7 @@ const generateJsxCodeBlockForToastApi = (title: string, fileName: string) =>
 const data: ReferenceEntityTemplateSchema = {
   name: 'Toast API',
   description:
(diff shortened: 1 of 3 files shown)


armaan-ghosh (approved):
🚀 👍

packages/ui-extensions/docs/surfaces/point-of-sale/reference/apis/toast-api.doc.ts
```diff
@@ -53,7 +52,7 @@ Toast content is limited to plain text and can't include rich formatting, links,
           'show',
         ),
         description:
-          "Display a toast notification from a tile. This example demonstrates using `shopify.toast.show()` to display a brief, non-intrusive message that automatically disappears after a specified duration. This is useful for confirmations, status updates, or success messages that don't require user interaction.",
+          "Display a toast notification from a tile. This example demonstrates using `shopify.toast.show()` to display a brief, non-intrusive message that automatically disappears. This is useful for confirmations, status updates, or success messages that don't require user interaction.",
```
BrittaEvansFenton, line 55: do we want to point out that it is displayed for 5 seconds?

something like this:


`his example demonstrates using `shopify.toast.show()` to display a brief, 5 second long, non-intrusive message that automatically disappears.`

cappslock: We can be explicit about it automatically disappearing, but IMO Toasts are generally understood to be temporary nudges, so I don't think it's adding much. I would prefer to not be explicit about duration as we're otherwise hiding that from the user and it may be subject to change.

PR #3683, opened 2025-12-18
Add Tabs and related components to StandardComponents

cappslock:


packages/ui-extensions/src/surfaces/point-of-sale/components/targets/StandardComponents.ts
```diff
@@ -31,6 +31,10 @@ export type StandardComponents =
   | 'Section'
   | 'Spinner'
   | 'Stack'
```
BrittaEvansFenton, line 33: Is there one for each component? Switch too?

cappslock: Good call, we need Switch as well. Added.

BrittaEvansFenton (approved):

PR #3695, opened 2026-01-05
remove extra fields on Discount API

jonathanhamel4:
### Background

Removes fields that are not yet supported for discount api 2026-01. Those were not added yet to any release. 

It is based on this comment https://github.com/Shopify/ui-extensions/pull/3641/files#r2662549011

### DIFF
--- a/.changeset/eager-dogs-carry.md
+++ b/.changeset/eager-dogs-carry.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+Remove unsupported fields for the Discount API. Those fields will be added in a later API.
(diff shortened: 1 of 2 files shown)


.changeset/eager-dogs-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+Remove unsupported fields for the Discount API. Those fields will be added in a later API.
```
jonathanhamel4, line 5: I made it a `patch` because those fields have merged, but it has not published yet. It was merged before the break.

devisscher (approved):

jonathanhamel4:
@vividviolet , would we be okay to merge this change for 2026-01 ?

vividviolet:
Yes, let's remove these if they will not be yet be supported

PR #3731, opened 2026-01-21
Stop gitignoring generated docs JSON files

billfienberg:
## Summary

Track generated documentation JSON files in version control instead of gitignoring them.

## Changes

- Remove `packages/ui-extensions/docs/**/generated` from `.gitignore`
- Add generated documentation files for Admin Extensions and App Home surfaces

## Why

Previously, generated docs were excluded from version control and had to be regenerated during the sync process. By tracking these files directly, downstream consumers can sync documentation without needing to run build commands in this repository.

This simplifies the documentation pipeline and makes the sync process more reliable.

## Test plan

- [x] Verified generated JSON files are valid
- [x] Confirmed files were previously gitignored and are now tracked


packages/ui-extensions/docs/surfaces/admin/generated/admin_extensions/2025-10/generated_docs_data.json
jplhomer, line 1: qq: is there a reason this is nested in a `2025-10` folder? Since versioning strategy is branch-based, does it make more sense to drop that extra segment of the path and store in `admin_extensions` instead?

billfienberg: > [!NOTE]
> The following summary was provided by Claude. 

---

The version folder exists for shopify-dev integration. When these docs get synced to shopify-dev, the copy operation preserves the folder structure:

```
ui-extensions: admin_extensions/2025-10/generated_docs_data.json
      ↓
shopify-dev: db/data/docs/templated_apis/admin_extensions/2025-10/
```

shopify-dev serves documentation for multiple API versions simultaneously (2025-07, 2025-10, unstable, etc.), so having the version embedded in the path allows a simple recursive copy without needing to transform paths based on which branch the files came from.

That said, you raise a fair point—within a single branch's context, the version folder is somewhat redundant since the branch only generates one version's docs. If we wanted to flatten it, we'd need the sync process to handle placing files in the correct version folder based on the source branch. The current design trades that complexity for a straightforward copy operation.

.prettierignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 packages/*/build
 packages/web-pixels-extension/src/schemas/pixel-events.jtd.json
-packages/ui-extensions/src/surfaces/checkout/**/*.d.ts
\ No newline at end of file
+packages/ui-extensions/src/surfaces/checkout/**/*.d.ts
+packages/ui-extensions/docs/**/generated/**/*.json
```
billfienberg, line 5: An earlier run [failed](https://github.com/Shopify/ui-extensions/actions/runs/21225932137/job/61072775392) because of linting. 

@jplhomer, do you think there'd be any harm in running prettier on the generated files?

jplhomer: @billfienberg I think it'd probably be really slow and unnecessary so ignoring is the right move here 👍

packages/ui-extensions/docs/surfaces/admin/generated/admin_extensions/2025-10/generated_docs_data.json
jplhomer: Ahh ok. Ideally we can move to the path where shopify-dev consumes these docs with its daily job. I'll have to dig into this, but I believe the `generated_path` defined in `config.js` on expects the JSON files to exist at that level and not in any nested folders. But I also understand the impacts of changing this location on your other script assumptions in `extensibility` 🤔

billfienberg:
Closing due to age

PR #3774, opened 2026-01-30
Add context-specific docs system for Admin vs App Home

mcvinci:
## This PR: 
- **Prototypes a context-aware docs schema**: Extends `SharedReferenceEntityTemplateSchema` to support a `contexts` field, allowing writers to define different component descriptions for different surfaces, where necessary
- **Integrates context resolution into the build pipeline**: Modifies `build-docs.mjs` to automatically resolve context-specific overrides during doc generation, ensuring Admin UI docs (/docs/api/admin-extensions) and App Home docs (/docs/api/app-home) display appropriate descriptions for each surface
- **Updates the Badge component as an example**: Demonstrates how to define context-specific descriptions

### Tophatting

Check out this branch, and run `yarn docs:admin 2026-01`. This will generate the docs for App Home and Admin UI. Run `dev s` in `shopify-dev` to view the docs. 

You can also review other UI extensions references (like Checkout) to see that it uses the default "shared" description.

Screenshot 2026-01-29 at 6 16 19 PM

Screenshot 2026-01-29 at 6 16 27 PM

Screenshot 2026-01-29 at 6 16 32 PM

### DIFF
--- a/packages/ui-extensions/docs/surfaces/admin/build-docs.mjs
+++ b/packages/ui-extensions/docs/surfaces/admin/build-docs.mjs
@@ -250,8 +250,34 @@ const templates = {
   }),
 };
 
+const resolveContextualDescription = (entry, isExtensions) => {
+  // Determine the deployment context based on which docs we're building
+  const context = isExtensions ? 'admin' : 'app-home';
+  
+  // If the entry has context-specific overrides, resolve them
+  if (entry.contexts && entry.contexts[context]) {
+    const contextOverride = entry.contexts[context];
+    
+    // Apply context-specific overrides
+    if (contextOverride.description) {
+      entry.description = contextOverride.description;
+    }
+    if (contextOverride.related) {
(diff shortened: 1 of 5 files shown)


johndcollett (commented):
I think this makes sense and would be replaced by partials when we do the migration from .doc.ts to MDX. Does that match up with what you were thinking of in a post migration world?

mcvinci:
> I think this makes sense and would be replaced by partials when we do the migration from .doc.ts to MDX. Does that match up with what you were thinking of in a post migration world?

@johndcollett Yes, and that's actually a really good point. Given that we're moving in that direction relatively soon, I wonder if it's worthwhile to go down this path at all in the `ui-extensions` repo.

@timtrevor-shopify @chadwcarlson Maybe we can quickly touch base this week on this. I have a possible plan! ❤️

mcvinci:
Closing in favour of: https://github.com/Shopify/shopify-dev/pull/67784

PR #3788, opened 2026-02-02
[2026-01] Update metafield docs

avocadomayo:
### Background

Add deprecation warning and guidance for `MetafieldRemoveChange` and `MetafieldUpdateChange` in checkout metafields API.

This PR will kick off a patch to `2026-01`. We typically do not do a release for doc changes, but making an exception here since we want developers to benefit from annotations while referencing types.

### Checklist

### DIFF
--- a/.changeset/giant-trains-repeat.md
+++ b/.changeset/giant-trains-repeat.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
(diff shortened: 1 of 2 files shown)


jamesvidler (approved):

lrsterrett (approved):
Looks good, my only thought would be to mention that these will be removed in api version 2026-04 since we know that already.

SaifBenChahed (approved):

PR #3789, opened 2026-02-02
chore: update shopify-dev asset path

majd-shopify:
### Background

Assets in shopify-dev are no longer under `react-app` and have moved to under `content-v2`

See https://github.com/Shopify/shopify-dev/pull/67138 for more context

This automated PR https://github.com/Shopify/shopify-dev/pull/67623 still attempts to copy assets into the old path, `build-docs.mjs` is called from the [extensibility](https://github.com/Shopify/extensibility/blob/2ee838e59f6c462fadcffb78c54847c5c17ace27/scripts/sync-docs.js#L99) repo and the path in `build-docs.mjs` is incorrect

### Solution

Update the path to the new asset path in shopify-dev

Also, is this something I will have to do for all the version branches in this repo? or will this suffice?

### 🎩

- review PRs mentioned and change

### Checklist

- [] I have updated relevant documentation (is there any documentation for this?)

### DIFF
--- a/packages/ui-extensions/docs/surfaces/admin/build-docs.mjs
+++ b/packages/ui-extensions/docs/surfaces/admin/build-docs.mjs
@@ -511,7 +511,7 @@ try {
     path.join(docsPath, 'screenshots'),
     path.join(
       shopifyDevPath,
-      'react-app/public/images/templated-apis-screenshots/admin',
+      'content-v2/assets/images/templated-apis-screenshots/admin',
     ),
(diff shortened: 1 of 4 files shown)


johndcollett (commented):
this is correct but this path should also be defined for checkout, customer accounts, and pos in a similar build-docs.mjs

We should also update these for the other versions in the ui-extensions repo... sadly

johndcollett (approved):
I approve these changes

PR #3823, opened 2026-02-05
[Target APIs content]: Examples and descriptions for 2025-07

SteveSill:
# Admin UI Extensions - Examples 2025-07

## Summary

This PR adds comprehensive code examples for all Admin UI Extension Target APIs in the 2025-10 version. All examples follow the modern Preact pattern established in POS UI Extensions, using `.jsx` files with the global `shopify` object.

### New Examples Created (36 files)

Added working code examples for 8 APIs that previously had none:

**Core APIs:**
- Action Extension API (3 examples)
- Block Extension API (3 examples)
- Print Action Extension API (3 examples)
- Standard API (3 examples)

**Contextual APIs:**
- Customer Segment Template Extension API (3 examples)
- Discount Function Settings API (3 examples)
- Validation Settings API (3 examples)
- Order Routing Rule API (3 examples)
- Product Details Configuration API (3 examples)
- Product Variant Details Configuration API (3 examples)
- Purchase Options Card Configuration API (3 examples)

**Utility APIs:**
- Should Render API (3 examples)
image

PR #3826, opened 2026-02-05
[Target APIs content]: Examples and descriptions for 2024-10

SteveSill:
# Admin UI Extensions - Examples 2024-10

## Summary

This PR adds comprehensive code examples for all Admin UI Extension Target APIs in the 2025-10 version. All examples follow the modern Preact pattern established in POS UI Extensions, using `.jsx` files with the global `shopify` object.

image

image

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/api/action/action.doc.ts
+++ b/packages/ui-extensions/src/surfaces/admin/api/action/action.doc.ts
@@ -8,6 +8,25 @@ const data: ReferenceEntityTemplateSchema = {
   type: 'API',
   requires:
     'the [`AdminAction`](/docs/api/admin-extensions/{API_VERSION}/polaris-web-components/settings-and-templates/adminaction) component.',
+  defaultExample: {
+    description:
+      'Send selected product IDs to your backend for bulk processing. This example shows how to map selected items, make an authenticated API call, and close the modal when the operation completes.',
+    codeblock: {
+      title: 'Process selected products',
+      tabs: [
+        {
+          title: 'React',
+          code: './examples/process-selected-products.tsx',
+          language: 'tsx',
+        },
+        {
+          title: 'TS',
+          code: './examples/process-selected-products.ts',
+          language: 'ts',
+        },
+      ],
+    },
+  },
   definitions: [
     {
       title: 'ActionExtensionApi',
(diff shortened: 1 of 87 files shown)


packages/ui-extensions/src/surfaces/admin/api/picker/examples/direct-api.ts
```diff
@@ -0,0 +1,42 @@
+import {extension, Button} from '@shopify/ui-extensions/admin';
```
mcvinci: I don't think we need all these examples in this version, given the Picker API isn't available?

packages/ui-extensions/src/surfaces/admin/api/discount-function-settings/examples/configure-discount-threshold.tsx
```diff
@@ -0,0 +1,47 @@
+import React, {useState} from 'react';
+import {
+  reactExtension,
+  useApi,
+  TextField,
+  Button,
+  Banner,
+} from '@shopify/ui-extensions-react/admin';
+
+const ConfigureDiscountThreshold = () => {
```
mcvinci: Same here. I don't think we need these examples in this version, given the Discount Function Settings API isn't available in this version.

packages/ui-extensions/src/surfaces/admin/api/purchase-options-card/examples/manage-subscription.ts
```diff
@@ -0,0 +1,53 @@
+import {extension, Text, Button, Banner} from '@shopify/ui-extensions/admin';
```
mcvinci: Same here. This API isn't available in this version, so we can remove the examples.

packages/ui-extensions/src/surfaces/admin/api/should-render/examples/bulk-selection-check.ts
```diff
@@ -0,0 +1,13 @@
+import {extension} from '@shopify/ui-extensions/admin';
```
mcvinci: Should Render API isn't available in this version, so we should remove the example files.

mcvinci (approved):
There are a handful of example files that we can actually clean up from this PR, given that some of the APIs don't exist in this version. I've added some comments about that below. 

🚀 🚀 Also, approving, pending these suggestions are considered for this version as well: https://github.com/Shopify/ui-extensions/pull/3823#pullrequestreview-3759861120

PR #3876, opened 2026-02-07
fix broken script for 2023-04

laurelthorburn:
**Admin docs build fixes**

- **generate-docs “Unable to find code file”**  
  The Custom Protocols section used inline `code` strings; generate-docs treats `code` as file paths. Added five example files under `staticPages/examples/` (link-shopify-admin, fetch-shopify-admin, link-app, link-extension, link-relative) and pointed the doc at them so the build succeeds.

- **Copy to shopify-dev**  
  After a successful build, the script now copies `docs/surfaces/admin/generated/` (including `targets.json`) into shopify-dev at `templated_apis/admin_extensions/<API_VERSION>/` when shopify-dev is present at `../../../shopify-dev`. Strips `https://shopify.dev` from `generated_docs_data.json` for Spin and rewrites `/unstable` to the version for stable builds.

- **Script behavior**  
  Build script already takes `API_VERSION` as the first argument (e.g. `yarn docs:admin 2023-04`) and defines `fail_and_exit`; no change to that. Copy step runs only when the docs build and targets generation succeed.

### DIFF
--- a/packages/ui-extensions/docs/surfaces/admin/build-docs.sh
+++ b/packages/ui-extensions/docs/surfaces/admin/build-docs.sh
@@ -42,3 +42,26 @@ targets_exit=$?
 if [ $targets_exit -ne 0 ]; then
   fail_and_exit $targets_exit
 fi
+
+# Make sure https://shopify.dev URLs are relative so they work in Spin (same as checkout)
+if [ -f ./$DOCS_PATH/generated/generated_docs_data.json ]; then
+  sed -i.bak 's|https://shopify.dev||gi' ./$DOCS_PATH/generated/generated_docs_data.json 2>/dev/null || true
+  rm -f ./$DOCS_PATH/generated/generated_docs_data.json.bak
+fi
+
+# Copy the generated docs (including targets.json) to shopify-dev when repo is present as sibling
+if [ -d ../../../shopify-dev ]; then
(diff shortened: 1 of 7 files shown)


packages/ui-extensions/docs/surfaces/admin/build-docs.sh
```diff
@@ -42,3 +42,26 @@ targets_exit=$?
 if [ $targets_exit -ne 0 ]; then
   fail_and_exit $targets_exit
 fi
+
+# Make sure https://shopify.dev URLs are relative so they work in Spin (same as checkout)
+if [ -f ./$DOCS_PATH/generated/generated_docs_data.json ]; then
+  sed -i.bak 's|https://shopify.dev||gi' ./$DOCS_PATH/generated/generated_docs_data.json 2>/dev/null || true
+  rm -f ./$DOCS_PATH/generated/generated_docs_data.json.bak
+fi
```
johndcollett, line 50: I don't think we want this

johndcollett: or maybe just modify the comment to not include spin

johndcollett (approved):
but it works

packages/ui-extensions/docs/surfaces/admin/build-docs.sh
laurelthorburn: damn saw this AFTER the fact

PR #3899, opened 2026-02-11
Add a testing library for ui-extensions

kumar303:
This adds a new `ui-extensions-tester` library, built during [Hack Days](https://vault.shopify.io/hackdays/153/projects/22955-App-Extensions-Test-tool-for-UI-extensions) (internal Shopify link).

## All stacks

* 🥞 `ui-extensions-tester` package setup <-- you are here
* 🥞 [A complete mock checkout API](https://app.graphite.com/github/pr/Shopify/ui-extensions/3902/Create-mocks-for-the-entire-checkout-extension-API)
* 🥞 [A complete mock point-of-sale API](https://app.graphite.com/github/pr/Shopify/ui-extensions/3904/Create-mocks-for-the-entire-point-of-sale-extension-API)
* 🥞 [A complete mock customer-account API](https://app.graphite.com/github/pr/Shopify/ui-extensions/3905/Create-mocks-for-the-entire-Customer-Account-extension-API)
* 🥞 [A complete mock Admin API](https://app.graphite.com/github/pr/Shopify/ui-extensions/3914/Create-mocks-for-the-entire-Admin-extension-API)

## 👁️ How to review this PR

* 📖 Read the [developer-facing documentation](https://github.com/Shopify/ui-extensions/blob/02-11-create_a_testing_library_for_ui-extensions_proof_of_concept_/packages/ui-extensions-tester/README.md)
* See one of the stacks above to view surface-specific mocks

## Example

Here's what it looks like so far in something like `your-app/extensions/your-ui-extension/tests/Extension.test.ts`:

```tsx
import {expect, test, beforeEach, afterEach} from 'vitest';
import {setUpExtension} from '@shopify/ui-extensions-tester';

// This reads from ../shopify.extension.toml
const extension = getExtension('purchase.checkout.block.render');

beforeEach(() => {
  extension.setUp();
});

afterEach(() => {
  extension.tearDown();
});

test('renders an Enter button', async () => {
  await extension.render();
  const button = extension.querySelector('s-button');
  expect(button!.textContent).toEqual('Enter');
});
```

More complete example: [`checkout-basic-testing-example/tests/Checkout.test.ts`](https://github.com/Shopify/ui-extensions/blob/02-11-create_a_testing_library_for_ui-extensions_proof_of_concept_/examples/testing/checkout-basic-testing-example/extensions/checkout-basic-testing-example/tests/Checkout.test.ts)

<hr />

Run it like this after checking out the branch for this PR:

```
yarn build
pushd examples/testing/checkout-basic-testing-example
npm install
npm test
```

# 🎩 

To try it in a standalone app, you have to install a pre-built package: [shopify-ui-extensions-tester-v2026.4.0-rc.1.tgz](https://github.com/user-attachments/files/25871729/shopify-ui-extensions-tester-v2026.4.0-rc.1.tgz). This is because `/snapit` doesn't work before there's at least one public release.

Add it to a `package.json` file like:

```
"@shopify/ui-extensions-tester": "file:../shopify-ui-extensions-tester-v2026.4.0-rc.1.tgz"
```

### DIFF
--- a/packages/ui-extensions-tester/README.md
+++ b/packages/ui-extensions-tester/README.md
@@ -1,0 +1,411 @@
+# 🧪 @shopify/ui-extensions-tester
+
+Write unit tests for [Shopify UI extensions](../ui-extensions) to ensure correctness and prevent regressions.
+
+This testing library provides strongly typed mocks of the extension API--like the `shopify` global--so you can verify the correctness of your extension without needing a real Shopify host.
+
+## 📋 Requirements
+
+- **API version `2025-10` or later** in your `shopify.extension.toml`
+- **Node.js v20.20.0** or later
+- **a mock DOM** such as [`environment: 'jsdom'`](https://vitest.dev/config/environment.html) in [`vitest`](https://vitest.dev/)
+- **Test isolation** — extensions rely on the `shopify` global, so each test file must run in its own environment. We recommend [`vitest`](https://vitest.dev/) in [isolate mode](https://vitest.dev/config/isolate.html#isolate) (enabled by default).
+
+## 📋 Recommendations
+
+- **TypeScript** — we recommend TypeScript to enforce API compliance against mock objects
+- **@testing-library/preact** — if your extension uses [Preact](https://preactjs.com/), we recommend installing [`@testing-library/preact`](https://preactjs.com/guide/v10/preact-testing-library/) for its `fireEvent` and `waitFor` helpers
+
+## 📦 Installation
+
+Install the tester as a dev dependency alongside your preferred test runner:
+
+```bash
+npm install --save-dev @shopify/ui-extensions-tester vitest
+```
+
+If your extension renders with Preact, also install `@testing-library/preact`:
+
+```bash
+npm install --save-dev @testing-library/preact
+```
+
+## 🏗️ Adding to an existing extension
+
+If your extension was seeded from an older template using <a href="https://shopify.dev/docs/api/shopify-cli/app/app-generate-extension"><code>shopify app generate extension</code></a>, follow these steps.
+
+<details>
+<summary>Expand for details</summary>
+
+The template gives you a project structure like this:
+
+```
+my-app/
+├── extensions/
+│   └── my-extension/
+│       ├── src/
+│       │   └── Checkout.jsx
+│       ├── package.json
+│       ├── shopify.d.ts
+│       ├── shopify.extension.toml
+│       └── tsconfig.json
+├── package.json
+└── shopify.app.toml
+```
+
+You need to add a few things:
+
+### 1. Add dependencies to the root `package.json`
+
+Your extension's own `package.json` (inside `extensions/my-extension/`) already lists `@shopify/ui-extensions` for Shopify CLI. However, tests and typechecking run from the **root** project directory, so you also need `@shopify/ui-extensions` in the root `package.json` — otherwise TypeScript and vitest won't be able to resolve it.
+
+```jsonc
+{
+  "scripts": {
+    // ...existing scripts
+    "test": "vitest run",
+    "typecheck": "tsc --noEmit --project extensions/my-extension/tsconfig.json"
+  },
+  "dependencies": {
+    "@shopify/ui-extensions": "latest"
+  },
+  "devDependencies": {
+    "@shopify/ui-extensions-tester": "latest",
+    "@testing-library/preact": "^3.2.0",
+    "typescript": "^5.0.0",
+    "vitest": "^3.0.0"
+  }
+}
+```
+
+### 2. Create `vitest.config.ts` at the project root
+
+```ts
+import {defineConfig} from 'vitest/config';
+
+export default defineConfig({
+  esbuild: {
+    jsx: 'automatic',
+    jsxImportSource: 'preact',
+  },
+  test: {
+    environment: 'jsdom',
+  },
+});
+```
+
+### 3. Update the extension's `tsconfig.json`
+
+Add the `tests` directory to the `include` array so your test files are typechecked:
+
+```jsonc
+{
+  "compilerOptions": {
+    "jsx": "react-jsx",
+    "jsxImportSource": "preact",
+    "target": "ES2020",
+    "strict": true,
+    "checkJs": true,
+    "allowJs": true,
+    "moduleResolution": "node",
+    "esModuleInterop": true,
+    "skipLibCheck": true
+  },
+  "include": [
+    "./src",
+    "./tests",
+    "./shopify.d.ts"
+  ]
+}
+```
+
+### 4. Create a `tests/` directory inside your extension
+
+```
+extensions/
+└── my-extension/
+    ├── src/
+    │   └── Checkout.jsx
+    └── tests/
+        └── Checkout.test.ts   ← your tests go here
+```
+
+### 5. Add a triple-slash reference in each test file
+
+Add a [triple-slash directive](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) for your target so TypeScript knows about Shopify element types (`s-button`, `s-text`, etc.):
+
+```ts
+/// <reference types="@shopify/ui-extensions/purchase.checkout.block.render" />
+```
+
+The path must match the target you pass to `getExtension()`.
+
+</details>
+
+## 🏊‍♀️ Getting started
+
+Every test file follows the same pattern: create an extension harness, call `extension.setUp()` before each test, call `extension.tearDown()` after.
+
+```ts
+import {getExtension} from '@shopify/ui-extensions-tester';
+import {beforeEach, afterEach} from 'vitest';
+
+const extension = getExtension(
+  'purchase.checkout.block.render',
+);
+
+beforeEach(() => {
+  extension.setUp();
+});
+
+afterEach(() => {
+  extension.tearDown();
+});
+```
+
+`setUp()` creates a complete mock `shopify` global compliant with the target. `tearDown()` clears the DOM and removes the global, among other surface-specific things.
+
+### 🔍 Rendering and querying elements
+
+Call `extension.render()` to import and execute your extension's callback, then query the DOM with standard DOM APIs like [`document.body.querySelector()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector) and [`document.body.querySelectorAll()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll):
+
+```ts
+test('it handles an empty cart', async () => {
+  await extension.render();
+
+  const text =
+    document.body.querySelector('s-text')!;
+  expect(text.textContent).toEqual(
+    'No items in cart',
+  );
+});
+```
+
+### 🎨 Mocking shopify API values
+
+The test setup will create a `shopify` global with sensible defaults for the target. You can mutate global property values on `extension.shopify` before rendering.
+
+```ts
+test('it handles an empty order', async () => {
+  extension.shopify.order.value = undefined;
+
+  await extension.render();
+
+  const text =
+    document.body.querySelector('s-text')!;
+  expect(text.textContent).toEqual(
+    'Order not found',
+  );
+});
+```
+
+### 🖱️ Triggering events
+
+To simulate how a user would interact with your UI extension, you can call [`dispatchEvent()`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent) or use `fireEvent` from `@testing-library/preact`. When an event triggers an async state change (like a Preact re-render), wrap follow-up assertions in `await waitFor()` to wait for the DOM to settle:
+
+```ts
+import {
+  fireEvent,
+  waitFor,
+} from '@testing-library/preact';
+
+test('it handles date field changes', async () => {
+  await extension.render();
+
+  const dateField = document.body.querySelector(
+    's-date-field',
+  )!;
+  dateField.value = '1990-05-20';
+  fireEvent.change(dateField);
+
+  const button =
+    document.body.querySelector('s-button')!;
+  fireEvent.click(button);
+
+  await waitFor(() => {
+    const banner =
+      document.body.querySelector('s-banner')!;
+    expect(banner.textContent).toEqual('Saved');
+  });
+});
+```
+
+### 🔒 Safely mocking mutation functions
+
+When mocking without strict typing, like with [`vitest` mocks](https://vitest.dev/api/vi.html#mocking-functions-and-objects), you can use a surface-specific `createResult()` helper to return type-safe values:
+
+```ts
+import {createResult} from '@shopify/ui-extensions-tester/checkout';
+
+const applyMetafieldChange = vi
+  .fn()
+  .mockResolvedValue(
+    createResult('applyMetafieldChange', {
+      type: 'error',
+      message:
+        'Could not apply metafield changes',
+    }),
+  );
+extension.shopify.applyMetafieldChange =
+  applyMetafieldChange;
+```
+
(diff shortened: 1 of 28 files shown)


kumar303:
* **#3914** <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/3914?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3905** <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/3905?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3904** <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/3904?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3902** <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/3902?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3899** <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/3899?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/3899?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `2026-04-rc`


<h2></h2>

packages/ui-extensions-tester/src/checkout/index.ts
kumar303, line 1: You can ignore this file. The full checkout API gets added in another stack.

packages/ui-extensions-tester/README.md
```diff
@@ -0,0 +1,363 @@
+# 🧪 @shopify/ui-extensions-tester
```
cpeddecord, line 1: nit: naming. Random thoughts without much merit:

- `ui-extensions-testing-library` - similar to React Testing Library
- `ui-extensions-test`\- shorter for no good reason

packages/ui-extensions-tester/src/index.ts
```diff
@@ -0,0 +1,437 @@
...
+  options?: {configSearchDir?: string},
+): Extension<T> {
+  const configSearchDir =
+    options?.configSearchDir ?? path.dirname(getCallerFile());
+  const tomlPath = findToml(configSearchDir);
+  const tomlDir = path.dirname(tomlPath);
+  const tomlContent = fs.readFileSync(tomlPath, 'utf-8');
+  validateApiVersion(tomlContent);
+  const modulePath = parseTargetModule(tomlContent, target);
+  const resolvedModule = path.resolve(tomlDir, modulePath);
```
cpeddecord, line 123: fail fast here or wait until calling `.render()`?

packages/ui-extensions-tester/src/navigation.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export type {
+  Navigation,
+  NavigationHistoryEntry,
+  NavigationNavigateOptions,
+} from '@shopify/ui-extensions/customer-account';
+
+// ---------------------------------------------------------------------------
+// createNavigationResult
+// ---------------------------------------------------------------------------
```
cpeddecord: :fire:

packages/ui-extensions-tester/src/navigation.ts
```diff
@@ -0,0 +1,80 @@
...
+      return state === undefined
+        ? undefined
+        : JSON.parse(JSON.stringify(state));
+    },
+  };
+}
+
+// ---------------------------------------------------------------------------
+// createMockNavigation
+// ---------------------------------------------------------------------------
```
cpeddecord: :fire: unless you think we need these

kumar303: Thanks. It's such a dumb robot :joy:

packages/ui-extensions-tester/README.md
```diff
@@ -0,0 +1,364 @@
...
+When mocking without strict typing, like with [`vitest` mocks](https://vitest.dev/api/vi.html#mocking-functions-and-objects), you can use surface-specific helpers to return type-safe values:
+
+```ts
+import {createApplyMetafieldChangeResult} from '@shopify/ui-extensions-tester/checkout';
+
+const applyMetafieldChange = vi
+  .fn()
+  .mockResolvedValue(
+    // This helper ensures the return type is correct.
+    createApplyMetafieldChangeResult({
```
vividviolet: Is it possible to make this more generic instead of having to pull out explicit helpers?

```
mockApiResult('applyMetafieldChange', {
      type: 'error',
      message:
        'Could not apply metafield changes',
    })
```

packages/ui-extensions-tester/README.md
```diff
@@ -0,0 +1,364 @@
...
+
+Exports from `@shopify/ui-extensions-tester`:
+
+### `getExtension(target, options?)`
+
+Returns an extension test harness for the given target. It reads `shopify.extension.toml`, finds the module for the given target, and provides helpers to mock the environment and render the extension. It locates `shopify.extension.toml` by walking up from the calling test file's directory, and falls back to searching `extensions/` under the current working directory.
+
+| Option            | Type     | Default                       | Description                                                |
+| ----------------- | -------- | ----------------------------- | ---------------------------------------------------------- |
+| `configSearchDir` | `string` | calling test file's directory | Directory to start searching for `shopify.extension.toml`. |
```
vividviolet, line 348: Should this just be a file path pointing to the toml? We can't accept multiple tomls so I'm not sure if searching makes sense?

packages/ui-extensions-tester/README.md
```diff
@@ -0,0 +1,364 @@
...
+
+A mock `fetch()` function installed as `globalThis.fetch` during `setUp()` and removed during `tearDown()`.
+
+Override it with a mock to control responses:
+
+```ts
+import {createFetchResponse} from '@shopify/ui-extensions-tester';
+
+extension.fetch = vi.fn().mockResolvedValue(
+  createFetchResponse({
```
vividviolet: Do we need this helper? Can't they do `new Response()`?

packages/ui-extensions-tester/README.md
```diff
@@ -0,0 +1,364 @@
...
+### `createFetchResponse(options?)`
+
+Creates a `Response` for mocking `fetch()` return values. Accepts standard `ResponseInit` fields plus an optional `body` that is serialised as JSON.
+
+- `status` — HTTP status code (default `200`)
+- `statusText` — HTTP status text
+- `headers` — response headers as `HeadersInit` (default `{'content-type': 'application/json'}`)
+- `body` — response body, serialised as JSON (default `{}`)
+
+#### `extension.querySelector(selector)`
```
vividviolet: Why do we need this? Couldn't they use `document.body.querySelector` directly?

vividviolet (commented):
This is amazing!  I have a few questions/suggestions

packages/ui-extensions-tester/README.md
kumar303: The only reason is for this [type overload](https://github.com/Shopify/ui-extensions/blob/4b27af205c606554747ca1b2aa810b5eac6820c9/packages/ui-extensions-tester/src/index.ts#L103-L105) which should be in the official DOM types anyway. We _could_ add a `.d.ts` file to override `document.body` types (I think?) but that seems a little too heavy handed and magical. What do you think? Should we do it in `.d.ts`?

vividviolet: Doesn't the generated `shopify.d.ts` already work with our custom elements? If the tests can be configured to use it then there's no need to override anything

packages/ui-extensions-tester/README.md
kumar303: Good call. I removed it and changed everything to use `new Response()`.

packages/ui-extensions-tester/README.md
kumar303: In practice, I don't think anyone would need to call `configSearchDir`. The default behavior of searching in walking up from the `tests` directory should work fine for everyone calling `getExtension()`.   
  
If I were to change this option to `configPath` then it would make sense from an API perspective but we'd lose the ability to easily write unit tests covering the config searching logic. I tried a few other approaches for unit testing but it became too difficult (lots of flaky tests).   
  
It's cleaner for internal testing purposes to expose `configSearchDir` and since this isn't an option that many developers would call anyway, I'm not worried about how it's not a useful option. Someone _might_ find it useful as an escape hatch for an edge case I haven't thought of yet.

packages/ui-extensions-tester/README.md
kumar303: Oh hey, I took a closer look at this and it was just because of a bug in the POS types: https://github.com/shop/issues-retail/issues/25351  
  
I removed the `extension.querySelector()` thing and rewrote all the tests to use `document.body.querySelector()`. I added assertions for the POS test suite just for now until we fix the bug. All the types compile just fine :sparkles:

packages/ui-extensions-tester/README.md
kumar303: I haven't thought too deeply about this but importing the existing helpers is pretty straight forward. Are you concerned with how many helpers there are?

vividviolet: Yeah, it's just about simplifying things. Users will only need to learn to use a single mocker `mockApiResult`. I think the implementation for these would be identical for the most part - it's just the typing the differs.

kumar303: Yessss, this is much cleaner. Thank you. Every surface is updated and you can see an example in the [main docs](https://github.com/Shopify/ui-extensions/blob/02-11-create_a_testing_library_for_ui-extensions_proof_of_concept_/packages/ui-extensions-tester/README.md). I went with `createResult()` to match `create*` in all other helpers and to keep the name short but I can change it.

packages/ui-extensions-tester/src/index.ts
```diff
@@ -0,0 +1,432 @@
...
+  return ext as Extension<T>;
+}
+
+function validateApiVersion(toml: string): void {
+  const match = toml.match(/^\s*api_version\s*=\s*"([^"]+)"/m);
+  const tomlVersion = match?.[1];
+  if (tomlVersion !== API_VERSION) {
+    throw new Error(
+      `api_version "${tomlVersion ?? '(not found)'}" does not match ` +
+        `the version supported by @shopify/ui-extensions-tester ("${API_VERSION}"). ` +
```
vividviolet, line 231: This means we'd have to do do a new version of the test library for every library release even though most of the code remains the same. I do see the value in making it simple for developers though. We might want to automate this somehow

packages/ui-extensions-tester/src/mocks/signals.ts
```diff
@@ -0,0 +1,16 @@
+import type {SubscribableSignalLike} from '@shopify/ui-extensions/checkout';
+
+/**
+ * Creates a mock `SubscribableSignalLike` that wraps a static value.
+ * The `subscribe` callback is never invoked since test values are static.
```
vividviolet, line 5: How do we write tests to work with changing signals then?

packages/ui-extensions-tester/src/tests/helpers.ts
```diff
@@ -0,0 +1,140 @@
...
+import * as path from 'path';
+import * as os from 'os';
+
+import {API_VERSION} from '../api-version';
+
+export const TOML_NAME = 'shopify.extension.toml';
+
+/**
+ * Creates an isolated temp directory for a single test file.
+ * Each test file gets its own directory so tests can run in parallel
```
vividviolet, line 11: Can you clarify why we need to do this? Tests shouldn't mutate the toml so why do we need to copy these tests to their own directory and clone the toml?

vividviolet (approved):
Looks great! Thanks for addressing all the feedback. This is good to merge - I just have a few questions that could be addressed if needed in follow up prs

packages/ui-extensions-tester/src/index.ts
kumar303: Yes, we'd need to release `ui-extensions-tester` alongside `ui-extensions` but if no changes are needed then the version bump would happen automatically. It gets set in a [build step](https://github.com/Shopify/ui-extensions/blob/82c4bcef533a6c6fb9d84f848f829aeb06cb5005/packages/ui-extensions-tester/loom.config.ts#L21).

packages/ui-extensions-tester/src/mocks/signals.ts
kumar303: Since preact is the default, I think most extensions could be tested in a functional style. In other words, the mock `shopify` object won't have working signals but I don't think that matters.

To illustrate:

Let's say your extension has a button to change cart quantities and it also subscribes to line items. You wouldn't need the subscription to be fully functional. You could just test it in two steps:

```typescript
// test that the button works
const mock = vi.fn().mockResolvedValue(
  createResult('applyCartLineChange');
);
extension.shopify.applyCartLineChange = mock;

await extension.render();
const button = document.body.querySelector('s-button');
fireEvent('click', button);
// Make sure the button changes the quantity
await waitFor(() => {
  expect(mock).toHaveBeenCalledWith(...);
});
```

...

```typescript
// test that the extension renders the lines
const line = createLineItem();
// This essentially simulates a subscription update:
extension.shopify.lines = [line];

await extension.render();
const text = document.body.querySelector('s-text');
// Make sure the line's quantity was rendered:
expect(text.textContent).toContain(line.quantity);
```

This testing style _assumes_ the shopify API fully works as specified -- I think that assumption is fine.

packages/ui-extensions-tester/src/tests/helpers.ts
kumar303: This is just for the internal `ui-extensions-tester` test suite. Developers would never use this. 

It's necessary because some of the tests need to work with different TOML files to simulate different configurations. The fixtures getting copied here are just for defining an extension that can be imported and executed. The fixtures don't include a TOML file.

The first approach I tried was to write a local TOML file before each test but I started seeing flakiness. It worked better when creating a unique TOML file per test.

packages/ui-extensions-tester/src/index.ts
kumar303: I think waiting to call `.render()` is fine. It will fail in the test which is better than crashing the whole module.

packages/ui-extensions-tester/src/mocks/signals.ts
kumar303: Thanks for calling this out, Trish. I [documented the suggested pattern](https://github.com/Shopify/ui-extensions/blob/02-11-create_a_testing_library_for_ui-extensions_proof_of_concept_/packages/ui-extensions-tester/README.md#-testing-extension-code-that-relies-on-signals) to use for this.

jamesvidler (approved):
I ran into a snag with the error `ReferenceError: __filename is not defined` when running the tests, so I had to add the following to my vitetest.config  


```
export default defineConfig({
  esbuild: {
    jsx: "automatic",
    jsxImportSource: "preact",
  },
  test: {
    environment: "jsdom",
//NEEDED TO ADD THIS, or a I got an error ReferenceError: __filename is not defined when running the tests
    server: {
      deps: {
        inline: ["@shopify/ui-extensions-tester"],
      },
    },
  },
});
```

  
It looks like we've patched that now, but I wasn't able to get the updated tgz package to relfect the changes.  
  
I'm comfortable approving and we can test again once the package is publicly released and fix forward if neccessary or update the docs.

kumar303:
### Merge activity

* **Mar 12, 2:49 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/ui-extensions/3899).
* **Mar 12, 2:49 PM UTC**: @kumar303 merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/ui-extensions/3899).

PR #3919, opened 2026-02-18
[2026-04-rc]: Transition POS component names from Pascal case to sentence case

mcvinci:
## Context

In older versions (2025-07 and earlier), components use React/JSX syntax, so PascalCase names like `ButtonGroup` make sense — that's what devs write in their code. But in versions 2025-10+, the syntax shifted to web components (`<s-button-group>`), so PascalCase no longer maps to anything developers actually write in code. We're switching to sentence case for component names in versions 2025-10+.

## This PR:
- Updates all POS component names to use sentence case versus Pascal case in the 2026-04-rc version
- Uses the `{API_VERSION}` variable in links for ease of maintenance going forward

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/pin-pad-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/pin-pad-api.ts
@@ -4,7 +4,7 @@ export interface PinPadApiContent {
   /**
    * Shows a PIN pad to the user in a modal dialog. The `onSubmit` function is called when the PIN is submitted and should validate the PIN, returning `'accept'` or `'reject'`.
    *
-   * • **When accepted**: Modal dismisses and triggers the `onDismissed` callback—perform any post-validation navigation in this callback rather than in `onSubmit`.
+   * • **When accepted**: The modal dismisses and triggers the `onDismissed` callback—perform any post-validation navigation in this callback rather than in `onSubmit`.
    *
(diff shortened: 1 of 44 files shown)


sordaz00 (approved):
Looks good

jhoffmcd:
Hi @mcvinci  I think this PR may have introduced a doc generation bug for multi-word component names like `ChoiceList`. The name field was changed from `ChoiceList` to `Choice list`, but the doc generator uses name to construct the props type lookup `({name}ElementProps)`, which now produces `Choice listElementProps` instead of `ChoiceListElementProps`.

Running `yarn docs:checkout 2026-01` fails with:

```
Error generating docs: Error: Type Choice listElementProps was not found. Did you mean: ChoiceListElementProps?
```

A quick fix might be setting type: 'ChoiceList' on the top-level type field (currently type: '') so the generator doesn't fall back to name for the lookup. Let me know if you'd like me to open a fix!

mcvinci:
Hey @jhoffmcd! Thanks for connecting about this, and apologies for introducing the bug. 😢 If you're able to put together a PR with that fix, I'm definitely happy to take a look!

jhoffmcd:
@mcvinci  Thank you! I'll let you know when I have something ready!

jhoffmcd:
@mcvinci Here is the PR: https://github.com/Shopify/ui-extensions/pull/4048. Thanks again!

PR #3950, opened 2026-02-23
2024-07 gen docs v1

johndcollett:
### Background

Closes https://github.com/Shopify/temp-project-mover-Archetypically-20260312105649/issues/388

**TL;DR:** We are ready to begin the migration of POS docs to the latest major gen docs release.

### Solution

- **Doc generation:** Added `/** @publicDocs */` immediately before every top level interface for POS ONLY.

- **Why this approach:** The existing doc pipeline uses `@publicDocs` to decide what to include; adding the tag is the minimal, consistent fix and matches how other public API surfaces are documented.

`yarn docs:point-of-sale 2024-07` to generate the generated docs data v2 json file in the shopify-dev repo

### Checklist

## Documentation reference tracker:
- [x] POS
- [x] Admin
- [x] Checkout
- [ ] Customer


laurelthorburn:
Only migrating version 2025-07 and newer

PR #3953, opened 2026-02-23
Fix invalid enum for discount classes in 2026-01

jonathanhamel4:
### Background

When updating our internal app to 2026-01, I realized that our ui-extension type contract was wrong and that the type was invalid. The data returned from admin returns a lower cased value, coming from a string union type. 

I will also make a PR to 2026-04-rc.

### DIFF
--- a/.changeset/fast-llamas-press.md
+++ b/.changeset/fast-llamas-press.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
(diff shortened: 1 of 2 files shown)


devisscher (approved):

jonathanhamel4:
cc @vividviolet just making sure I'm ok to add a patch to 2026-01.

vividviolet:
Yeah, go for it

PR #3958, opened 2026-02-24
revert pascal type name in component definitions helper function

lrsterrett:
Reverting naming convention changes in `packages/ui-extensions/src/docs/shared/component-definitions.ts` based on this [thread](https://shopify.slack.com/archives/C0683EBQ55M/p1771949645067069?thread_ts=1771941828.772789&cid=C0683EBQ55M). Changes were part of this [PR](https://github.com/Shopify/ui-extensions/pull/3957)

In older versions (2025-07 and earlier), components use React/JSX syntax, so PascalCase names like ButtonGroup make sense — that's what devs write in their code. But in versions 2025-10+, the syntax shifted to web components (<s-button-group>), so PascalCase no longer maps to anything developers actually write in code. We've switching to sentence case for component names in versions 2025-10+.

POS already has these updates in production, and when we release Admin and App Home over the next week or so, they'll also have the casing updates.

### DIFF
--- a/packages/ui-extensions/src/docs/shared/component-definitions.ts
+++ b/packages/ui-extensions/src/docs/shared/component-definitions.ts
@@ -32,13 +32,6 @@ interface ComponentDoc {
   extraExamples?: ReferenceEntityTemplateSchema['examples'];
 }
 
-function toPascalCase(name: string): string {
-  return name
-    .split(' ')
-    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
-    .join('');
-}
-
 function buildDefinitions({
   name,
   description,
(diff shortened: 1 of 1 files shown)


majd-shopify (approved):

andrewmcgov (approved):
Should we be regenerating the docs JSON files along with this?

mcvinci (approved):
Thanks so much, @lrsterrett!

majd-shopify:
> Should we be regenerating the docs JSON files along with this?

Good call, I think it affects them!

lrsterrett:
> Should we be regenerating the docs JSON files along with this?

I don't think so as these changes were needed for them to build anyways. Don't think the changes affect anything that shows up in the docs, just the references/names need to actually get built

PR #3995, opened 2026-02-26
Code examples for App Home Polaris typography and content section

timtrevor-shopify:
### Background

As part of an overall update to the App Home / App Bridge docs, we're improving the titles and descriptions of code examples for Polaris components, as well as removing any redundant examples, and adding new examples for missing use cases.

### Solution

This PR makes updates to the following component pages:
[Chip](https://shopify.dev/docs/api/app-home/polaris-web-components/typography-and-content/chip)
[Heading](https://shopify.dev/docs/api/app-home/polaris-web-components/typography-and-content/heading)
[Paragraph](https://shopify.dev/docs/api/app-home/polaris-web-components/typography-and-content/paragraph)
[Text](https://shopify.dev/docs/api/app-home/polaris-web-components/typography-and-content/text)
[Tooltip](https://shopify.dev/docs/api/app-home/polaris-web-components/typography-and-content/tooltip)

The following code examples have been removed and added:

**Chip**
Removed example files:
`basic-usage.html` — Very similar to the default example; both just show a single chip with a label. Updated the default to add color and accessibility labels.
`product-status.html` — Shows three chips with different colors and icons, which is essentially the same concept as `color-variants.html`. Both demonstrate using multiple chips with different color props.

**Heading**
Removed example files:
`basic-heading.html` — Contains `<s-heading>Product details</s-heading>`, which is functionally identical to the default (`<s-heading>Online store dashboard</s-heading>`). No unique props demonstrated.

**Paragraph**
Removed example files:
`basic-usage.html` — A plain paragraph with default styling, duplicating the default example.
`admin-ui-patterns.html` — Shows five paragraphs with different tones, which overlaps with `with-tone-and-color.html`. (Consolidated the best of both into with-tone-and-color.html, which now shows all five tones (info, success, caution, warning, critical) instead of just two).

Updated example files:
`with-tone-and-color.html` — Expanded from 2 tones (info, success) to all 5 available tones (info, success, caution, warning, critical), absorbing the content from the deleted admin-ui-patterns example.

Added example files:
`subdued-color.html` — Demonstrates `color="subdued"` for secondary/helper text. The tones example already covered tone variations but always used `color="base"`. The subdued color option for de-emphasized text (helper text, disclaimers, secondary descriptions) is a common pattern that now has its own example.

**Text**
Removed example files:
`basic-usage.html` — Plain text with no props, redundant with the default example which already shows basic text usage.
`strong-text.html` — Shows `type="strong"`, but the default example already demonstrates this with `<s-text type="strong">Name: </s-text>`.

**Tooltip**
Updated example files:
`with-icon-button.html` — Fixed an icon/tooltip mismatch which could have caused confusion: the original used an info icon but had tooltip text saying "Delete item permanently". Changed to a button with tooltip text "View product details and inventory status".
`basic-usage.html` — Cleaned up by removing the unnecessary <s-text> wrapper inside the tooltip content (plain text is sufficient).

Add example files:

`keyboard-shortcut.html` — Shows a tooltip displaying a keyboard shortcut (⌘S) on a Save button. The best practices explicitly call out "Include the button's action and keyboard shortcut if available" as a key tooltip use case, but no existing example demonstrated this pattern.

### Checklist


packages/ui-extensions/src/surfaces/admin/components/Chip/Chip.doc.ts
SteveSill: Not part of your PR, but is this a circular link?

We should also get rid of the em-dash here.

The best practices say "Don't use chips for actions....they're visual indicators, not buttons" which is correct. But it doesn't direct the reader to `ClickableChip` as the alternative for interactive use cases. The Polaris spec says "For interactive chips, use ClickableChip."

Suggested addition:
> For interactive or dismissible chips, use [clickable chip](/docs/api/{API_NAME}/{API_VERSION}/polaris-web-components/typography-and-content/clickable-chip) instead.

SteveSill (approved):
Just one comment, not related to your PR. But might be good to fix up.

Custom prop validation script checked every `<s-*>` tag in all HTML files against the Polaris spec prop lists: **all props valid, all values valid.**

packages/ui-extensions/src/surfaces/admin/components/Chip/Chip.doc.ts
timtrevor-shopify: Thanks @SteveSill , incorporated these changes. I think these same best practices descriptions are also in previous API versions too, so you should check on your Admin UI pages.

PR #4001, opened 2026-02-26
feat: add automation workflow

majd-shopify:
## Problem

Part of https://github.com/Shopify/temp-project-mover-Archetypically-20260312105649/issues/7

## Solution

Check issue since this is a public repo


avocadomayo (approved):
The doc generation flow feels more intuitive with this change, ie. you make a type change and the corresponding doc change flows down to `shopify-dev` ❤️

PR #4064, opened 2026-03-09

sordaz00:
## Summary
- Creates placeholder example directories for two Customer Account targets that were missing them
- Enables example content to be added in follow-up PRs

**Targets:**
- `[customer-account.order.page.render](https://shopify.dev/docs/api/customer-account-ui-extensions/latest/targets/full-page/customer-account-order-page-render)` (order-specific full page extension)
- `[customer-account.page.render](https://shopify.dev/docs/api/customer-account-ui-extensions/latest/targets/full-page/customer-account-page-render)` (general full page extension)

Closes: https://github.com/Shopify/shopify-dev/issues/68753

## Test plan
- [ ] Verify directories are created in correct location
- [ ] Confirm placeholder examples follow existing patterns


mcvinci (approved):
Thanks, @sordaz00! Do these example directories also need to be added for version 2026-04-rc, 2025-10, and 2025-07?

PR #4072, opened 2026-03-09
[2026-04] Remove metafield examples from checkout docs

avocadomayo:
### Background

A collection of doc changes to `2026-04` Checkout UI Extension docs:
- Remove remaining uses of checkout metafield API from examples
- Update TOML examples to use `2026-04`
- Update generated doc data


packages/ui-extensions/docs/surfaces/checkout/staticPages/examples/configuration/default.example.toml
avocadomayo, line 1: Files in `packages/ui-extensions/docs/surfaces/checkout/staticPages/*` are no longer used for page content generation but keeping these consistent until we delete them

lrsterrett (approved):

packages/ui-extensions/docs/surfaces/checkout/generated/generated_static_pages.json
jamesvidler, line 1: This file isn't used anymore, right? Can we remove it entirely?

packages/ui-extensions/docs/surfaces/checkout/generated/generated_docs_data.json
jamesvidler, line 1: Looks like this generated file might have been accidentally checked into git. We typically don't store it here in the repo (its very large) and only commit it into shopify-dev. Was there a reason we wanted to start tracking it in the repo? If not, you can also delete this file.

.changeset/five-rings-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+[Checkout] Update code examples
```
jamesvidler, line 5: Just an FYI if the only changes are docs, then you don't need to add a changelog. Because a changelog will result in releasing a new npm package. Since this is removing examples though etc for an RC, I think add a changelog for this is fine.

packages/ui-extensions/docs/surfaces/checkout/staticPages/examples/configuration/default.example.toml
jamesvidler: I would also be ok with removing these entirely - which would force us to start getting used to updating this in shopify-dev repo which we need to start doing anyways.

packages/ui-extensions/docs/surfaces/checkout/reference/examples/purchase.checkout.shipping-option-list.render-after/default.example.tsx
```diff
@@ -59,12 +59,14 @@ function Extension() {
...
-        value: isChecked ? 'yes' : 'no',
-        valueType: 'string',
+      await shopify.applyMetafieldChange({
+        type: 'updateCartMetafield',
+        metafield: {
+          namespace: metafieldNamespace,
+          key: metafieldKey,
+          value: isChecked ? 'yes' : 'no',
+          type: 'single_line_text_field',
+        },
```
jamesvidler, line 69: I noticed in this code example we don't actually read any existing value for metafields (i.e preselect the checkbox if the metafield already exists). It would be nice to add that, but maybe outside the scope of these changes you are doing. Also, if we were to do that we'd need to add a TOML extension example to show that you need to add the `isGift-oneTimePurchase` and `isGift-subscription` metafield keys so they can be read. I don't think we need this if we are simply setting and forgetting like we are doing here (maybe? worth checking...)

jamesvidler (approved):
I think we can delete some of the files being modified here either in this PR or a follow-up. No blocking comments.

packages/ui-extensions/docs/surfaces/checkout/generated/generated_docs_data.json
avocadomayo: Shoot I started committing these by accident a few weeks ago. Removed, thank you 🙏

PR #4076, opened 2026-03-10

sordaz00:
## Summary
Cherry-pick of #4064 for 2025-10 branch.

Creates placeholder example directories for two Customer Account targets:
- `customer-account.order.page.render`
- `customer-account.page.render`


mcvinci (approved):
Similarly, approving this, but given targets examples are sourced / authored in MDX in `shopify-dev` / the monorepo, I'm not sure that we need these changes.

PR #4077, opened 2026-03-10

sordaz00:
## Summary
Cherry-pick of #4064 for 2026-04-rc branch.

Creates placeholder example directories for two Customer Account targets:
- `customer-account.order.page.render`
- `customer-account.page.render`


mcvinci (approved):
Hey @sordaz00! Approving this, but given that targets examples are authored / sourced from MDX in `shopify-dev` / the monorepo, I'm not entirely sure if we need these changes.

PR #4078, opened 2026-03-10
add discount index targets

jonathanhamel4:
### Background

Adds discount index bulk action targets that are already added in [Core](https://github.com/shop/world/pull/460701) and about to be shipped in Admin under a beta.

Do I need to add to other versions? Link and UI Extensions for actions aren't versioned the same way

### Solution

Allows to use `discount-index.selection-action.should-render` and `discount-index.selection-action.render` extension points.

### DIFF
--- a/.changeset/grumpy-lamps-switch.md
+++ b/.changeset/grumpy-lamps-switch.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+Adds discount-index.selection-action extension point
--- a/packages/ui-extensions/src/surfaces/admin/extension-targets.ts
+++ b/packages/ui-extensions/src/surfaces/admin/extension-targets.ts
@@ -320,6 +320,14 @@ export interface ExtensionTargets {
   >;
 
   /**
+   * An action target that appears in the **More actions** menu on the discount index page when multiple discounts are selected. Use this to create workflows for bulk discount operations or batch data updates.
+   */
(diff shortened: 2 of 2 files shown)


jonathanhamel4:
@mcvinci I want to add a new target that is already supported in core. Do I need to add the generated docs or are they pulled automatically? Pinged you since the blame of those files pointed to you generating them. When I run like `yarn docs:admin 2026-04`, there seem to be a lot of noise in those files.

lindadamus (approved):

mcvinci:
> @mcvinci I want to add a new target that is already supported in core. Do I need to add the generated docs or are they pulled automatically? Pinged you since the blame of those files pointed to you generating them. When I run like `yarn docs:admin 2026-04`, there seem to be a lot of noise in those files.

@majd-shopify Could you confirm whether the generated docs need to be committed? I think committing the generated files is primarily for tophatting / previewing purposes, but not sure if they need to be committed to get pulled automatically into Shopify.dev.

@jonathanhamel4 On a related note, we're aiming to release a big refresh to the App Home and Admin UI extensions docs by end of week, so we're going to be regenerating the docs quite frequently over the next while.

You can see what things will look like here: 
- https://shopify-dev-staging8.shopifycloud.com/docs/api/admin-extensions/latest
- https://shopify-dev-staging8.shopifycloud.com/docs/api/app-home

mcvinci (approved):

PR #4083, opened 2026-03-10
[Customer account UI extensions 2024-10]: Set up component sidebar

mcvinci:
## This PR: 
- Adds .doc.ts files and examples for ~60 shared checkout components (plus Chat and StyleHelper) to the customer account surface, so component reference pages are generated directly instead of linking out to checkout
- Migrates the doc build pipeline from build-docs.sh to build-docs.mjs with parallel doc generation and proper type resolution
- Creates build-doc-shared.mjs (utility module required by build-docs.mjs) and helper.docs.ts for the customer-account surface (fixing ~13 component docs that failed to compile due to a missing import)
- Passes the full checkout source as --typesInput and adds React types for reference docs, with surface blanking during builds to avoid duplicate type conflicts
- Removes the empty category overview page so subcategories render directly from component entries
- Aligns subcategories across shared component docs to match the new sidebar IA (e.g., "Feedback" → "Feedback and status indicators", "Interactive" → "Media and visuals")
- Relates to https://github.com/shop/world/pull/486452
- Partially closes https://github.com/Shopify/temp-project-mover-Archetypically-20260313091522/issues/7


sordaz00 (approved):
All seems good! One question: Claude notes that this new component documentation references 127 image files, but only 41 screenshots exist in the customer-account screenshots folder. We are planning to use previews - should all of the references be removed? If the references remain with the build will fail or be fine?

mcvinci:
> All seems good! One question: Claude notes that this new component documentation references 127 image files, but only 41 screenshots exist in the customer-account screenshots folder. We are planning to use previews - should all of the references be removed? If the references remain with the build will fail or be fine?

Good question! I think that we can probably clean up those references as part of working on component examples a bit later on. In addition, these `doc.ts` files will be going away shortly as @midhatdlg is migrating the customer account UI extensions docs to MDX. If the references remain for the time being, the build won't fail though.

PR #4090, opened 2026-03-11
Fix hardcoded 2025-10-rc component URLs in version upgrade guide (2025-10)

rennyG:
### Background

The static page file \`upgrading-to-2025-10.doc.ts\` exists on each version branch as a "how to upgrade to the current version" guide. On the \`2025-10\` branch, its title is "Upgrading to 2025-10". On \`2026-01\` it says "Upgrading to 2026-01", and so on — the content is updated per branch, but the filename stays the same.

Inside the content, there is a component migration table with links to polaris web component pages. These links were originally written with \`2025-10-rc\` in the URL (from when the file was authored during the RC period). The \`2025-10-rc\` version was never deployed to shopify.dev — only \`2025-10\` (stable) was — so these links 404 and fail MDX link validation in CI.

### Solution

Replace all \`/checkout-ui-extensions/2025-10-rc/\` references in the component migration table with \`/checkout-ui-extensions/2025-10/\`, pointing to the correct stable version where these polaris web component pages (e.g. Clickable chip, Scroll box, Press button) actually exist.

On the \`2026-01\` and \`2026-04-rc\` branches, the docs generator replaces the version in URLs with the current branch version — so fixing \`2025-10-rc\` → \`2025-10\` on those branches ensures the generator produces valid versioned links rather than broken ones.

### 🎩

- Verify the migration guide page renders correctly with valid links to polaris web components (e.g. Clickable chip, Scroll box, Press button)

### Checklist


mcvinci (approved):
🌴

PR #4108, opened 2026-03-13
fix(docs): Rename localization.doc.ts to fix URL slug collision (2025-01)

sordaz00:
## Summary
- Renames `order-status-api/localization.doc.ts` → `order-status-api/localization-order-status.doc.ts`
- Fixes URL slug collision where both "Localization API" and "Localization API (Order Status)" were generating the same URL `/apis/localization`

## Problem
Both APIs had source files named `localization.doc.ts`, causing the doc generation system to produce identical URL slugs.

## Solution
Rename the order-status version to `localization-order-status.doc.ts` so each API gets a unique slug:
- Localization API → `/localization`
- Localization API (Order Status) → `/localization-order-status`

## Test plan
- [ ] Verify docs generate without errors
- [ ] Confirm both Localization APIs have distinct URLs


sordaz00:
Not needed, able to fix in World alone.

PR #4114, opened 2026-03-16
ui-extensions-tester: support explicit resource management

kumar303:
Support [explicit resource management](https://github.com/tc39/proposal-explicit-resource-management) (`using` keyword) and prefer this as the library interface.

### Notes

- **New** **`setUpExtension`** **function** as the entry point
- **Caching**: allow `setUpExtension()` to be called repeatedly
- **Backward compatibility**: The existing `getExtension` function remains unchanged for runtimes that don't support Explicit Resource Management

# 🎩

📄 [Preview the README](https://github.com/Shopify/ui-extensions/blob/03-16-ui-extensions-tester_support_explicit_resource_management_using_keyword_/packages/ui-extensions-tester/README.md)

Using the snapshot packages (below), I verified I could create a test suite via `using`. This required some build fixes 👍 

### Checklist

### DIFF
--- a/packages/ui-extensions-tester/src/index.ts
+++ b/packages/ui-extensions-tester/src/index.ts
@@ -32,21 +32,19 @@ type Mutable<T> = T extends (...args: any[]) => any
   ? {-readonly [K in keyof T]: Mutable<T[K]>}
   : T;
 
-interface Extension<T extends AnyExtensionTarget> {
-  /**
-   * Sets up an extension environment for testing.
-   *
-   * For example, it creates a mock `shopify` global with some defaults.
-   */
-  setUp(): void;
+/**
+ * `Symbol.dispose` for runtimes that support it, with a polyfill
+ * fallback so the library works on older Node versions too.
+ */
+export const SymbolDispose: typeof Symbol.dispose = ((Symbol as any).dispose ??
+  Symbol.for('Symbol.dispose')) as typeof Symbol.dispose;
 
-  /**
-   * Tears down the extension environment.
-   *
-   * For example, it resets the `shopify` global and clears `document.body`.
-   */
-  tearDown(): void;
-
+/**
+ * Members shared by both {@link ExtensionHarness} (returned by
+ * `getExtension`) and {@link DisposableExtensionHarness} (returned
+ * by `setUpExtension`).
+ */
+interface BaseExtensionHarness<T extends AnyExtensionTarget> {
   /**
    * Imports and executes the extension module's default export,
    * rendering the extension into `document.body`.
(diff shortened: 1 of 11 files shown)


kumar303:
* **#4114** <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/4114?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/4114?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `2026-04-rc`


<h2></h2>

robin-drexler:
> Caching: allow setUpExtension() to be called repeatedly

can you elaborate on this a bit? Wouldn't you want a new one to be created every time you call `setUpExtension()`?
Could caching cause any surprises for users?

packages/ui-extensions-tester/src/index.ts
```diff
@@ -94,6 +92,163 @@ interface Extension<T extends AnyExtensionTarget> {
...
+ * using extension = setUpExtension('purchase.checkout.block.render');
+ * ```
+ */
+interface DisposableExtensionHarness<T extends AnyExtensionTarget>
+  extends BaseExtensionHarness<T> {
+  [SymbolDispose](): void;
+}
+
+class Extension<T extends AnyExtensionTarget> implements ExtensionHarness<T> {
+  private _target: T;
```
robin-drexler: nit/question why `_`? Instead of private, you'd also use `#`?

packages/ui-extensions-tester/src/index.ts
```diff
@@ -112,114 +267,46 @@ interface Extension<T extends AnyExtensionTarget> {
...
-      previousFetch = (globalThis as any).fetch;
-      previousNavigation = (globalThis as any).navigation;
-      (globalThis as any).shopify = deepWritableProxy(
-        createMockTargetApi(target),
-      );
-      (globalThis as any).fetch = fetchImpl;
-      (globalThis as any).navigation = navigationImpl;
-    },
+): ExtensionHarness<T> {
+  const cacheKey = JSON.stringify([target, options]);
```
robin-drexler: The robot found this:

 The cache key is computed before the default configSearchDir is resolved:

  const cacheKey = JSON.stringify([target, options]);
  // options is undefined → key is '["purchase.checkout.block.render"]'

  But inside the Extension constructor, the default is caller-dependent:

  const configSearchDir = options?.configSearchDir ?? path.dirname(getCallerFile());

  Two different test files calling getExtension('purchase.checkout.block.render') without options produce the same cache key but should resolve different configSearchDir values. The second caller silently gets the first caller's instance.
  This is the most common usage pattern (no explicit configSearchDir), so this would be hit frequently.

  Fix: Either resolve the default configSearchDir before computing the cache key, or include the resolved path in the key.

packages/ui-extensions-tester/src/index.ts
kumar303: I was thinking of callers that won't be coming from TypeScript. Would they still see all the properties if I use `#`? It doesn't matter much but I thought underscores would be clear that the properties are private.

robin-drexler: > Would they still see all the properties if I use #? 

what do you mean by this? Those fields are private, callers should not see those, right?

packages/ui-extensions-tester/src/index.ts
kumar303: Ah, nice, good catch.

packages/ui-extensions-tester/src/index.ts
kumar303: Ah, that's true. I was thinking of plain objects for some reason but since this is a class it doesn't matter. I'll switch it to `#` properties.

kumar303:
> > Caching: allow setUpExtension() to be called repeatedly
> 
> can you elaborate on this a bit? Wouldn't you want a new one to be created every time you call `setUpExtension()`? Could caching cause any surprises for users?

@robin-drexler The only thing cached is the extension object itself which looks for and parses the TOML file on construction. As for surprises, yes, you just made me realize tests that run in watch mode won't pick up changes made to the TOML file. I just fixed that. I can't think of any other surprises.

robin-drexler (approved):

kumar303:
### Merge activity

* **Mar 18, 3:31 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/ui-extensions/4114).
* **Mar 18, 3:32 PM UTC**: [Graphite](https://app.graphite.com/github/pr/Shopify/ui-extensions/4114) rebased this pull request as part of a merge.
* **Mar 18, 3:36 PM UTC**: @kumar303 merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/ui-extensions/4114).

PR #1418, opened 2023-10-04
apply global eslint rules

jamesvidler:
### Background

Implements improvements to eslint config that were already done in the private checkout-web package.

### Solution

Remove file-based eslint configs infavor of path based overrides.

### 🎩

- ...

### Checklist

- [ ] I have :tophat:'d these changes
- [ ] I have updated relevant documentation

PR #4123, opened 2026-03-17
feat(pos): add TypeScript types for pos.app.ready.data background target

vctrchu:
## What

Add TypeScript types for the `pos.app.ready.data` persistent background extension target.

### New types

- **`DataTargetApi<T>`** — API surface for non-rendering data targets. Includes non-UI APIs (cart, storage, session, locale, connectivity, device, product search); excludes UI-presenting APIs.
- **`ShopifyEventMap`** — Maps event names to their typed `Event` subclasses (`transactioncomplete`, `cashtrackingsessionstart`, `cashtrackingsessioncomplete`).
- **`DataExtensionTargets`** — New target interface with `pos.app.ready.data` as a `RunnableExtension<DataTargetApi, undefined>`.
- **`DataExtensionTarget`** — Type alias for data target keys.
- **`ShopifyGlobal`** — POS `shopify` global type, augmented with `addEventListener` / `removeEventListener` keyed on `ShopifyEventMap`. Follows the same `declare global` pattern Checkout uses.

### Event types

- **`TransactionCompleteEvent`** — Discriminated union of three module-local variant interfaces (`SaleCompleteEvent | ReturnCompleteEvent | ExchangeCompleteEvent`), each extending `BaseTransactionCompleteEvent`. All fields are `readonly`. Narrow on `event.transactionType` (`'Sale'` | `'Return'` | `'Exchange'`) to access per-type fields (`lineItems`, `refundId`, `lineItemsAdded`, etc.). Shared fields (`orderId`, `grandTotal`, `customer`, `paymentMethods`, …) are available without narrowing. The three variants are internal scaffolding — only `TransactionCompleteEvent` is part of the public surface.
- **`BaseTransactionCompleteEvent`** — Internal shared base extending `Event`. Contains the 12 fields common to Sale/Return/Exchange (orderId, customer, grandTotal, taxTotal, paymentMethods, shippingLines, etc.).
- **`CashTrackingSessionStartEvent`** — `Event` subclass with `id: number` and `openingTime: string` (ISO 8601).
- **`CashTrackingSessionCompleteEvent`** — Adds `closingTime: string`.

### Folder structure

Event types live at the surface-level `point-of-sale/events/` (sibling to `api/` and `globals.ts`), not nested inside `api/data-target-api/`. Host events are dispatched on the global `shopify` object and consumable by any target, so they aren't scoped to a specific API. The layout mirrors the existing `api.ts` + `api/` pattern:

```
point-of-sale/
├── api.ts     + api/       (APIs)
├── events.ts  + events/    (host events)
└── globals.ts
```

### `shopify` global augmentation

POS host events are received via `shopify.addEventListener()` on the global `shopify` object, consistent with TAG proposal Shopify/ui-api-design#1418.

```ts
shopify.addEventListener('transactioncomplete', (event) => {
  if (event.transactionType === 'Sale') {
    console.log(event.lineItems, event.grandTotal);
  }
});
```

## Why

POS apps need a persistent background extension that starts when POS loads and runs for the session lifetime. This replaces the existing fire-and-forget `.event.observe` targets with a single long-lived target that receives events via `shopify.addEventListener`, aligning with the shape TAG approved in Shopify/ui-api-design#1418.

## Related PRs

- **ui-api-design (TAG proposal):** Shopify/ui-api-design#1418
- **extensibility host (`eventListenerPlugin` + `eventDispatchFactory`):** Shopify/extensibility#1064
- **pos-mobile wiring:** shop/world#508117, shop/world#508119
- **Observe target deprecation:** Shopify/ui-extensions#4146

## Status

Draft until TAG approves Shopify/ui-api-design#1418. Types will publish with the 2026-07 quarterly release train.

### DIFF
--- a/packages/ui-extensions-tester/src/index.ts
+++ b/packages/ui-extensions-tester/src/index.ts
@@ -201,10 +201,7 @@ class Extension<T extends AnyExtensionTarget> implements ExtensionHarness<T> {
     (globalThis as any).navigation = this.#navigationImpl;
   }
 
-  #addEventListener = (
-    type: string,
-    listener: (event: any) => void,
-  ): void => {
+  #addEventListener = (type: string, listener: (event: any) => void): void => {
     let set = this.#eventListeners.get(type);
     if (!set) {
       set = new Set();
--- a/packages/ui-extensions-tester/src/index.ts
+++ b/packages/ui-extensions-tester/src/index.ts
@@ -1,7 +1,11 @@
 import * as fs from 'fs';
 import * as path from 'path';
 
-import type {AnyExtensionTarget, ApiForTarget} from './targets';
+import type {
+  AnyExtensionTarget,
+  ApiForTarget,
+  EventMapForTarget,
+} from './targets';
 import {isCheckoutTarget} from './targets';
 import {createMockTargetApi} from './mocks/target-apis';
 import {createMockNavigation, type Navigation} from './navigation';
@@ -99,12 +103,26 @@ interface BaseExtensionHarness<T extends AnyExtensionTarget> {
    * are ignored, and thrown errors are caught per-listener so one bad
    * listener doesn't block the others.
    *
+   * The `event` argument must be a real `Event` instance, matching what
+   * the host dispatches at runtime. Construct it with `new Event(type)`
+   * and attach payload fields via `Object.assign`:
+   *
    * ```ts
    * shopify.addEventListener('transactioncomplete', (event) => { ... });
-   * extension.dispatch('transactioncomplete', { transaction: {...} });
+   *
+   * const event = Object.assign(new Event('transactioncomplete'), {
+   *   transactionType: 'Sale',
+   *   orderId: 1,
+   *   grandTotal: { amount: 10, currency: 'USD' },
+   *   // ...
+   * });
+   * extension.dispatch('transactioncomplete', event);
    * ```
    */
-  dispatch(type: string, event?: unknown): void;
+  dispatch<K extends keyof EventMapForTarget<T>>(
+    type: K,
+    event: EventMapForTarget<T>[K],
+  ): void;
 }
 
 /**
@@ -217,8 +235,11 @@ class Extension<T extends AnyExtensionTarget> implements ExtensionHarness<T> {
     this.#eventListeners.get(type)?.delete(listener);
   };
 
-  dispatch(type: string, event?: unknown): void {
-    const listeners = this.#eventListeners.get(type);
+  dispatch<K extends keyof EventMapForTarget<T>>(
+    type: K,
+    event: EventMapForTarget<T>[K],
+  ): void {
+    const listeners = this.#eventListeners.get(type as string);
     if (!listeners) return;
     // Snapshot so listeners that register/unregister during dispatch
     // don't mutate the iteration.
--- a/packages/ui-extensions-tester/src/index.ts
+++ b/packages/ui-extensions-tester/src/index.ts
@@ -90,6 +90,21 @@ interface BaseExtensionHarness<T extends AnyExtensionTarget> {
    * ```
    */
   navigation: Navigation;
+
+  /**
+   * Fires a host event at every listener registered via
+   * `shopify.addEventListener(name, listener)`.
+   *
+   * Matches the `shopify.addEventListener` contract: listener return values
+   * are ignored, and thrown errors are caught per-listener so one bad
+   * listener doesn't block the others.
+   *
+   * ```ts
+   * shopify.addEventListener('transactioncomplete', (event) => { ... });
+   * extension.dispatch('transactioncomplete', { transaction: {...} });
+   * ```
+   */
+  dispatch(type: string, event?: unknown): void;
 }
 
 /**
@@ -138,6 +153,7 @@ class Extension<T extends AnyExtensionTarget> implements ExtensionHarness<T> {
   #previousFetch: typeof globalThis.fetch | undefined;
   #navigationImpl: Navigation = createMockNavigation();
   #previousNavigation: any;
+  #eventListeners = new Map<string, Set<(event: any) => void>>();
 
   constructor(target: T, options?: {configSearchDir?: string}) {
     const configSearchDir =
@@ -174,11 +190,49 @@ class Extension<T extends AnyExtensionTarget> implements ExtensionHarness<T> {
     this.#previousFetch = (globalThis as any).fetch;
     this.#previousNavigation = (globalThis as any).navigation;
     this.#navigationImpl = createMockNavigation();
+    this.#eventListeners.clear();
     (globalThis as any).shopify = deepWritableProxy(
-      createMockTargetApi(this.#target),
+      Object.assign(createMockTargetApi(this.#target), {
+        addEventListener: this.#addEventListener,
+        removeEventListener: this.#removeEventListener,
+      }),
     );
     (globalThis as any).fetch = this.#fetchImpl;
     (globalThis as any).navigation = this.#navigationImpl;
+  }
+
+  #addEventListener = (
+    type: string,
+    listener: (event: any) => void,
+  ): void => {
+    let set = this.#eventListeners.get(type);
+    if (!set) {
+      set = new Set();
+      this.#eventListeners.set(type, set);
+    }
+    set.add(listener);
+  };
+
+  #removeEventListener = (
+    type: string,
+    listener: (event: any) => void,
+  ): void => {
+    this.#eventListeners.get(type)?.delete(listener);
+  };
+
+  dispatch(type: string, event?: unknown): void {
+    const listeners = this.#eventListeners.get(type);
+    if (!listeners) return;
+    // Snapshot so listeners that register/unregister during dispatch
+    // don't mutate the iteration.
+    for (const listener of [...listeners]) {
+      try {
+        listener(event);
+      } catch {
+        // Fire-and-forget: per the shopify.addEventListener contract,
+        // listener errors must not affect other listeners.
+      }
+    }
   }
 
   get shopify(): Mutable<ApiForTarget<T>> {
@@ -233,6 +287,7 @@ class Extension<T extends AnyExtensionTarget> implements ExtensionHarness<T> {
       document.body.innerHTML = '';
     }
     delete (globalThis as any).shopify;
+    this.#eventListeners.clear();
     if (this.#previousFetch === undefined) {
       delete (globalThis as any).fetch;
     } else {
--- a/packages/ui-extensions/src/surfaces/point-of-sale/events/transaction-complete-event.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/events/transaction-complete-event.ts
@@ -1,0 +1,155 @@
+import type {Customer, Discount, LineItem} from '../types/cart';
+import type {Money} from '../types/money';
+import type {Payment} from '../types/payment';
+import type {ShippingLine} from '../types/shipping-line';
+import type {TaxLine} from '../types/tax-line';
+
+/**
+ * Shared fields on every transaction-complete event, available regardless of
+ * `transactionType`.
+ */
+interface BaseTransactionCompleteEvent extends Event {
+  /**
+   * The transaction type identifier indicating which kind of transaction was completed (for example, `'Sale'` for new purchases, `'Return'` for refunds, `'Exchange'` for item swaps). Narrow on this field to access transaction-type-specific properties.
+   */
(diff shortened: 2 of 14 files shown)


vctrchu:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/4123?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#4123** <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/4123?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/4123?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4124** <a href="https://app.graphite.com/github/pr/Shopify/ui-extensions/4124?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `2026-04-rc`


<h2></h2>

NathanJolly (commented):
This PR has the target definition and DataTargetApi, but it's missing two things that should ship together:

1. **addEventListener / removeEventListener** — needs to be part of `DataTargetApi` (or the target's API surface). The generic types are defined in the TAG proposal (ui-api-design PR #1418), and the surface-specific `PosEventMap` with concrete event types + payloads should be defined here.

2. **PosEventMap** — the typed event map with `transaction_complete`, `cash_tracking_session_start`, `cash_tracking_session_complete` and their payload interfaces. This is what makes addEventListener type-safe for POS.

Without these, extensions targeting `pos.app.ready.data` have no way to type-safely listen for events.

vctrchu:
Added both. `DataTargetApi` now includes `addEventListener` and `removeEventListener` using the generic types from the TAG proposal (ui-api-design #1418). Also added `PosEventMap` with `transaction_complete`, `cash_tracking_session_start`, and `cash_tracking_session_complete` — no `cart_update` since that's state, not an event.

packages/ui-extensions/src/surfaces/point-of-sale/events/transaction-complete-event.ts
```diff
@@ -0,0 +1,155 @@
...
+ */
+interface BaseTransactionCompleteEvent extends Event {
+  /**
+   * The transaction type identifier indicating which kind of transaction was completed (for example, `'Sale'` for new purchases, `'Return'` for refunds, `'Exchange'` for item swaps). Narrow on this field to access transaction-type-specific properties.
+   */
+  readonly transactionType: 'Sale' | 'Return' | 'Exchange';
+  /**
+   * The unique numeric identifier for the Shopify order created by this transaction. This ID links the POS transaction to the order record in Shopify's system and can be used for order lookups, tracking, and API operations. Returns `undefined` when order creation is pending.
+   */
+  readonly orderId?: number;
```
js-goupil, line 19: Should this be Order { id: number } like customer?

packages/ui-extensions/src/surfaces/point-of-sale/globals.ts
```diff
@@ -1,6 +1,39 @@
...
+ *
+ * @publicDocs
+ */
+export interface ShopifyGlobal {
+  /**
+   * Register a listener for a POS host event. Listeners are fire-and-forget:
+   * their return values are ignored, and their errors are caught without
+   * affecting the host or other listeners.
+   */
+  addEventListener<K extends keyof ShopifyEventMap>(
```
js-goupil, line 26: This might override the addEventListener that we create in navigation-api.ts. If anything, we should move `'currententrychange` key into the `ShopifyEventMap` in events.ts

js-goupil (commented):
Couple of tiny things that we should address before merging so we don't break the navigation listener

packages/ui-extensions/src/surfaces/point-of-sale/globals.ts
vctrchu: These are on two different globals: shopify.addEventListener ([globals.ts:16](https://github.com/Shopify/ui-extensions/blob/3f13e3fed8dea9cf7b4f300143bdc70d9379f43f/packages/ui-extensions/src/surfaces/point-of-sale/globals.ts#L16-L19)) vs navigation.addEventListener ([navigation-api.ts:64](https://github.com/Shopify/ui-extensions/blob/3f13e3fed8dea9cf7b4f300143bdc70d9379f43f/packages/ui-extensions/src/surfaces/point-of-sale/api/navigation-api/navigation-api.ts#L64-L67)). No TS-level override; the nav listener keeps its own signature.

Unifying currententrychange under ShopifyEventMap is worth considering, but it'd touch the existing navigation API and depends on how the host dispatches (Shopify/extensibility#1064). I'd rather keep this PR types-only for the new data target and open a separate thread for unification if we want it

packages/ui-extensions/src/surfaces/point-of-sale/events/transaction-complete-event.ts
vctrchu: Kept it raw to match the existing shape. BaseTransactionComplete already uses orderId?: number ([base-transaction-complete.ts:20](https://github.com/Shopify/ui-extensions/blob/3f13e3fed8dea9cf7b4f300143bdc70d9379f43f/packages/ui-extensions/src/surfaces/point-of-sale/types/base-transaction-complete.ts#L19-L21)), and the sibling IDs on this same event (refundId, returnId, exchangeId) are all raw numbers too. 

Customer is the outlier in POS, it seems like every other ID field (productId, variantId, customerId, shopId, locationId, etc.) is a raw number

runmad:
**Review summary: NEEDS CHANGES** (one HIGH, two MEDIUMs). Type design is solid overall, discriminated union on `transactionType`, `readonly` fields, clean event/API separation.

### HIGH, Event name strings are an untested contract with the runtime
`ShopifyEventMap` uses `transactioncomplete`, `cashtrackingsessionstart`, `cashtrackingsessioncomplete`. These type-check regardless of what the host dispatches, so a mismatch is silent: listeners compile but never fire. Please confirm the exact strings against the host `eventDispatchFactory` / mobile wiring, export them as named constants, and add a contract test that registers a listener and verifies dispatch reaches it.

### MEDIUM, No mock `shopify` global in the tester harness
`createDataTargetMock` returns a correct `DataTargetApi`, but extension authors will call `shopify.addEventListener` in their code. Without a tester-provided `shopify` shim, tests throw at runtime. Add a mock `shopify` global with `addEventListener`/`removeEventListener` and a dispatch helper so tests can fire events deterministically.

### MEDIUM, `@ts-ignore` on the `shopify` global declaration is fragile
The `@ts-ignore` suppressing the collision with `build/ts/globals.d.ts` hides real merge conflicts. If another surface redeclares `shopify` with a different shape, precedence is compile-order-dependent with no error. Prefer `declare global { interface }` merging or a shared base type across surfaces.

### LOW (optional follow-ups)
- `addEventListener` lacks `options?: AddEventListenerOptions`, adding it now avoids a breaking change later for `once` / `AbortSignal`.
- `CashTrackingSessionCompleteEvent` duplicates `id` and `openingTime` from the start event instead of extending it.
- Naming inconsistency: new `ShopifyEventMap` uses lowercase keys while legacy `event/data.ts` uses PascalCase and now re-exports the new types, creating two parallel hierarchies. Worth a doc note on the canonical path.

Happy to re-review once the HIGH and at least the tester-mock MEDIUM are addressed.

runmad (changes requested):
See comments ^

vctrchu:
@runmad 

Thanks for the review. Went through each item:

HIGH — event-name contract. POS_EVENT_NAMES now exported from point-of-sale/events.ts and threaded through the generated pos.app.ready.data d.ts; ShopifyEventMap keys off the constants. 

MEDIUM — tester mock. Done. addEventListener / removeEventListener on the mock shopify global plus extension.dispatch(type, event) on the harness. 10 tests cover register/fire/remove, error isolation, dedup, teardown. Validated end-to-end against my own test extension.

MEDIUM — @ts-ignore. Not fixed. Tightened the comment to explain the collision, but the structural fix (shared base + per-surface declaration merging) is a cross-surface refactor outside the scope of this PR.

LOW — AddEventListenerOptions. Skipping. Adding an optional third param later is non-breaking.

LOW — Cash session duplication. Fixed. Shared CashTrackingSessionEvent base.

LOW — parallel hierarchies. Legacy event/data.ts types aren't imported anywhere in-package. We'll be deprecating in a follow-up PR

runmad (approved):
**Re-review: APPROVED** ✅

Thanks for the thorough follow-up. Each item addressed:

**HIGH — event-name contract.** `POS_EVENT_NAMES` constants + `ShopifyEventMap` keyed off them is the right shape. One follow-up worth tracking: the new tester tests are self-referential (tester dispatches, tester listens), so they don't yet verify host-side dispatch actually reaches listeners. When the mobile dispatcher wiring lands, add a contract test that exercises the real host path.

**MEDIUM — tester mock.** Covered. `shopify.addEventListener` / `removeEventListener` on the mock global plus `extension.dispatch` helper, with 10 tests for register/fire, fan-out, cross-event isolation, remove, dedup, error isolation, teardown cleanup, and no-op on unregistered. Good coverage.

**MEDIUM — `@ts-ignore` on `shopify` global.** Deferred is fine. The structural fix (shared base + per-surface declaration merging) is a legitimately cross-surface refactor. The tightened comment makes the intent explicit. Worth a follow-up ticket.

**LOWs.** Cash session base shared ✓. `AddEventListenerOptions` deferred (non-breaking) ✓. Legacy `event/data.ts` deprecation in follow-up PR ✓.

Re: @js-goupil's navigation collision concern — agree with Victor's reply, different globals (`shopify.` vs `navigation.`), no TS-level override. Unifying `currententrychange` into `ShopifyEventMap` is reasonable future work but out of scope here.

Two optional follow-ups to track after merge:
1. Host-contract test for event dispatch once the dispatcher wiring lands
2. Cross-surface `shopify` global declaration cleanup to remove the `@ts-ignore`

runmad (approved):

packages/ui-extensions-tester/src/index.ts
```diff
@@ -90,6 +90,21 @@ interface BaseExtensionHarness<T extends AnyExtensionTarget> {
...
+   * Fires a host event at every listener registered via
+   * `shopify.addEventListener(name, listener)`.
+   *
+   * Matches the `shopify.addEventListener` contract: listener return values
+   * are ignored, and thrown errors are caught per-listener so one bad
+   * listener doesn't block the others.
+   *
+   * ```ts
+   * shopify.addEventListener('transactioncomplete', (event) => { ... });
+   * extension.dispatch('transactioncomplete', { transaction: {...} });
```
aaronschubert0: @vctrchu this is throwing me a little. Is this just assigned to the extension because of how we've structured our `ui-extensions-tester` or in what scenario would an extension dispatch host specific events?

packages/ui-extensions-tester/src/index.ts
```diff
@@ -90,6 +90,21 @@ interface BaseExtensionHarness<T extends AnyExtensionTarget> {
...
+   * ```
+   */
+  dispatch(type: string, event?: unknown): void;
```
aaronschubert0: Are we able to type this more strictly to match the `ShopifyEventMap` that was defined in `events.ts`?

aaronschubert0 (commented):
Couple of questions :slightly_smiling_face: otherwise looks good.

packages/ui-extensions/src/surfaces/point-of-sale/events/transaction-complete-event.ts
js-goupil: It would make sense then to refactor them all together in a future version and deprecate these fields. I just don't think the way we currently do things is correct

js-goupil (approved):
Looks great! We can refactor the ID stuff later

packages/ui-extensions-tester/src/index.ts
vctrchu: The`extension.dispatch()` is purely a test-side method to simulate the host firing an event, an extension doesn't ever use it.

Checkout outlines this in their [README](https://github.com/Shopify/ui-extensions/blob/3f13e3fed8dea9cf7b4f300143bdc70d9379f43f/packages/ui-extensions-tester/README.md#%EF%B8%8F-triggering-events)

packages/ui-extensions-tester/src/index.ts
vctrchu: Good call, yea I'll update it to use the real types

PR #4125, opened 2026-03-17
fix(docs): Rename to Order Status Localization API for correct sorting

sordaz00:
## Summary
Renames "Localization (Order Status) API" to "Order Status Localization API" to fix alphabetical sorting in the docs navigation.

## Problem
The parentheses in "Localization (Order Status) API" caused the API to sort incorrectly - it appeared after "Shop API" instead of in its correct alphabetical position.

## Solution
Rename to "Order Status Localization API" which:
- Generates clean slug: `order-status-localization-api`
- Sorts correctly under "O" (after Note API, before Shop API)
- Removes problematic parentheses from the name

## Test plan
- [ ] Verify docs generate without errors
- [ ] Confirm API appears in correct alphabetical order in navigation


mcvinci (approved):
So beautiful!

PR #4147, opened 2026-03-18
Fix missed Sheet.doc.ts category rename

rennyG:
Missed checkout-specific `Sheet.doc.ts` in #4143. The shared component was updated but this surface-specific override at `src/surfaces/checkout/components/Sheet/Sheet.doc.ts` still had `category: 'Polaris web components'`.


mcvinci (approved):
🚀

PR #4149, opened 2026-03-18
Remove `ui` from StandardApi in checkout UI extensions 2026-01

lrsterrett:
## Summary
- Removes the `ui` property, `Ui` interface, and `Overlay` interface from `StandardApi` in checkout UI extensions
- Removes the associated `Docs_Standard_Ui` doc interface

The `ui.overlay.close()` API no longer works for closing modals starting with API version 2025-10.

Corresponding world/shopify-dev [PR](https://github.com/shop/world/pull/522093)
Corresponding world/checkout-web [PR](https://github.com/shop/world/pull/502645)

## Tophatting
- See that existing ui extension code has types for `ui` property on `shopify` global api
- See that with `"@shopify/ui-extensions": "0.0.0-snapshot-20260319134601"` shows no `ui` field as part of `shopify` global

Before
image

After
image

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jamesvidler (dismissed):

jamesvidler (approved):
I dismissed an earlier review since this would be removing a type for a stable version and thought it could cause a regression in UI extenions. But, its just a type and it already didn't do anything. I think its fine to remove.

lrsterrett:
> I dismissed an earlier review since this would be removing a type for a stable version and thought it could cause a regression in UI extenions. But, its just a type and it already didn't do anything. I think its fine to remove.

Thanks @jamesvidler 🙌  Yea, unfortunately we just missed this and it never worked since 2025-10

avocadomayo (commented):
Should we also remove usage from examples? This may not be all of them:

- https://github.com/Shopify/ui-extensions/blob/2026-01/packages/ui-extensions/docs/surfaces/checkout/reference/examples/purchase.checkout.shipping-option-item.render-after/default.example.tsx
- https://github.com/Shopify/ui-extensions/blob/2026-01/packages/ui-extensions/docs/surfaces/checkout/reference/examples/ui-components/overlay-close.example.ts

igor10k (approved):

igor10k:
@avocadomayo great catch!

@lrsterrett those examples should be updated to use the new api. Something like `<s-button command="--hide" commandFor="my-modal">`. In their current form the close button most probably doesn't work there.

lrsterrett:
Yes, great catch @avocadomayo I thought we had already removed all code references to ui.overlay.close()

I changed the one. The other is no longer referenced so just deleted it

PR #4169, opened 2026-03-18
[App Home]: Fix docs build script

mcvinci:
## This PR: 
- Fixes duplicate Polaris web components in generated App Home docs: The `transformJson` merge filter was missing the "Polaris web components" category, causing entries from the Shopify Dev docs source to duplicate those already produced by the doc generator.
- Adds "Polaris web components" to the category exclusion list so only the freshly generated entries are included in the final output, matching the existing handling of "Web components" and "Patterns".


timtrevor-shopify (approved):
Thanks for taking care of this, Michelle!

PR #4242, opened 2026-03-31
fix: post-process generated_docs_data_v2.json in checkout build 2026-04-rc

rennyG:
## Summary

- Adds `generated_docs_data_v2.json` to the URL-stripping `run_sed` call (strips `https://shopify.dev` to make URLs relative)
- Adds `generated_docs_data_v2.json` to the version-replacement `for` loop (replaces `/latest` with actual API version)

The v2 file was being generated but skipped by both post-processing steps, leaving ~74 absolute URLs that caused staging links to point to production shopify.dev.

## Test plan

- [x] Ran `yarn docs:checkout 2026-04-rc` (generate-docs 1.1.0 already on this branch)
- [x] Verified `grep -c 'https://shopify.dev' generated_docs_data_v2.json` returns 0 (v2 URLs now relative)
- [x] Verified `grep -c 'https://shopify.dev' generated_docs_data.json` returns 0 (v1 still correct)

Also opened against `2026-01` (#4243) and `2025-10` (#4244).


packages/ui-extensions/docs/surfaces/checkout/build-docs.sh
jgray7019: In the other versions I just removed the `--delete` to stop removing images if you want to update that in here 🙏

jgray7019 (approved):

packages/ui-extensions/docs/surfaces/checkout/build-docs.sh
rennyG: done!

PR #4243, opened 2026-03-31
fix: post-process generated_docs_data_v2.json in checkout build 2026-01

rennyG:
## Summary

- Adds `generated_docs_data_v2.json` to the URL-stripping `run_sed` call (strips `https://shopify.dev` to make URLs relative)
- Adds `generated_docs_data_v2.json` to the version-replacement `for` loop (replaces `/latest` with actual API version)

The v2 file was being generated but skipped by both post-processing steps, leaving absolute URLs that caused staging links to point to production shopify.dev.

## Test plan

- [x] Temporarily bumped generate-docs from 0.19.8 to 1.1.0 (not committed) to produce v2 output
- [x] Ran `yarn docs:checkout 2026-01`
- [x] Verified `grep -c 'https://shopify.dev' generated_docs_data_v2.json` returns 0 (v2 URLs now relative)
- [x] Verified `grep -c 'https://shopify.dev' generated_docs_data.json` returns 0 (v1 still correct)

Also opened against `2026-04-rc` (#4242) and `2025-10` (#4244).


jgray7019 (approved):
🎩 and works!

PR #4251, opened 2026-04-02
Add types for admin.app.intent.render

JoviDeCroock:
### Background

This adds the types for supporting admin.app.intent.render

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/extension-targets.ts
+++ b/packages/ui-extensions/src/surfaces/admin/extension-targets.ts
@@ -13,6 +13,7 @@ import type {
   CustomerSegmentTemplateApi,
   CustomerSegmentTemplate,
   StandardApi,
+  IntentRenderApi,
 } from './api';
 import {
   ShouldRenderApi,
@@ -22,6 +23,7 @@ import type {BlockExtensionComponents} from './components/BlockExtensionComponen
 import type {ActionExtensionComponents} from './components/ActionExtensionComponents';
 import type {PrintActionExtensionComponents} from './components/PrintActionExtensionComponents';
 import type {FunctionSettingsComponents} from './components/FunctionSettingsComponents';
+import type {FormExtensionComponents} from './components/FormExtensionComponents';
 
 /**
  * Maps extension target identifiers to their corresponding extension types. Each target represents a specific location or context in the Shopify admin where extensions can render or execute. Use these targets to define where your extension appears and what capabilities it has access to.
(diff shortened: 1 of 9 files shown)


packages/ui-extensions/src/surfaces/admin/api/tools/tools.ts
```diff
@@ -0,0 +1,28 @@
...
+) => Record<string, any> | Promise<Record<string, any>>;
+
+/**
+ * The `Tools` object provides methods for registering and removing runtime handlers for app tools declared in your extension configuration. Use this API to expose app data lookups and app actions that Sidekick can invoke on behalf of merchants.
+ */
+export interface Tools {
+  /**
+   * Registers a tool handler for your app. If a handler with the same name already exists, it is replaced. Returns a cleanup function that unregisters the tool.
+   */
+  register: (name: string, handler: ToolHandler) => () => void;
```
olavoasantos: We should make sure this won't interfere with the autogenerated types

packages/ui-extensions/src/surfaces/admin/extension-targets.ts
```diff
@@ -665,6 +667,14 @@ export interface ExtensionTargets {
     StandardApi<'admin.app.tools.data'>,
     undefined
   >;
+
+  /**
+   * Renders an admin extension for handling app intents.
+   */
+  'admin.app.intent.render': RenderExtension<
+    StandardRenderingExtensionApi<'admin.app.intent.render'>,
```
JoviDeCroock: CC @vividviolet should we use a custom type here so we can do `data: {intent:Input}` or is that not needed?

vividviolet: Yeah, I think we need a base type for intents with `data.intent`and `shopify.intents`for this target. I think we can merge this first and then do a patch follow up because we also want typechecking to work with the input and output schema

packages/ui-extensions/src/surfaces/admin/components/IntentExtensionComponents.ts
```diff
@@ -0,0 +1,9 @@
+import {StandardComponents} from './StandardComponents';
```
vividviolet: I think we can re-use the `FormExtensionComponents` for now instead of creating a new one

vividviolet (approved):
Looks good to me! Just need to add a changeset

JoviDeCroock:
@vividviolet there is one in https://github.com/Shopify/ui-extensions/blob/5226e0245c64681ec4e60c911ffc89ec9c428d71/.changeset/add-admin-app-intent-render-target.md

vividviolet:
Perfect

PR #4279, opened 2026-04-08
add public docs to ALL top level types and rerun the 4 reference doc …

laurelthorburn:
### Background

Screenshot 2026-04-08 at 14 45 28

I especially loved chasing down `SelectPayload`

image

---------------------------------

No ticket to close, related to our gen docs v1 migration creating the gen_docs_data_v2.json files for all 4 references. 

The v2 docs generator (`@shopify/generate-docs`) uses the `@publicDocs` JSDoc tag as entry points, then transitively follows type references to build `generated_docs_data_v2.json`. Many top-level types that developers encounter in API signatures (like `SelectPayload`, `ResourcePickerOptions`, `Product`, `Collection`, etc.) were only appearing in the generated docs because they were *transitively referenced* by another `@publicDocs` type -- they didn't have their own `@publicDocs` tag, and some weren't even exported.

This meant these types were fragile: if the parent type that referenced them changed, they could silently disappear from docs. It also meant developers couldn't import some of these types directly.

### Solution

Added `export` and `@publicDocs` to all top-level API types across **all four surfaces** (admin, checkout, POS, customer-account). "Top-level" means named types that appear in API signatures developers encounter -- parameters, return types, properties of API objects.

**What changed:**
- ~537 types updated across ~80+ files
- Types that were already exported but missing `@publicDocs` got the tag added
- Types that were not exported (like `SelectPayload`, `Auth`, `PickupLocation`) got both `export` and `@publicDocs`
- Internal helper types (`WithSelection`, `ResourceTypes`, conditional type extractors) were intentionally left unchanged

**Impact on generated docs (strictly additive, zero types removed):**

| Surface | Types before | Types after | New |
|---------|-------------|-------------|-----|
| Admin | 367 | 387 | +20 |
| Checkout | 629 | 647 | +18 |
| Customer Account | 585 | 632 | +47 |
| POS | 162 | 221 | +59 |

The new types appearing in the JSON are ones that were previously only pulled in transitively and now have their own explicit `@publicDocs` entry point, making them resilient to upstream refactors.

TypeScript compilation passes cleanly with no errors. All four docs build scripts (`yarn docs:admin`, `yarn docs:checkout`, `yarn docs:customer-account`, `yarn docs:point-of-sale`) run successfully for `2026-04-rc`.

### :tophat:

- Verified TypeScript compiles with zero errors (`npx tsc --project packages/ui-extensions/tsconfig.json`)
- Ran all four docs build commands for `2026-04-rc` and confirmed successful generation
- Compared `generated_docs_data_v2.json` before/after: strictly additive (no types removed)
- Spot-checked that internal helper types (e.g., `WithSelection`, `ResourceTypes`) were NOT tagged

### Checklist


mcvinci (approved):
Thanks, @laurelthorburn!

PR #4280, opened 2026-04-09
Add missing Navigation and Intents API doc wrapper types (2026-04-rc)

sordaz00:
Adds Docs_Standard_NavigationApi and Docs_Standard_IntentsApi wrapper types to docs.ts. These were missing, causing the Navigation and Intents API pages to render without properties in the v2 documentation.


mcvinci (approved):

sordaz00:
The Navigation and Intents types are showing on stage8 without the wrapper types. The v2 rendering system is pulling them directly from the StandardApi type resolution, not from Docs_Standard_* wrappers.

This means:

Your fix branch wrappers were a v1 approach that v2 doesn't need
The Navigation/Intents props are showing because the v2 generator extracts them directly from the StandardApi interface in standard-api.ts
You don't need to merge the fix PRs — the v2 system handles this natively
Closing the PRs won't break anything since the wrappers were never on the publish branch

PR #4294, opened 2026-04-13
bump admin-ui-components 1.63.1

kyledurand:
Bump polaris components for docs


mcvinci (changes requested):
Hey @kyledurand! It looks like bumping `admin-ui-components` will override a lot of the description improvements we recently made for the Admin UI extensions reference. As a result, I don't think we should merge this as-is. 

Would you mind reaching out to our engineers in  [#devtools-proj-templated-refs](https://shopify.enterprise.slack.com/archives/C099RJCHAKE) so they can help investigate?

kyledurand:
Closing in favour of https://github.com/Shopify/ui-extensions/pull/4302 which bumps to 1.64.0

PR #4302, opened 2026-04-15
bump admin-ui-components 1.64.0

kyledurand:
Supercedes: https://github.com/Shopify/ui-extensions/pull/4294

Generated by backporting all relevant jsdocs to polaris in https://app.graphite.com/github/pr/shop/world/614173


mcvinci (commented):
Hey Kyle! 

Sorry for dropping the ball on this; I got pulled into a different project recently, so have been ramping up in that space. I took another look just now, but I'm still noticing a lot of descriptions that are either being removed entirely, or are being replaced with shorter, sub-optimal alternatives. 

For example, most of what I'm seeing in `packages/ui-extensions/src/surfaces/admin/components/Clickable.d.ts` (and many other docs) are descriptions that have been shortened or reverted back from the improvements made. In files like `packages/ui-extensions/src/surfaces/admin/components/Checkbox.d.ts`, I'm noticing descriptions are being removed entirely. 

I'm still hesitant to ship this as-is, given how much context is being pulled out of the prop descriptions.

PR #4320, opened 2026-04-20
Add ConsentPolicy description to consent component .d.ts files (2026-07-rc)

sordaz00:
Adds inline JSDoc for the policy property so the ConsentPolicy type description flows through the doc generator.


mcvinci (approved):
Thanks, @sordaz00! Same link update suggestions as https://github.com/Shopify/ui-extensions/pull/4322#pullrequestreview-4144286746

PR #4335, opened 2026-04-23
Intent helpers

JoviDeCroock:
### Background

This adds generic types that the CLI can consume to fine-tune the input/output types of intents with what's defined in the schemas

### DIFF
--- a/packages/ui-extensions/src/surfaces/admin/api.ts
+++ b/packages/ui-extensions/src/surfaces/admin/api.ts
@@ -16,3 +16,9 @@ export type {ValidationSettingsApi} from './api/checkout-rules/validation-settin
 export type {PurchaseOptionsCardConfigurationApi} from './api/purchase-options-card-action';
 export type {DiscountFunctionSettingsApi} from './api/discount-function-settings/discount-function-settings';
 export type {IntentRenderApi} from './api/intents/intent-render';
+export type {
+  ShopifyGeneratedIntentResponse,
+  ShopifyGeneratedIntentVariant,
+  WithGeneratedIntents,
+  WithGeneratedTools,
+} from './api/generated/generated';
--- a/packages/ui-extensions/src/surfaces/admin/api/generated/generated.ts
+++ b/packages/ui-extensions/src/surfaces/admin/api/generated/generated.ts
@@ -1,0 +1,76 @@
+/** Response handle on `shopify.intents.response` with a typed `ok` payload. */
+export interface ShopifyGeneratedIntentResponse<Data = unknown> {
+  ok(data?: Data): Promise<void>;
+}
+
+export interface ShopifyGeneratedIntentVariant<
+  Request = unknown,
+  ResponseData = unknown,
+> {
+  request: Request;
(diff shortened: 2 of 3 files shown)


packages/ui-extensions/src/surfaces/admin/api/generated/generated.ts
```diff
@@ -0,0 +1,156 @@
...
+ * One member of the `ShopifyGeneratedIntentVariants` union the CLI emits. The
+ * CLI pairs a generated request interface (discriminated by literal
+ * `action`/`type`) with the generated output type.
+ */
+export interface ShopifyGeneratedIntentVariant<
+  Request = unknown,
+  ResponseData = unknown,
+> {
+  request: Request;
+  response?: ShopifyGeneratedIntentResponse<ResponseData>;
```
JoviDeCroock: Is this expected to be optional? We had this before but I wasn't sure back then

olavoasantos: if we have an intent request we should have a response. At most the expected payload (ResponseData) is optional (e.g response.ok() is valid if no output schema is defined)

packages/ui-extensions/src/surfaces/admin/api/generated/generated.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Utility types the Shopify CLI composes over an extension target's base `Api`
+ * with types derived from user-authored tool and intent schemas.
+ */
+
+/** Response handle on `shopify.intents.response` with a typed `ok` payload. */
+export interface ShopifyGeneratedIntentResponse<Data = unknown> {
+  ok(data?: Data): Promise<void>;
+}
+
+/** One member of the `ShopifyGeneratedIntentVariants` union the CLI emits. */
```
vividviolet: Can we remove these comments about CLI? Technically developers can pull these in if they wanted to use them

packages/ui-extensions/src/surfaces/admin/api.ts
```diff
@@ -16,3 +16,9 @@ export type {ValidationSettingsApi} from './api/checkout-rules/validation-settin
 export type {PurchaseOptionsCardConfigurationApi} from './api/purchase-options-card-action';
 export type {DiscountFunctionSettingsApi} from './api/discount-function-settings/discount-function-settings';
 export type {IntentRenderApi} from './api/intents/intent-render';
+export type {
+  ShopifyGeneratedIntentResponse,
+  ShopifyGeneratedIntentVariant,
+  WithGeneratedIntents,
+  WithGeneratedTools,
+} from './api/generated/generated';
```
vividviolet, line 24: Couldn't we put these inside the `/api/intents/intent-render`? They are just generics

JoviDeCroock: Mainly did it because `WithGeneratedTools` didn't really fit that picture

vividviolet: Ok, I see what you mean. These are overriding the default ones. Good to keep as-is

vividviolet (approved):

PR #4360, opened 2026-04-26
Fix cart-line-item target description and commandFor/command patterns

sordaz00:
## Summary
- Enriched `CartLineItemApi.target` description in customer-account `cart-line-item.ts` to mention merchandise, quantity, cost, and attributes
- Updated POS `InteractionProps`: rewrote `commandFor`/`command` descriptions, converted `@see` tags to inline links, improved option descriptions
- Fixed checkout `command` description to reference `commandFor` target instead of self-referential `command` link

## Test plan
- [ ] `yarn build` passes
- [ ] Verify descriptions render correctly in generated docs


mcvinci (approved):
🚀 Same note about the description for `target: SubscribableSignalLike<CartLine>`.

PR #4362, opened 2026-04-26
Fix cart-line-item target description

sordaz00:
## Summary
- Enriched `CartLineItemApi.target` description in customer-account `cart-line-item.ts` to mention merchandise, quantity, cost, and attributes
- Verified: `commandFor`/`command` properties do not exist on this version — no `InteractionProps` fix needed (no `components-shared.d.ts`, and `shared.ts` has no commandFor patterns)

## Test plan
- [ ] `yarn build` passes
- [ ] Verify descriptions render correctly in generated docs

### DIFF
--- a/packages/ui-extensions/src/surfaces/customer-account/api/cart-line/cart-line-item.ts
+++ b/packages/ui-extensions/src/surfaces/customer-account/api/cart-line/cart-line-item.ts
@@ -4,7 +4,7 @@ import {CartLine} from '../order-status/order-status';
 
 export interface CartLineItemApi {
   /**
-   * The cart line the extension is attached to.
+   * The cart line that this extension is attached to. Use this to read the line item's merchandise, quantity, cost, and attributes.
    */
   target: StatefulRemoteSubscribable<CartLine>;
 }


packages/ui-extensions/src/surfaces/customer-account/api/cart-line/cart-line-item.ts
```diff
@@ -4,7 +4,7 @@ import {CartLine} from '../order-status/order-status';
 
 export interface CartLineItemApi {
   /**
-   * The cart line the extension is attached to.
+   * The cart line that this extension is attached to. Use this to read the line item's merchandise, quantity, cost, and attributes.
```
mcvinci, line 7: I'm not sure if there's another way to describe this prop, but the first sentence still sounds a bit funky to me. Can an extension be attached to a cart line?

mcvinci (approved):

packages/ui-extensions/src/surfaces/customer-account/api/cart-line/cart-line-item.ts
sordaz00: This feedback is valid, but the description is still clear. P3 for potential follow up improvement.

PR #4376, opened 2026-04-28
test: rc-port-tracker workflow against 2026-04

robin-drexler:
Throwaway PR to verify the rc-port-tracker workflow fires on PRs targeting the latest stable branch. Will be closed after verification.


robin-drexler:
Closing — the workflow doesn't trigger on PRs targeting non-default branches until it's merged onto the default branch first. Will test post-merge instead.

PR #4384, opened 2026-04-29
Remove version-dated callouts from API descriptions

sordaz00:
## Summary
Strip "as of version YYYY-MM" from deprecation notices and caution blocks in checkout `standard.ts`. These are migration artifacts — the behavior is now standard and the dates are irrelevant to current developers.

- **`market`**: Removed redundant Caution block; `@deprecated` now explains *why* instead of *when*
- **`instructions`**: Removed version date from Caution; states the requirement directly
- **`extensionPoint`**: `@deprecated` now says what to use instead of when it was deprecated

Closes: https://github.com/shop/issues-learn/issues/2001

## Test plan
- [ ] Verify generated docs no longer show version-dated callouts
- [ ] Cascade to 2026-04, 2026-01, 2025-10, 2025-07

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout/api/standard/standard.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/api/standard/standard.ts
@@ -439,11 +439,14 @@ export interface Localization {
   country: SubscribableSignalLike<Country | undefined>;
 
   /**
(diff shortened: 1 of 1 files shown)


mcvinci (approved):
Thanks, @sordaz00! Same suggestions as https://github.com/Shopify/ui-extensions/pull/4387#pullrequestreview-4200291000.

PR #4418, opened 2026-04-30
Update checkbox docs to support label slot (2025-10)

justinhenricks:
### Background

The `label` prop on the `s-checkbox` component previously only accepted a plain string value, limiting the ability to include rich content such as links within checkbox labels.

Depends on: https://github.com/shop/world/pull/663974

### Solution

The `label` prop on the checkout and customer-account `s-checkbox` component now accepts either a plain string or an `HTMLElement` slot. When passed as an `HTMLElement`, only plain text and `s-link` elements are rendered as label content — any additional elements are sanitized while preserving their text content.

A new `CheckboxElementSlots` interface has been introduced to define this slot behavior, and the `label` prop has been moved out of `CheckboxElementProps` and into `CheckboxElementSlots` to reflect this distinction. The component documentation has also been updated to include slot definitions.

### 🎩

- `s-checkbox` labels can now include `s-link` elements for richer, more accessible label content

### Checklist

### DIFF
--- a/.changeset/checkbox-rich-label.md
+++ b/.changeset/checkbox-rich-label.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+The `label` prop on the checkout and customer-account `s-checkbox` component now accepts a label as a slot in addition to a plain string, label slots can include only plain text and s-links.
--- a/packages/ui-extensions/src/surfaces/checkout/components/Checkbox.d.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/components/Checkbox.d.ts
@@ -10,10 +10,21 @@
 /// <reference lib="DOM" />
 import type {CheckboxProps$1} from './components-shared.d.ts';
 
-/** @publicDocs */
+/**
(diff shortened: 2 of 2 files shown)


packages/ui-extensions/src/surfaces/checkout/components/Checkbox/Checkbox.doc.ts
```diff
@@ -20,6 +20,12 @@ const data: ReferenceEntityTemplateSchema = {
         'Learn more about [registering events](/docs/api/checkout-ui-extensions/latest/using-polaris-components#event-handling).',
       type: 'CheckboxElementEvents',
     },
+    {
+      title: 'Slots',
+      description:
+        'Learn more about [component slots](/docs/api/checkout-ui-extensions/latest/using-polaris-components#slots).',
+      type: 'CheckboxElementSlots',
```
mcvinci: We can remove this change. `doc.ts` files are now considered legacy files, as this content how now moved to MDX in the `shopify-dev` zone of the monorepo.

mcvinci (approved):
Thanks, @justinhenricks! Just left one minor comment.

robin-drexler (approved):

PR #3783, opened 2026-01-30
Fix /snapit trigger in deploy workflow

billfienberg:
### Background

The `/snapit` command wasn't working on PRs targeting the `2026-04-rc` branch. The `issue_comment` trigger was lost when merging from `2025-10` - likely a merge conflict resolution error.

Same fix as #3782 for the `2026-01` branch.

### Solution

Added the missing `issue_comment` trigger to the deploy workflow:

```yaml
issue_comment:
  types:
    - created
```

### Checklist

- [x] I have :tophat:'d these changes

---
🤖 This PR was generated with Claude Code

PR #4471, opened 2026-05-19
ci: gate Deploy changesets job to push events only

andy-chhuon:
## Summary

The `changesets` job in `.github/workflows/deploy.yml` is missing an event-type guard. The workflow runs on both `push` (release branches) and `issue_comment.created` (for `/snapit`). The `snapit` job correctly guards itself with `event_name == 'issue_comment'`, but `changesets` has no `if:` and fires on every `issue_comment.created` event comments from anyone on any issue or PR.

Introduced in #3783 when the `issue_comment` trigger was added; the new job was guarded but the existing `changesets` job was not backfilled with the inverse guard.

## Fix

Symmetric one-line guard mirroring the `snapit` job's pattern:

```yaml
  changesets:
    name: Deploy
    if: ${{ github.event_name == 'push' }}
    runs-on: ubuntu-latest
```

After this change, `issue_comment.created` events fire only the `snapit` job; `push` events to release branches fire only the `changesets` job. Reported via bug bounty.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -56,6 +56,7 @@ jobs:
 
   changesets:
     name: Deploy
+    if: ${{ github.event_name == 'push' }}
     runs-on: ubuntu-latest
     permissions:
       contents: read


robin-drexler:
@andy-chhuon did you perhaps pick the wrong base branch?

andy-chhuon:
@robin-drexler Yes sorry wasnt meant to be ready for review yet

fatbattk (approved):

vctrchu (approved):

vctrchu:
Does 2025-10 or 2025-07 need this as well?

andy-chhuon:
@vctrchu Nope, 2025-10 already has [this](https://github.com/Shopify/ui-extensions/blob/2025-10/.github/workflows/deploy.yml#L56-L59) fix and [2025-07](https://github.com/Shopify/ui-extensions/blob/2025-07/.github/workflows/deploy.yml#L3-L13) doesn't have the issue_comment trigger

PR #4478, opened 2026-05-20
2026 04 v1 cleanup

laurelthorburn:
Part of https://github.com/shop/issues-learn/issues/1601

### Background

Same v1 cleanup as #4470, but targeting `2026-04-rc` so the older active release branch stops dragging the v1 artifacts.

The v1 generator consumes hand-authored `.doc.ts` files and emits `generated_docs_data.json`, `generated_static_pages.json`, and `generated_category_pages.json`. We've fully moved to v2, which reads `@publicDocs` JSDoc tags off the source `.ts` files and emits `generated_docs_data_v2.json`. Keeping the v1 files around invites accidental regressions and confuses contributors.

### Solution

Scoped to `2026-04-rc` (mirror of the cleanup landed on `2026-07-rc` in #4470):

- **Bump `@shopify/generate-docs` to `1.2.2`** — the latest stable release; v2-only CLI surface.
- **Delete all `.doc.ts` files** (~410 files across admin, checkout, customer-account, point-of-sale).
- **Delete v1 generated JSONs** for `2026-04-rc` and surface-level paths: `generated_docs_data.json`, `generated_static_pages.json`, `generated_category_pages.json`.
- **Delete v1 type-helper sidecars in `src/`** that consumed removed exports from generate-docs (`ReferenceEntityTemplateSchema`, `CodeTabType`, `ExampleType`, etc.):
  - `src/docs/` (whole tree — `component-definitions.ts`, `docs-type.ts`, `components/*.ts` metadata sidecars)
  - `src/surfaces/admin/docs-types.ts`
  - `src/surfaces/checkout/helper.docs.ts`
  - `docs/surfaces/checkout/reference/helper.docs.ts`

  These were metadata-only files (name/description/category records) consumed exclusively by the deleted `.doc.ts` files. None of the real component source was touched.
- **Slim the four build scripts** (`admin`, `checkout`, `customer-account`, `point-of-sale`) to v2-only:
  - Drop v1-specific flags (`--isLandingPage`, `--isCategoryPage`, `--typesInput`), the `transformJson` step, `.doc.ts` → `.doc.js` tsc compile, and dead template helpers (htmlWrapper, jsxWrapper, etc.).
  - Drop the `build-docs-fast.sh` / `build-docs-watch.sh` checkout scripts (v1-only).
- **Stop copying screenshots to shopify-dev** in every surface's build script.
[...]

PR #4470, opened 2026-05-19
2026 07 v1 cleanup

laurelthorburn:
Closes https://github.com/shop/issues-learn/issues/1601

### Background

Removing all `@shopify/generate-docs` v1 artifacts from the repo. The v1 generator consumes hand-authored `.doc.ts` files and emits `generated_docs_data.json`, `generated_static_pages.json`, and `generated_category_pages.json`. We've fully moved to v2, which reads `@publicDocs` JSDoc tags off the source `.ts` files and emits `generated_docs_data_v2.json`. The v1 files are dead weight — keeping them around invites accidental regressions and confuses contributors.

### Solution

Scoped to `2026-07-rc` (no changes to `2026-04` artifacts):

- **Bump `@shopify/generate-docs` to `1.2.2`** — the latest stable release; v2-only CLI surface.
- **Delete all `.doc.ts` files** (~410 files across admin, checkout, customer-account, point-of-sale).
- **Delete v1 generated JSONs** for `2026-07-rc` and surface-level paths: `generated_docs_data.json`, `generated_static_pages.json`, `generated_category_pages.json`.
- **Delete v1 type-helper sidecars in `src/`** that consumed removed exports from generate-docs (`ReferenceEntityTemplateSchema`, `CodeTabType`, `ExampleType`, etc.):
  - `src/docs/` (whole tree — `component-definitions.ts`, `docs-type.ts`, `components/*.ts` metadata sidecars)
  - `src/surfaces/admin/docs-types.ts`
  - `src/surfaces/checkout/helper.docs.ts`
  - `docs/surfaces/checkout/reference/helper.docs.ts`

  These were metadata-only files (name/description/category records) consumed exclusively by the deleted `.doc.ts` files. None of the real component source was touched.
- **Slim the four build scripts** (`admin`, `checkout`, `customer-account`, `point-of-sale`) to v2-only:
  - Drop v1-specific flags (`--isLandingPage`, `--isCategoryPage`, `--typesInput`), the `transformJson` step, `.doc.ts` → `.doc.js` tsc compile, and dead template helpers (htmlWrapper, jsxWrapper, etc.).
  - Drop the `build-docs-fast.sh` / `build-docs-watch.sh` checkout scripts (v1-only).
- **Stop copying screenshots to shopify-dev** in every surface's build script.
- **Fix `build-docs-targets-json.mjs`** (admin + POS) to locate output dirs via `generated_docs_data_v2.json` now that `generated_docs_data.json` is gone.
[...]

PR #4479, opened 2026-05-20
2026 01 v1 cleanup

laurelthorburn:
Part of https://github.com/shop/issues-learn/issues/1601

### Background

Same v1 cleanup as #4470 (2026-07-rc) and #4478 (2026-04-rc), but targeting `2026-01` so the oldest active release branch also drops the v1 artifacts.

The v1 generator consumes hand-authored `.doc.ts` files and emits `generated_docs_data.json`, `generated_static_pages.json`, and `generated_category_pages.json`. We've fully moved to v2, which reads `@publicDocs` JSDoc tags off the source `.ts` files and emits `generated_docs_data_v2.json`. Keeping the v1 files around invites accidental regressions and confuses contributors.

### Solution

Scoped to `2026-01-rc` (mirror of the cleanups landed on `2026-07-rc` in #4470 and `2026-04-rc` in #4478):

- **Bump `@shopify/generate-docs` to `1.2.2`** — the latest stable release; v2-only CLI surface. 2026-01-rc was pinned at `0.19.8`.
- **Delete all `.doc.ts` files** (~358 files across admin, checkout, customer-account, point-of-sale).
- **Delete v1 generated JSONs** for `2026-01-rc` and surface-level paths: `generated_docs_data.json`, `generated_static_pages.json`, `generated_category_pages.json`.
- **Delete v1 type-helper sidecars in `src/`** that consumed removed exports from generate-docs (`ReferenceEntityTemplateSchema`, `CodeTabType`, `ExampleType`, etc.):
  - `src/docs/` (whole tree — `component-definitions.ts`, `docs-type.ts`, `components/*.ts` metadata sidecars)
  - `src/surfaces/admin/docs-types.ts`
  - `src/surfaces/checkout/helper.docs.ts`
  - `docs/surfaces/checkout/reference/helper.docs.ts`
  - `docs/surfaces/customer-account/reference/helper.docs.ts`
  - `docs/surfaces/point-of-sale/reference/helpers/helper.docs.ts`

  These were metadata-only files (name/description/category records) consumed exclusively by the deleted `.doc.ts` files. None of the real component source was touched.
- **Slim the four build scripts** (`admin`, `checkout`, `customer-account`, `point-of-sale`) to v2-only:
  - Drop v1-specific flags (`--isLandingPage`, `--isCategoryPage`, `--typesInput`), the `transformJson` step, `.doc.ts` → `.doc.js` tsc compile, and dead template helpers (htmlWrapper, jsxWrapper, etc.).
  - Drop the `build-docs-fast.sh` / `build-docs-watch.sh` checkout scripts (v1-only).
- **Stop copying screenshots to shopify-dev** in every surface's build script.
- **Fix `build-docs-targets-json.mjs`** (admin + POS) to locate output dirs via `generated_docs_data_v2.json` now that `generated_docs_data.json` is gone.
- **Fix `buildTargetDts.ts`** — `copyComponentDefinitions` now `mkdirSync`'s the components build destination. Previously the dir existed only as a side effect of tsc emitting the now-deleted `.doc.ts` files; on a clean CI build with no `.doc.ts` files compiled, `copyFileSync` failed with ENOENT.
- **Fix `yarn.lock`** — the `@shopify/generate-docs@1.2.2` `resolved` URL had been pinned to `npm.shopify.io` (Shopify's internal registry) because of local `.npmrc` scope routing. Rewrote it to `registry.npmjs.org` so CI (which can't auth to the internal mirror) can install. Integrity sha512 is identical between registries.
- **Leave scaffolding alone** — `create-doc-files.js`, the `gen-docs:admin` workspace script, and `tsconfig.docs.json` are kept since they're authoring tools, not v1 build artifacts.

### 🎩

- \`yarn docs:admin 2026-01`
- \`yarn docs:checkout 2026-01`
- \`yarn docs:customer-account 2026-01`
- \`yarn docs:point-of-sale 2026-01`

Each should produce only \`generated_docs_data_v2.json\` + \`targets.json\` under the surface's versioned output dir, with no v1 files anywhere.

CI parity locally (clean \`build/\` removed first): \`yarn lint\`, \`yarn type-check\`, \`yarn build\`, \`yarn test --no-watch\` all pass.

### Checklist


jgray7019 (commented):
Can we delete all of the `generated_docs_data.json` and `generated_static_pages.json` files plz 🧹

packages/ui-extensions/docs/surfaces/admin/build-docs.mjs
```diff
@@ -30,479 +34,22 @@ const worldDBPath = path.join(
 );
 const worldExists = existsSync(worldPath);
 
-const generatedDocsDataFile = 'generated_docs_data_v2.json';
```
johndcollett, line 33: are we good to remove all this logic?

laurelthorburn: Yep — all safe to remove 🤖

That whole block (decodeHTML / escapeForJSTemplate / composeStyles / allowedProperties / stylesToString / renderIconPreviewJsxTemplate / jsxWrapper) was the v1 self-hosted Preact + Sucrase live-preview system that hydrated <jsx-render> islands via cdn.shopify.com/shopifycloud/jsx-builder/. It only ran against the .doc.ts files (now deleted) and the templates/icon-renderer/jsx-render.html template (also gone). v2 docs (@publicDocs JSDoc on real source files → generated_docs_data_v2.json) don't go through any of it, and Polaris previews on shopify.dev are served by a separate system.

— Laurel's robot assistant 🤖

johndcollett (approved):

PR #4455, opened 2026-05-13
Add textAlign to Paragraph and improve JSDoc

justinhenricks:
## TLDR

Add `textAlign` to `ParagraphElementProps` and replace the `type` prop, aligning ui-extensions types with the underlying component change.

## What changed

- **Added** `textAlign` prop (`'start' | 'end' | 'center' | 'auto'`) to `ParagraphElementProps`
- **Removed** `type` prop (`'paragraph' | 'small'`)
- **Added** `@publicDocs` annotations to `ParagraphElement` and `ParagraphProps`
- **Improved** JSDoc for `BaseElementProps`, `BaseElementPropsWithChildren`, and their properties

## Context

The underlying `s-paragraph` web component change shipped in https://github.com/shop/world/pull/577805, which wired `textAlign` through the Polaris web component pipeline. This PR updates the ui-extensions type declarations to match.

PR #4480, opened 2026-05-20
Add textAlign to Paragraph and improve JSDoc

justinhenricks:
Backport of https://github.com/Shopify/ui-extensions/pull/4455 to 2025-10.

## TLDR
Add `textAlign` to `ParagraphProps` and replace the `type` prop, aligning ui-extensions types with the underlying component change.

## What changed
- **Added** `textAlign` prop (`'start' | 'end' | 'center' | 'auto'`) to `ParagraphProps`
- **Removed** `type` prop (`'paragraph' | 'small'`)
- **Improved** JSDoc for `BaseElementProps`, `BaseElementPropsWithChildren`


packages/ui-extensions/docs/surfaces/checkout/generated/targets.json
```diff
@@ -2723,19 +2723,43 @@
   "purchase.address-autocomplete.suggest": {
     "components": [],
     "apis": [
-      "AddressAutocompleteStandardApi",
-      "AddressAutocompleteSuggestApi"
+      "AddressAutocompleteSuggestApi",
```
robin-drexler, line 2726: @mcvinci @rennyG can you judge if that's ok?

robin-drexler (approved):

packages/ui-extensions/docs/surfaces/checkout/generated/targets.json
rennyG: @johndcollett is this related to the buggy behaviour we were seeing yesterday?

johndcollett: this is correct. I don't think I committed the changes when I modified the script in here just in shopify-dev.

PR #4489, opened 2026-05-27
Add POS staff member session signal

akhayoon:
### Background

https://github.com/shop/issues-retail/issues/29701

POS UI extensions currently expose `session.currentSession.staffMemberId` as a static value captured when an extension instance starts. Extensions need a reactive session field so they can respond when the pinned-in staff member changes.

### Solution

Add `session.staffMember` to the POS Session API as a `ReadonlySignalLike<StaffMember | undefined>`, matching existing signal-based APIs like `cart.current` and `connectivity.current`.

The `StaffMember` shape is intentionally limited to `{ id }`. This is a reactive replacement for the existing public `currentSession.staffMemberId` field. I opted not to expose `firstName` / `lastName` in this PR because Shopify's internal PII guidance treats names as PII/personal data, including employee or contractor names:

- https://vault.shopify.io/page/4631

Staff display names can be considered separately with API/privacy review if needed.

This keeps `currentSession.staffMemberId` for backwards compatibility and marks it deprecated. The runtime wiring will happen in the companion `Shopify/extensibility` PR.

Changes:

- Add `StaffMember` with `id`
- Add `SessionApiContent.staffMember`
- Export `StaffMember` from the POS API barrel
- Update POS tester factories to include the new signal
- Add a minor changeset

### 🎩

- `yarn build`

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/session.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/session.ts
@@ -1,6 +1,17 @@
 import type {CurrencyCode} from '../../../shared';
 
 /**
+ * Defines a staff member in POS.
+ * @publicDocs
+ */
+export interface StaffMember {
+  /**
+   * The staff member ID.
+   */
+  id: number;
+}
+
+/**
  * Defines information about the current POS session.
  * @publicDocs
  */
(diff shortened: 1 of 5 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/types/session.ts
```diff
@@ -1,5 +1,26 @@
...
+
+  /**
+   * The staff member's first name, when available.
+   */
+  firstName?: string;
+
+  /**
+   * The staff member's last name, when available.
+   */
+  lastName?: string;
```
vctrchu: Aren't these considered PII? We probably don't want to expose these

vctrchu:
@akhayoon For your knowledge, you can comment `/snapit` on any ui extension PR and it will deploy a snapshot build to insert into your extension. This can be useful it you want to validate the entire package or share it with someone else to test.

vctrchu (approved):

PR #4500, opened 2026-06-05
Expose POS background scanner sources and scan events

vctrchu:
### Background

POS background data extensions need a way to react to scanner input while POS is running, without requiring merchants to open an extension UI first.

The merchant use case is passive scanner awareness from a background target: an associate scans with POS hardware, and the extension can react to that new scan without first opening a modal/action extension UI.

### Design iteration

This PR originally explored exposing a read-only scanner surface on `pos.app.ready.data` with both:

```ts
shopify.scanner.sources.current.value;
shopify.scanner.scannerData.current.subscribe(...);
```

After review discussion, we changed the shape. The feedback was that background data targets should not become a dumping ground for stateful signals when the thing being represented is actually an occurrence. Scan payloads are closer to cart/transaction events than durable state: an extension wants to know that a scan happened now, not read whatever the last scan happened to be.

The final design is therefore:

- Scanner availability remains state: `shopify.scanner.sources`.
- Scan payloads become events: `shopify.addEventListener('scan', ...)`.
- Camera scanner controls remain unavailable on background targets.

### API decision

This PR intentionally splits scanner information into two shapes:

1. Scanner availability is state, so it is exposed as a read-only signal:

```ts
shopify.scanner.sources.current.value;
```

2. Scan payloads are occurrences, so they are delivered as events:

```ts
shopify.addEventListener('scan', event => {
  handleScan(event.data, event.source);
});
```

We should not expose background scan payloads as `shopify.scanner.scannerData.current`. A scan is not durable state like cart contents or scanner availability. Keeping the last scan as a readable signal creates stale "last scanned value" semantics: an extension could read or subscribe after a scan and process an old payload as if it were new. Delivering scan payloads through `shopify.addEventListener('scan', ...)` makes the timing explicit: a scan happened now.

### Solution

- Added `ReadonlyScannerApiContent` / `ReadonlyScannerApi` for background scanner source access.
- Added `ReadonlyScannerApi` to `DataTargetApi`.
- Added `ScanEvent` for `shopify.addEventListener('scan', ...)`.
- Exported scanner source and scan event helper types from the POS API barrel.
- Updated `@shopify/ui-extensions-tester` data target mocks to expose scanner sources without `scannerData` or camera controls.

On `pos.app.ready.data`, developers can use:

```ts
shopify.scanner.sources.current.value;

shopify.addEventListener('scan', event => {
  handleScan(event.data, event.source);
});
```

The background target does not expose:

```ts
shopify.scanner.scannerData;
shopify.scanner.showCameraScanner;
shopify.scanner.hideCameraScanner;
```

Foreground/action scanner targets keep the existing full Scanner API shape for compatibility.

### What this PR does not do

This is the public TypeScript contract and tester support only. Live scan delivery still requires companion runtime work:

- extensibility: expose only scanner sources on `pos.app.ready.data` and route scan payloads through the existing host event listener path.
- POS Mobile: listen for native scanner results while background extensions are enabled and dispatch `scan` events to the background runtime.

### How to test

1. Add the latest snapshot from this PR to an extension `package.json` and install it.
2. Set `api_version = "2026-07"` and target `pos.app.ready.data` in `shopify.extension.toml`.
3. In the background module, read scanner sources with `shopify.scanner.sources.current.value`.
4. Register a scan listener with `shopify.addEventListener('scan', event => handleScan(event.data, event.source))`.
5. Confirm `shopify.scanner.sources` compiles and `shopify.addEventListener('scan', ...)` is typed.
6. Confirm `shopify.scanner.scannerData`, `showCameraScanner`, and `hideCameraScanner` are not available on the background target.

Live runtime scanner delivery requires the companion extensibility runtime and POS Mobile changes.

### Checklist

- [x] I have updated relevant documentation through JSDoc / generated types
- [x] I have added a changeset

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/scanner-api/scanner-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/scanner-api/scanner-api.ts
@@ -42,16 +42,16 @@ export interface ScannerSources {
  */
 export interface ScannerData {
   /**
-   * Current available scanner sources with subscription support. The `value` property provides current sources, and `subscribe` listens for changes. Use to monitor which scanners are available.
+   * Current scanner data with subscription support. The `value` property provides current scan data, and `subscribe` listens for changes. Use to receive scanner events.
    */
   current: ReadonlySignalLike<ScannerSubscriptionResult>;
 }
 
 /**
- * The `ScannerApi` object provides scan results and scanner controls.
+ * The readonly scanner API provides scan results and scanner source information.
  * @publicDocs
  */
-export interface ScannerApiContent {
+export interface ReadonlyScannerApiContent {
   /**
    * Access current scan data and subscribe to new scan events. Use to receive real-time scan results.
    */
@@ -60,6 +60,13 @@ export interface ScannerApiContent {
    * Access available scanner sources on the device. Use to check which scanners are available (camera, external, or embedded).
    */
   sources: ScannerSources;
+}
+
+/**
+ * The `ScannerApi` object provides scan results and scanner controls.
+ * @publicDocs
+ */
+export interface ScannerApiContent extends ReadonlyScannerApiContent {
   /**
    * Show the camera scanner.
    */
@@ -68,6 +75,14 @@ export interface ScannerApiContent {
    * Hide the camera scanner.
    */
   hideCameraScanner: () => void;
+}
+
+/**
+ * The readonly scanner API provides access to scanner data and scanner source information. Access these properties through `shopify.scanner` to monitor scan events and available scanner sources.
+ * @publicDocs
+ */
+export interface ReadonlyScannerApi {
+  scanner: ReadonlyScannerApiContent;
 }
 
 /**
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/scanner-api/scanner-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/scanner-api/scanner-api.ts
@@ -48,14 +48,10 @@ export interface ScannerData {
 }
 
 /**
- * The readonly scanner API provides scan results and scanner source information.
+ * The readonly scanner API provides scanner source information. Scan events are delivered separately through `shopify.addEventListener('scan', ...)`.
  * @publicDocs
  */
 export interface ReadonlyScannerApiContent {
-  /**
-   * Access current scan data and subscribe to new scan events. Use to receive real-time scan results.
-   */
-  scannerData: ScannerData;
   /**
    * Access available scanner sources on the device. Use to check which scanners are available (camera, external, or embedded).
    */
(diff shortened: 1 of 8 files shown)


henryStelle (approved):

packages/ui-extensions/src/surfaces/point-of-sale/api/scanner-api/scanner-api.ts
```diff
@@ -60,6 +60,13 @@ export interface ScannerApiContent {
    * Access available scanner sources on the device. Use to check which scanners are available (camera, external, or embedded).
    */
   sources: ScannerSources;
+}
+
+/**
+ * The `ScannerApi` object provides scan results and scanner controls.
```
fatbattk: nit: `scan data, scanner sources, and [...]`.

fatbattk (approved):

packages/ui-extensions/src/surfaces/point-of-sale/api/scanner-api/scanner-api.ts
vctrchu: Good call, updated to `scan data, scanner sources, and camera controls`. Thanks!

js-goupil (changes requested):
We're exposing signals which are stateful to bodies that are supposed to be run once. 

The solution should be to create a new event that is observable when scan events happen on POS. The risk here is that we're introducing a pattern where the app ready target can be a dumping ground for signals. By this same logic, we could remove the cart change events and have extensions use the cart signals.

Example of what I'm proposing:

```
render('pos.app.ready.data', () => {
   addEventListener('scanevent', (scanData) => {
       // Do something with scan data
   });
   
   addEventListener('scannersourceupdate', () => {
             // Do something with scanner sources
   })
```

henryStelle:
> By this same logic, we could remove the cart change events and have extensions use the cart signals.

That doesn't seem like a bad idea to me. Constantly exposing signals would unlock allowing developers to use some of the fancy `effect` or `computed` signal helpers to simplify more complex logic depending on multiple states changing.

vctrchu:
Going to put this back into draft. We need to set some clear distinctions on when to use signals vs events.

js-goupil:
@henryStelle I'm thinking more about this, and I think that in some cases it makes sense to have signals, purely because we might want to read current values in the reaction to another event. Example: you just want to read the current cart, or read the current available scanners.  

That said, I'm starting to feel like scanner data in particular works better as an event, EVEN in regular UI based targets. It's kind of weird that we would look at the current scan data, for a scan that may have happened 2 minutes ago. Typically, we want to know that a scan happened, and do some work, and then forget about it. Many partners have complained about stale scan data. Similar to how we don't store the last completed transaction, and therefore it belongs as a named event that we subscribe to.

akhayoon:
Updated the PR description to capture the design iteration from the review discussion.

The current shape is:

- `scanner.sources` remains read-only state on `pos.app.ready.data`
- scan payloads are delivered as `scan` events through `shopify.addEventListener('scan', ...)`
- `scannerData`, `showCameraScanner`, and `hideCameraScanner` are not exposed on the background target

I also updated the `ScanEvent` JSDoc example to avoid logging scan payloads directly.

Could you re-review the final event-based API shape? If this matches the direction from the discussion, I’ll take the PR out of draft.

akhayoon (approved):

PR #4508, opened 2026-06-08
Document POS discount allocation API version

janezhu918:
### Background

Part of https://github.com/shop/issues-retail/issues/27165.

POS is moving custom fixed-amount line item discounts from ACROSS allocation to EACH allocation for POS UI Extensions API version 2026-07. Extensions on 2026-04 and earlier keep the existing behavior, where POS accepts the total line-item discount amount and converts it automatically.

### Solution

Update the Cart API JSDoc for `setLineItemDiscount` and `bulkSetLineItemDiscounts` to describe the 2026-07 behavior plainly: `FixedAmount` line item discount amounts are per-unit amounts, including the quantity 2 example.

### Developer changelog

I’ll create a developer changelog entry for the version-delta framing: 2026-04 and earlier accepted the total fixed-amount line item discount and converted it automatically; 2026-07 expects a per-unit `FixedAmount` amount. Example: passing `'5.00'` on a line item with quantity 2 results in a $10.00 total discount.

### Generated docs plan

After this merges into `2026-07-rc`, trigger the shop/world POS UI Extensions API regen workflow with `api_version = 2026-07-rc` and use the bot-generated shop/world companion PR as the generated-docs PR.

### 🎩

- Ran `yarn docs:point-of-sale 2026-07-rc` in this repo.
- Used the generated files to preview the 2026-07 Cart API docs locally in `shopify-dev`.
- Confirmed the `FixedAmount` per-unit description appears inline on the `setLineItemDiscount()` and `bulkSetLineItemDiscounts()` method descriptions.
- Also ran:
  - `yarn build`
  - `yarn type-check`

Screenshot 2026-06-09 at 1 13 05 PM

Screenshot 2026-06-09 at 1 13 02 PM

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/cart-api/cart-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/cart-api/cart-api.ts
@@ -180,7 +180,7 @@ export interface MutableCartApiContent {
   removeLineItemProperties(uuid: string, keys: string[]): Promise<void>;
 
   /**
-   * Apply a discount to a specific line item using its `UUID`. Specify the discount type (`'Percentage'` or `'FixedAmount'`), title, and amount value with improved discount allocation tracking.
+   * Apply a discount to a specific line item using its `UUID`. Specify the discount type (`'Percentage'` or `'FixedAmount'`), title, and amount value with improved discount allocation tracking. `FixedAmount` discounts use per-unit amounts. For example, passing `'5.00'` on a line item with quantity 2 results in a $10.00 total discount.
    *
    * @param uuid the uuid of the line item that should receive a discount
    * @param type the type of discount applied (example: 'Percentage')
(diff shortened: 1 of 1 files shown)


jiricrispeyn (commented):
Should the version-delta framing live in the changelog instead of the docsite?

The docs are already versioned, anyone reading this page has selected 2026-07, so "Starting with API version 2026-07, `FixedAmount` discounts use per-unit amounts" is describing a *change between versions* rather than the behavior of *this* version, which feels like it might fit better as a changelog entry?

What if we split it:

**Developer changelog**: announce the behavior change (2026-04 and earlier = total amount converted automatically → 2026-07 = per-unit), with the worked example. That's the right place to flag the migration impact for `FixedAmount` callers.

**Docsite (this JSDoc)**: just describe current behavior plainly — `FixedAmount` amounts are applied per unit, with the `'5.00'` × qty 2 = $10.00 example. No version conditional.

That keeps each method's description timeless (it won't read awkwardly once 2026-07 isn't the latest) and puts the "heads up, this changed" signal where devs actually look for breaking changes.

One thing I'd double-check before we drop the version mention entirely: does the older (2026-04 and earlier) docsite snapshot get regenerated from this same source string? If the per-unit text would leak into the 2026-04 page, we'd need version gating in the generator rather than just editing the prose.

janezhu918:
Agreed, updated. I removed the version conditional from the method JSDoc so the generated docsite describes the 2026-07 behavior plainly, and added the worked example to both `setLineItemDiscount()` and `bulkSetLineItemDiscounts()`.

I also updated the PR body to track the developer changelog follow-up for the version-delta framing: 2026-04 and earlier accepted the total fixed-amount line item discount and converted it automatically; 2026-07 expects the per-unit `FixedAmount` amount.

I double-checked the regen workflow too: it checks out the requested API version before generating docs, so 2026-04 and earlier snapshots won’t pick up this 2026-07-rc source string unless we explicitly change those branches/versions.

vctrchu (approved):

PR #4510, opened 2026-06-09
fix: mark admin props as optional (remove required) -- fwd port

kyledurand:
Forward-port of #4507

image


sam-b-rose (approved):
Thanks, Kyle!

PR #4512, opened 2026-06-15
Add optional discountAllocations field to pos bundle lineItemComponent

daniszewskik:
## TLDR                                                                                                                            
   - Add optional `discountAllocations` to POS bundle `LineItemComponent`                                                             
   - Export the POS `DiscountAllocation` type for extension authors                                                                   
   - Regenerate `2026-07-rc` POS API docs and add a changeset                                                                         
                                                                                                                                      
## Why                                                                                                                             
 POS UI Extensions expose bundle component tax lines, but not bundle component discount allocations. This blocks compliance partners from reconstructing correct component-level tax splits for discounted bundles with mixed tax rates and tax-inclusive pricing. 
See https://github.com/shop/issues-retail/issues/21806.

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/generated/pos_ui_extensions/2026-07-rc/generated_docs_data_v2.json
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/generated/pos_ui_extensions/2026-07-rc/generated_docs_data_v2.json
@@ -443,6 +443,14 @@
         {
           "filePath": "src/surfaces/point-of-sale/types/cart.ts",
           "syntaxKind": "PropertySignature",
+          "name": "discountAllocations",
+          "value": "DiscountAllocation[]",
+          "description": "An array of discount allocations applied to this component, providing a detailed breakdown of how discounts are distributed across bundle components. Returns `undefined` if no allocations exist.",
+          "isOptional": true
+        },
+        {
+          "filePath": "src/surfaces/point-of-sale/types/cart.ts",
+          "syntaxKind": "PropertySignature",
           "name": "price",
           "value": "number",
           "description": "The price for the custom sale item as currency string. Must be a valid positive amount. Use for non-catalog items and custom pricing.",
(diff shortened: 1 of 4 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
```diff
@@ -199,6 +199,10 @@ export interface LineItemComponent {
    * An array of tax lines applied to this component.
    */
   taxLines: TaxLine[];
+  /**
+   * An array of discount allocations applied to this component, providing a detailed breakdown of how discounts are distributed across bundle components. Returns 'undefined' if no allocations exist. Use for enhanced tax reporting and compliance.
```
jiricrispeyn: Can we not take assumptions of the use case of the case that is exposed? e.g. Remove:
> Use for enhanced tax reporting and compliance.

jiricrispeyn (approved):
👍

packages/ui-extensions/docs/surfaces/point-of-sale/generated/pos_ui_extensions/2026-07-rc/generated_docs_data_v2.json
```diff
@@ -1814,6 +1822,24 @@
       "value": "export interface LocaleApi {\n  locale: LocaleApiContent;\n}"
     }
   },
+  "StaffMember": {
```
alexgraffeocohen: I assume it's OK to be adding unrelated docs when we generate for yours?

packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
alexgraffeocohen: Agreed, we should assume generic usability of any fields we add.

alexgraffeocohen (changes requested):
This looks good, but let's just correct the description.

alexgraffeocohen (approved):
Looks good, thanks! Does this need to be back-ported to any older versions?

js-goupil (approved):

PR #4513, opened 2026-06-15
Add Customer Account EXAF address types

marcelobarreto:
## Summary

Customer Account order address APIs are gaining structured EXAF address data in Customer Account Web. Public extension types and generated reference docs need to expose the same `extendedFields` and `addressCode` fields for extension developers.

## Approach

- Add `AddressExtendedFields` and structured address fields to Customer Account `MailingAddress`
- Export the new address type from the Customer Account API surface
- Document where the structured fields appear on order shipping and billing addresses
- Clarify `extendedFields` and `addressCode` nullability in source comments and generated docs
- Add a changeset for `@shopify/ui-extensions`
- Regenerate Customer Account docs for `2026-07-rc`
- Patch the Customer Account docs generation step so the generated `MailingAddress` docs include Customer Account-only structured fields despite the existing checkout component type-name collision

Related World runtime PR: https://github.com/shop/world/pull/862025
Related World authorization PR: https://github.com/shop/world/pull/825328
Related issue: https://github.com/shop/issues-addresses/issues/937

Co-authored-by: GPT-5.5 <noreply@openai.com>
Orchestrated-by: ae <noreply@shopify.com>


packages/ui-extensions/docs/surfaces/customer-account/build-docs.mjs
```diff
@@ -68,6 +68,100 @@ const cleanupTempFiles = async (tempFiles) => {
...
+   * code. The value is \`null\` when the order address doesn't have an address
+   * code.
+   *
+   * @example '01001-000'
+   */
+  addressCode?: string | null;
+
+`;
+
+const addCustomerAccountMailingAddressDocs = async (generatedDocsV2Path) => {
```
andrewmcgov, line 123: I don't think we should need to make any changes to this script just for this API. It feels like we are solving the problem at the wrong layer here. 

Still investigating the best way to avoid this, but I think we need a different approach here

andrewmcgov (changes requested):

PR #4531, opened 2026-06-26
feat(pos): add options to addLineItem

vctrchu:
Part of https://github.com/shop/issues-retail/issues/28967
[Dev Forum Request](https://community.shopify.dev/t/shopify-pos-ui-extension-shopify-cart-addlineitem-performance-bottleneck/33885/17?u=victor_chu)

## What

Add an optional third argument to the POS `cart.addLineItem` method so a single call can create a line item and decorate it with line-item properties:

```ts
cart.addLineItem(variantId, quantity, options?: {
  properties?: Record<string, string>; // line-item properties
});
```

Introduces the `AddLineItemOptions` type. Backwards compatible — the third argument is optional, so existing two-argument calls are unchanged.

## Why

Cafe-style POS extensions adding a "modified item" (e.g. coffee + modifiers) currently chain two extension API calls:

```ts
const uuid = await cart.addLineItem(variantId, 1);
await cart.addLineItemProperties(uuid, {modifiers: 'oat milk,extra shot'});
```

Each call awaits its own native SDK call and `syncDraftCheckout` round-trip — measured at ~3s p50 for the sequential pattern.

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api/cart-api/cart-api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api/cart-api/cart-api.ts
@@ -1,5 +1,6 @@
 import type {ReadonlySignalLike} from '../../../../shared';
 import type {
+  AddLineItemOptions,
   Address,
   Cart,
   CartUpdateInput,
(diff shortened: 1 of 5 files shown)


packages/ui-extensions/src/surfaces/point-of-sale/api/cart-api/cart-api.ts
```diff
@@ -123,12 +124,19 @@ export interface MutableCartApiContent {
...
    * @param quantity the number of this variant to add to the cart
+   * @param options optional line-item properties to apply to the new line item in the same operation
    * @returns {string} the UUID of the line item added, or the empty string if the user dismissed an oversell guard modal
    * @throws {Error} if POS fails to add the line item
    */
-  addLineItem(variantId: number, quantity: number): Promise<string>;
+  addLineItem(
+    variantId: number,
+    quantity: number,
+    options?: AddLineItemOptions,
```
js-goupil, line 138: Can we place these params in an interface while we have the opporutnity to make a breaking change? 

So then we can call it like this: `addLineItem({variant: 123, quantity: 5})`

js-goupil (approved):
Approving but I feel strongly about using better object style typing here. Not gonna block

packages/ui-extensions/src/surfaces/point-of-sale/api/cart-api/cart-api.ts
henryStelle: I like this - it also adds inline documentation to every call-site so an agent doesn't need to check the docs for `addLineItem` to learn the proper order of arguments.

vctrchu: Sure!

vctrchu: Post-poning this comment since theres plans on a larger refactor of the cart api

https://shopify.slack.com/archives/C07SN91L5L3/p1782998279283549?thread_ts=1782843535.744179&cid=C07SN91L5L3

> I think part of performance will be to re-think the Cart API in a more performant way and we can re-design the mutation functions then.

PR #4568, opened 2026-07-06
fix(pos): include Event Listeners API in targets.json reverse mapping

henryStelle:
## Problem

`BackgroundShopifyGlobal` (which provides `addEventListener`/`removeEventListener` on the `shopify` global) was not being picked up by `build-docs-targets-json.mjs`. The script discovers APIs by parsing type signatures and recursively resolving nested `*Api` names via `getNestedApis()`, but `BackgroundShopifyGlobal` is a global interface — not a type parameter in `DataTargetApi`'s intersection — so it was never discovered.

This caused `EventListenersApi` to be missing from the generated `targets.json` reverse mapping, which broke the `<ExtensionSupports entityName="Event Listeners API" />` component on the shopify-dev docs page.

## Fix

Following the existing `*Api` pattern used by `SessionApi`, `StorageApi`, etc.:

1. Created `EventListenersApi` + `EventListenersApiContent` types in `src/surfaces/point-of-sale/api/event-listeners-api/`
2. Added `EventListenersApi` to `DataTargetApi`'s type intersection in `data-target-api.ts`
3. Exported `EventListenersApi` from `api.ts`
4. Registered the file path in `build-docs-targets-json.mjs`'s `apiFilePaths` map

Now `getNestedApis('DataTargetApi')` naturally discovers `EventListenersApi`, which flows into the reverse mapping as `EventListenersApi -> [pos.app.ready.data]` — no special-case hacks needed.

## Regenerated

- `2026-07-rc/generated_docs_data_v2.json`
- `2026-07-rc/targets.json`


henryStelle:
Superseded by https://github.com/Shopify/ui-extensions/pull/4569

PR #4574, opened 2026-07-14
Add missing children to checkbox type (2025-10)

justinhenricks:
### Background

The documented `label` slot on `s-checkbox` could not be provided as JSX children because the generated Preact intrinsic element type used `BaseElementProps`, which does not define `children`.

Depends on: https://github.com/shop/world/pull/930614

### Solution

Regenerate `Checkbox.d.ts` so `s-checkbox` uses `BaseElementPropsWithChildren<CheckboxElement>`, exposing `preact.ComponentChildren` for slot content.

### 🎩

TypeScript consumers can now provide the documented `label` slot as children without a type error.

### Checklist

- [x] Generated component types from checkout-web

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout/components/Checkbox.d.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/components/Checkbox.d.ts
@@ -28,6 +28,16 @@ export interface BaseElementProps<TClass = HTMLElement> {
     slot?: Lowercase<string>;
 }
 /**
+ * The base properties for elements that have children, extending `BaseElementProps` with children support.
+ * @publicDocs
+ */
+export interface BaseElementPropsWithChildren<TClass = HTMLElement> extends BaseElementProps<TClass> {
+    /**
+     * The child elements to render within this component.
+     */
+    children?: preact.ComponentChildren;
+}
+/**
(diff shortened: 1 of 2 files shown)


packages/ui-extensions/src/surfaces/checkout/components/Checkbox.d.ts
```diff
@@ -27,6 +27,16 @@ export interface BaseElementProps<TClass = HTMLElement> {
      */
     slot?: Lowercase<string>;
 }
+/**
+ * The base properties for elements that have children, extending `BaseElementProps` with children support.
+ * @publicDocs
+ */
+export interface BaseElementPropsWithChildren<TClass = HTMLElement> extends BaseElementProps<TClass> {
```
robin-drexler, line 34: so weird that this isn't a shared type, but I see other components are doing the same?

robin-drexler (approved):

packages/ui-extensions/src/surfaces/checkout/components/Checkbox.d.ts
justinhenricks: yeah i was wondering that too 🤔

PR #4577, opened 2026-07-14
Add missing children to checkbox type (2026-07)

justinhenricks:
### Background

The documented `label` slot on `s-checkbox` could not be provided as JSX children because the generated Preact intrinsic element type used `BaseElementProps`, which does not define `children`.

Depends on: https://github.com/shop/world/pull/930614

### Solution

Regenerate `Checkbox.d.ts` so `s-checkbox` uses `BaseElementPropsWithChildren<CheckboxElement>`, exposing `preact.ComponentChildren` for slot content.

### 🎩

TypeScript consumers can now provide the documented `label` slot as children without a type error.

### Checklist

- [x] Generated component types from checkout-web

### DIFF
--- a/.changeset/checkbox-label-slot-children.md
+++ b/.changeset/checkbox-label-slot-children.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+Fix the `s-checkbox` JSX type so the documented `label` slot can be provided as slotted children. `Checkbox` now uses `BaseElementPropsWithChildren` (matching `Chip` and other slot-bearing components) instead of the childless `BaseElementProps`, which previously left `CheckboxElementSlots` unusable via JSX.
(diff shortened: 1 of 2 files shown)


justinhenricks:
> This PR targets a stable release branch (`2026-07`). Once merged, the change typically also needs to be forward-ported to `2026-07-rc` so it ships in the next release.
> 
> When you open the forward-port PR, include a line like this in its body so the `needs-rc-port` label gets removed automatically when that PR merges:
> 
> ```
> Forward-port of #4577
> ```
> 
> Accepted formats (comma-separated for multiple):
> 
> * `#4577`
> * `GH-4577`
> * `4577`
> * `https://github.com/Shopify/ui-extensions/pull/4577`
> 
> If a forward-port isn't needed (e.g., the change is stable-only), you can remove the `needs-rc-port` label manually.

https://github.com/Shopify/ui-extensions/pull/4578

robin-drexler (approved):

PR #4583, opened 2026-07-15
Hide tertiary button variant from s-button docs

sylvhama:
### Background

The generated `s-button` reference inherited the shared `ButtonProps.variant` description and advertised the `tertiary` variant, even though the public `ButtonElementProps.variant` type is narrowed to `auto`, `primary`, and `secondary`.

This is a docs-only JSDoc description change. The `tertiary` variant remains in the base `ButtonProps` type and in `@shopify/checkout-web-ui`; we just do not want to advertise it yet.

### Solution

Add a local JSDoc description to the narrowed `variant` property without changing its `Extract` type, then regenerate the 2026-07 customer-account and checkout docs data. Include the required patch changeset for `@shopify/ui-extensions`.

### Next steps

- After review, merge this PR into `2026-07`.
- The generated docs will sync automatically to Shopify Dev on the daily cadence; no manual deployment is required. For an urgent sync, contact `#help-shopify-dev-eng`.
- Forward-port this change to `2026-07-rc`, regenerate both docs surfaces for that version, and include `Forward-port of #4583` in the follow-up PR body.

### 🎩

- Ran `yarn docs:customer-account 2026-07`.
- Ran `yarn docs:checkout 2026-07`.
- Confirmed the generated button descriptions no longer include the `tertiary` bullet and the published value remains `'auto' | 'primary' | 'secondary'`.
- Ran `yarn build`, `yarn test`, `yarn type-check`, and `yarn lint`.

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/checkout/components/Button.d.ts
+++ b/packages/ui-extensions/src/surfaces/checkout/components/Button.d.ts
@@ -53,6 +53,15 @@ export interface ButtonElementProps extends Pick<ButtonProps$1, 'accessibilityLa
      * @default 'button'
      */
     type?: Extract<ButtonProps$1['type'], 'submit' | 'button'>;
+    /**
+     * The visual style variant of the button component, which controls its prominence and emphasis.
+     *
+     * - `'auto'`: Automatically determined by the button's context.
+     * - `'primary'`: High-emphasis style for the main action.
+     * - `'secondary'`: Medium-emphasis style for supporting actions.
+     *
+     * @default 'auto'
+     */
     variant?: Extract<ButtonProps$1['variant'], 'auto' | 'primary' | 'secondary'>;
 }
 export interface ButtonEvents extends Pick<ButtonProps$1, 'onClick'> {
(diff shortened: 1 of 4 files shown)


packages/ui-extensions/src/surfaces/checkout/components/Button.d.ts
```diff
@@ -53,6 +53,15 @@ export interface ButtonElementProps extends Pick<ButtonProps$1, 'accessibilityLa
...
     type?: Extract<ButtonProps$1['type'], 'submit' | 'button'>;
+    /**
+     * The visual style variant of the button component, which controls its prominence and emphasis.
+     *
+     * - `'auto'`: Automatically determined by the button's context.
+     * - `'primary'`: High-emphasis style for the main action.
+     * - `'secondary'`: Medium-emphasis style for supporting actions.
+     *
+     * @default 'auto'
+     */
```
sylvhama, line 64: Same as https://shopify.dev/docs/api/customer-account-ui-extensions/latest/web-components/actions/button#buttonelementprops-propertydetail-variant without `tertiary`

adrian-augustyn (approved):

RenataZakauske (approved):

robin-drexler (approved):

oliverigor (approved):
This seems fine, but we need to check the checkout-web side so this don't get overwritten in a next generation of the types

robin-drexler (approved):

PR #4601, opened 2026-07-28
Remove shopUser from documented AppMetafieldEntryTarget types

amertkara:
Resolves https://github.com/shop/issues-checkout/issues/145

## Summary

- Removes `'shopUser'` from the `AppMetafieldEntryTarget.type` union in the checkout surface `standard.ts`, and drops its JSDoc bullet.
- Shop User metafields were never released publicly and there is no intention to release them (privacy concerns). Access has already been revoked, and the merchant-facing handling code was previously removed from checkout-web in shop/world.
- Because the generated API reference renders the type union itself, the value is currently published on shopify.dev (e.g. `/docs/api/checkout-ui-extensions/2025-07/apis/metafields`), which led partners to attempt using an unsupported capability and caused repeat incidents (see issue).
- The static configuration page (`configuration#metafields`) was already cleaned up separately; this PR covers the remaining generated API reference.
- Runtime behavior is unchanged — this only narrows the published/documented type.

## Tophat

1. Confirm no remaining `shopUser` references in the checkout surface: `grep -rn \"shopUser\" packages/ui-extensions/src` → no results.
2. Regenerate the checkout API reference docs and verify `shopUser` no longer appears in the `AppMetafieldEntryTarget` type values.

### DIFF
--- a/.changeset/remove-shopuser-metafield-docs.md
+++ b/.changeset/remove-shopuser-metafield-docs.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
+Remove `shopUser` from the documented `AppMetafieldEntryTarget` resource types for checkout UI extensions. Shop User metafields were never released publicly and are not a supported metafield target for partner extensions; keeping the value in the public API reference caused partners to attempt using an unsupported capability.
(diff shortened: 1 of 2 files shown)


amertkara:
Closing in favor of fixing this at the source: the checkout-web copy in shop/world is the source of truth for these docs (per https://github.com/shop/issues-checkout/issues/145). The change will land in this repo through the usual ui-extensions docs sync.

PR #4613, opened 2026-08-12
Document borderRadius for POS s-image

ajanth-u:
## Problem

The POS `s-image` component supports `borderRadius` in API version `2026-07`, but its public type and generated reference omit the property. Extension code that uses the supported property fails type checking, and the reference doesn't explain its flow-relative shorthand.

## Solution

Expose and document `borderRadius` for `s-image` in `2026-07`, including its default, accepted design-system values, and one-to-four-value flow-relative order.

### DIFF
--- a/.changeset/quiet-images-curve.md
+++ b/.changeset/quiet-images-curve.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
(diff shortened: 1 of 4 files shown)


ajanth-u:
Closing at the author’s request.

PR #4614, opened 2026-08-12
Document borderRadius for POS s-image in 2026-10

ajanth-u:
## Problem

The POS `s-image` component supports `borderRadius` in API version `2026-07` and later, but the public type omits the property. Extension code that uses the supported property fails type checking.

## Solution

Forward-port the `2026-07` type and documentation fix into `2026-10-rc` so the next API version keeps the property.

Forward-port of #4613

Related Shopify Dev Docs PR: https://github.com/shop/world/pull/992375

### DIFF
--- a/.changeset/quiet-images-curve.md
+++ b/.changeset/quiet-images-curve.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': patch
+---
+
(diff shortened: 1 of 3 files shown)


ajanth-u:
Closing at the author’s request.

PR #4620, opened 2026-08-14
docs(pos): correct Discount.amount read behavior for 2026-07

vctrchu:
## What

Corrects the `Discount.amount` TSDoc on the POS cart types for the released 2026-07 API version: reading a discount from the cart returns the calculated monetary total in this field, not the value that was applied. The previous doc claimed reads return the percentage value, which is not what this version does (shop/issues-retail#33093, reported in https://community.shopify.dev/t/36265 and https://community.shopify.dev/t/36171).

Also warns that echoing a read discount back through `bulkCartUpdate` does not preserve it in this version, and points to 2026-10 where reads match writes (Shopify/extensibility#1657, shop/world#997169).

Older released branches have the same behavior; happy to cherry-pick this doc note to them if that's the convention.

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/types/cart.ts
@@ -246,7 +246,9 @@ export interface SellingPlan {
  */
 export interface Discount {
   /**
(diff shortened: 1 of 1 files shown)


vctrchu:
Closing — no longer needed. We're fixing the runtime to match these docs instead of documenting the bug: POS will return the input value ("10" for 10% off) as `Discount.amount` on all API versions. See shop/world#997169 and shop/issues-retail#33093.

PR #4623, opened 2026-08-14
Document borderRadius for POS s-image

ajanth-u:
## Problem

The POS `s-image` component supports `borderRadius` at runtime in API version `2026-07`, but its public TypeScript contract and generated property reference omit the property. Valid extension code therefore fails type checking, and partners can't find the property in the reference.

## Solution

Align the `2026-07` public contract with the existing runtime. Document the property's default, supported design-system values, and one-to-four-value flow-relative shorthand.


prakashraman:
Changes look right. Let me locally checkout the branch, and spin up the docs.

prakashraman:
image

Works, thanks.

prakashraman (approved):
LGTM. Could you followup and find out how to get this deployed?

fatbattk (approved):

PR #4625, opened 2026-08-14
Add auto to POS button variant type

henryStelle:
### Background

Fixes https://github.com/shop/issues-retail/issues/33323.

The POS `s-button` documentation identifies `auto` as the default variant, but the public button variant type only allowed `primary` and `secondary`.

### Solution

Add `auto` to the POS button variant type and generated documentation, and document it alongside the other supported variants.

### Validation

- `yarn build`
- `yarn type-check`
- `yarn lint`
- Targeted TypeScript compile check for `variant: 'auto'`

The full `yarn test -- packages/ui-extensions/` command is currently blocked by the missing `react-test-renderer` dependency in unrelated Checkout suites.

image

### 🎩

- Confirm a POS extension can type-check `<s-button variant="auto">`.

### Checklist

Forward-port of #4627, #4629, #4628, #4626


fatbattk (approved):
code lgtm but does `auto` actually work?
> determines the button variant based on its context

PR #4634, opened 2026-08-17
Add types for POS intercept ValidationAdd target values

vctrchu:
### Background

https://github.com/shop/issues-retail/issues/33533

The POS intercept API (#4520) shipped `ValidationAdd.target` as a bare `string` documented as a "JSON-path locator". The target contract we're keeping is the [Functions validation target model](https://shopify.dev/docs/api/functions/latest/cart-and-checkout-validation#supported-checkout-field-targets): an enumerated token table, string-matched, never evaluated as JSON paths. Bare `string` can never be narrowed later without a breaking change, and it silently accepts targets no host honors.

### Solution

Pin the grammar as template literal types while the intercept API is still `@private`, narrowed per intercepted event:

| Event | Field | Target value |
| --- | --- | --- |
| `cartvalidations` | cart | `$.cart` |
| `cartvalidations` | line item | `$.cart.lineItems['<uuid>']` |
| `paymentvalidations` | payment attempt | `$.payment` |

- `ValidationTarget` union of `CartValidationTarget` / `PaymentValidationTarget`; wire format stays a plain string, and unions can widen later (never narrow).
- `ValidationTargetFor<TEvent>` narrows interceptor returns per event: a `cartvalidations` interceptor cannot return `$.payment` (compile error), and vice versa. `InterceptResult` / `Operation` / `ValidationAdd` gain a defaulted type parameter, so existing structural consumers compile unchanged.
- Devs write the line target inline — TS contextually checks template expressions against template literal types:

  ```ts
  target: `$.cart.lineItems['${lineItem.uuid}']`  // checked char-for-char
  target: `$.cart.lineItems[${idx}]`              // compile error
  ```

- JSDoc pins the semantics: targets are data locators (host decides rendering, falls back to the event's root scope); line item uuids must be echoed from the same event's cart snapshot, not cached across events; bundle components target their parent line; unrecognized targets degrade to root scope rather than dropping the validation; handles are namespaced per extension and may repeat across targets.

### Alternatives Considered

1. **Keep `target?: string`** — ❌ can never be narrowed without a breaking change; typos and unsupported targets compile silently.
2. **JSONPath filter `$.cart.lineItems[?(@.uuid=='<uuid>')]`** — ❌ buys spec-valid JSONPath semantics the Functions token model doesn't have (its own table hardcodes `deliveryGroups[0]` against multi-group carts), at the cost of syntax hostile to extension developers.
3. **Positional `$.cart.lineItems[2]`** — ❌ cannot carry stable identity; `.filter().map((li, i))` silently targets the wrong line.
4. **Structured object / sibling uuid field** — ❌ breaks the single-string target contract shared with Functions.

### :tophat:

- Type-level only (`@private` API, no runtime change beyond the type declarations). `yarn type-check` passes.
- Verified narrowing with a scratch file: valid tokens assignable per event; cross-event tokens and positional indexes fail compilation with `@ts-expect-error` consumed.
- Host-side runtime sanitation of these tokens is tracked in https://github.com/shop/issues-retail/issues/33534 (extensibility repo); POS line-item rendering in https://github.com/shop/issues-retail/issues/32377.

### Checklist

### DIFF
--- a/packages/ui-extensions/src/surfaces/point-of-sale/events.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/events.ts
@@ -110,10 +110,60 @@ export interface PaymentValidationsEvent extends Event {
   readonly amount: MoneyV2;
 }
 
+/**
+ * Targets the whole cart. Validations with this target render at the cart
+ * scope (for example the cart banner).
+ *
+ * @private
+ */
+export type CartTarget = '$.cart';
+
+/**
+ * Targets one cart line item by its `uuid` from this event's `cart` snapshot,
+ * for example `$.cart.lineItems['adfd6b06-4a24-4f5f-9f4b-ea21f4432dd4']`.
+ *
+ * @private
+ */
+export type CartLineItemTarget = `$.cart.lineItems['${string}']`;
+
+/** @private */
+export type CartValidationTarget = CartTarget | CartLineItemTarget;
+
+/**
+ * Targets the payment attempt being intercepted.
+ *
+ * @private
+ */
+export type PaymentTarget = '$.payment';
+
+/** @private */
+export type PaymentValidationTarget = PaymentTarget;
+
+/**
+ * Where a validation applies, as an enumerated token following the
+ * [Functions validation target model](https://shopify.dev/docs/api/functions/latest/cart-and-checkout-validation#supported-checkout-field-targets).
+ * Targets are matched as exact strings, never evaluated as JSON paths.
+ *
+ * @private
+ */
+export type ValidationTarget = CartValidationTarget | PaymentValidationTarget;
+
+/**
+ * The validation targets valid for a given intercepted event.
+ *
+ * @private
+ */
+export type ValidationTargetFor<TEvent extends Event> =
+  TEvent extends CartValidationsEvent
+    ? CartValidationTarget
+    : TEvent extends PaymentValidationsEvent
+    ? PaymentValidationTarget
+    : ValidationTarget;
+
 /** @private */
 export type ShopifyInterceptor<TEvent extends Event> = (
   event: TEvent,
-) => InterceptResult;
+) => InterceptResult<ValidationTargetFor<TEvent>>;
 
 /**
  * The result an interceptor returns. An empty `operations` list allows the
@@ -121,8 +171,10 @@ export type ShopifyInterceptor<TEvent extends Event> = (
  *
  * @private
  */
-export interface InterceptResult {
-  operations: Operation[];
+export interface InterceptResult<
+  TTarget extends ValidationTarget = ValidationTarget,
+> {
+  operations: Operation<TTarget>[];
 }
 
 /**
@@ -130,8 +182,10 @@ export interface InterceptResult {
  *
  * @private
  */
-export interface Operation {
-  validationAdd?: ValidationAdd;
+export interface Operation<
+  TTarget extends ValidationTarget = ValidationTarget,
+> {
+  validationAdd?: ValidationAdd<TTarget>;
 }
 
 /** @private */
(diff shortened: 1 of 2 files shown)


vctrchu:
CI notes: lint failure was prettier indentation in my change — fixed and pushed. The `test` job failure is pre-existing on `2026-10-rc` (checkout testing example: `api_version "2026-10" does not match @shopify/ui-extensions-tester ("2027-00")`) — same failure on the branch tip before this PR ([#4625's CI run](https://github.com/Shopify/ui-extensions/actions?query=branch%3A2026-10-rc)). Not addressed here since this PR is point-of-sale types only.

packages/ui-extensions/src/surfaces/point-of-sale/events.ts
```diff
@@ -110,28 +110,82 @@ export interface PaymentValidationsEvent extends Event {
   readonly amount: MoneyV2;
 }
 
+/**
+ * Targets the whole cart. Validations with this target render at the cart
+ * scope (for example the cart banner).
```
henryStelle: I'm not sure we want to call out "cart banner" as they are all accessed through that banner.

packages/ui-extensions/src/surfaces/point-of-sale/events.ts
```diff
@@ -110,28 +110,82 @@ export interface PaymentValidationsEvent extends Event {
...
+/**
+ * Targets one cart line item by its `uuid` from this event's `cart` snapshot,
+ * for example `$.cart.lineItems['adfd6b06-4a24-4f5f-9f4b-ea21f4432dd4']`.
+ *
+ * @private
+ */
+export type CartLineItemTarget = `$.cart.lineItems['${string}']`;
+
+/** @private */
+export type CartValidationTarget = CartTarget | CartLineItemTarget;
```
henryStelle, line 132: No customer?

packages/ui-extensions/src/surfaces/point-of-sale/events.ts
```diff
@@ -110,28 +110,82 @@ export interface PaymentValidationsEvent extends Event {
...
+ * The validation targets valid for a given intercepted event.
+ *
+ * @private
+ */
+export type ValidationTargetFor<TEvent extends Event> =
+  TEvent extends CartValidationsEvent
+    ? CartValidationTarget
+    : TEvent extends PaymentValidationsEvent
+    ? PaymentValidationTarget
+    : ValidationTarget;
```
henryStelle: nit: this isn't very readable, could you use a map?

packages/ui-extensions/src/surfaces/point-of-sale/events.ts
vctrchu: Customer was descoped from this project. Let me find the decision

vctrchu: https://shopify.slack.com/archives/C08S31CR0V8/p1785157214992659

henryStelle (approved):

packages/ui-extensions/src/surfaces/point-of-sale/events.ts
```diff
@@ -142,15 +204,31 @@ export type ValidationLevel = 'WARNING' | 'ERROR';
...
+   * renders, falling back to the event's root scope. Defaults to the root
+   * scope (`$.cart` / `$.payment`).
+   *
+   * Line item uuids are only valid within the event that delivered them:
+   * echo `lineItems[n].uuid` from this event's cart snapshot, don't cache
+   * uuids across events. Bundle components are not addressable; target their
+   * parent line. Unrecognized targets degrade to the root scope — the
+   * validation still applies, rendered less specifically.
+   */
+  target?: TTarget;
```
aaronschubert0, line 232: @vctrchu `echo` feels out of context here, could your agent take another pass at this code comment please?

aaronschubert0 (approved):

packages/ui-extensions/src/surfaces/point-of-sale/events.ts
vctrchu: Reworded — swapped "echo" for "use" in 085950105.

PR #4637, opened 2026-08-18
docs: add calver changeset rules to AGENTS.md

vctrchu:
Docs-only follow-up to #4636.

Adds a **Changesets and versioning** section to `AGENTS.md`: packages here are calver-keyed (`major.minor` of the npm version *is* the Shopify API version), so breaking changes must ship as `minor` changesets — a `major` rolls the calendar year, produces a nonexistent API version, and breaks the tester's version check branch-wide. Also notes that pre-mode retains released changeset files, so version corrections must fix the changeset itself, not just `package.json`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,6 +7,23 @@ After every change, you must:
 3. **Check for type errors** — run the TypeScript compiler to verify there are no type errors.
 4. **Check for lint and formatting errors** — run the linter and formatter to ensure code quality.
 
+## Changesets and versioning
+
+Package versions are **calendar versions**: the `major.minor` of the npm
+version *is* the Shopify API version (`2026.10.x` → API version `2026-10`,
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -7,6 +7,23 @@ After every change, you must:
...
 
+## Changesets and versioning
+
+Package versions are **calendar versions**: the `major.minor` of the npm
+version _is_ the Shopify API version (`2026.10.x` → API version `2026-10`,
+derived at build time — see `packages/ui-extensions-tester/src/api-version.ts`).
+
+- **Never mark a changeset `major`.** A major bump rolls the calendar year
+  (`2026.10.0` → `2027.0.0` → API version "2027-00", which doesn't exist),
+  breaks the tester's version check on every PR, and publishes an RC that is
```
fatbattk, line 18: wondering if better to simplify and skip details other than it breaks things.

fatbattk (approved):

PR #4639, opened 2026-08-19
Forward-port POS s-image borderRadius to 2026-10

ajanth-u:
## Problem

The POS `s-image` component supports `borderRadius` at runtime from API version `2026-07`, but the upcoming `2026-10` public TypeScript contract omits the property.

## Solution

Forward-port the public Image type and documentation source so `borderRadius` remains supported in `2026-10`. The full `2026-10-rc` generated POS reference baseline will land separately.

Forward-port of #4638


prakashraman (approved):
Image

Tested locally. LGTM.

PR #4641, opened 2026-08-20
feat(point-of-sale): add resolution target types for cart and payment validations

vctrchu:
Resolves https://github.com/shop/issues-retail/issues/33801

### Background

POS intercept extensions (`shopify.intercept('cartvalidations' | 'paymentvalidations', ...)`) block a workflow by returning an `ERROR`-level `validationAdd`. When that happens, POS opens an app-owned remediation UI inside its resolution flow. Each interceptable event gets its own dedicated resolution render target so the public types can express per-event API availability.

This adds the TypeScript contracts for those targets. Everything lands `@private`, matching the rest of the intercept type family, until early access.

### Solution

Two new render targets in `RenderExtensionTargets`:

- `pos.cart.validations.resolution.render` — Standard APIs + Scanner + **mutating** Cart + `ResolutionApi<CartValidationsEvent>`
- `pos.payment.validations.resolution.render` — Standard APIs + Scanner + **read-only** Cart + `ResolutionApi<PaymentValidationsEvent>`

The cart is frozen during payment, so the payment target reuses `ReadonlyCartApi`; the cart target mutates via the existing `CartApi`. Neither target exposes the Action API (resolution is host-invoked, not merchant-invoked) or domain-scoped APIs.

New supporting types:

- `ResolutionApi<TEventData>` — target-scoped `shopify.resolution.event` signal carrying the intercepted event's data (`CartValidationsEventData`, `PaymentValidationsEventData` — the same fields the interceptor sees, without the DOM `Event` base), typed per event rather than as one broad union.
- `ResolutionComponents` — all standard components except `Page`, `POSBlock`/`PosBlock`, and `Tile`, since resolution UI renders inside a POS-owned flow step.

The tester's `PosMockFactory` map is exhaustively keyed over all POS targets, so mock factories for both new targets are included, with tests covering the mutating/read-only cart split and the resolution event shape.

### 🎩

image

image

- `yarn build`, `yarn test` (64 tests incl. 6 new in `pos-resolution-targets.test.ts`), eslint + prettier on touched files — all green

### Checklist


packages/ui-extensions/src/surfaces/point-of-sale/extension-targets.ts
```diff
@@ -370,6 +375,53 @@ export interface RenderExtensionTargets {
...
+   * uses its one step to resolve all of its findings for the event.
+   *
+   * Extensions at this target have read-only cart access — the cart is frozen
+   * during payment — and read the intercepted event through
+   * `shopify.resolution`. Staged outcomes shared with the background
+   * interceptor go through the Storage API.
+   *
+   * The resolution surface owns its footer's primary action, which advances
+   * the flow. `<s-button variant="primary">` is coerced to `secondary` on
+   * this target.
```
henryStelle: nit: I'm not a fan of how we're duplicating this message. It almost feels like this is the kind of thing we would want to say once in a shared _how do you write intercept and resolution tutorial_

henryStelle (approved):

packages/ui-extensions/src/surfaces/point-of-sale/extension-targets.ts
vctrchu: Ill remove it

vctrchu: Agreed — the full intercept → resolution walkthrough belongs in a shared tutorial, and that's planned for the docs follow-up PR alongside the reference examples. Keeping the sentence per target though: these TSDoc blocks compile into standalone per-target reference pages, so each page has to carry its own contract (same reason the existing `*.action.render` targets repeat their prose). The tutorial gets the long-form version; this stays the one-line contract note.

akhayoon (approved):

aaronschubert0 (approved):

PR #4659, opened 2026-08-28
Pin GitHub Actions to Node 24 v7 releases

Fionoble:
## What changed

- Upgrade every `actions/checkout` usage from v3/v4 to v7.0.0, pinned to immutable commit `9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0`.
- Upgrade `actions/setup-node` from v4 to v7.0.0, pinned to immutable commit `820762786026740c76f36085b0efc47a31fe5020`.

Both selected releases run natively on Node 24.

## Why

GitHub now runs JavaScript actions on Node 24 by default as part of the [Node 20 deprecation](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/). The `2026-10-rc` release workflow still uses `actions/checkout@v3`; after the runner migration, checkout configures credentials but its authenticated fetch fails.

This blocked `@shopify/ui-extensions@2026.10.0-rc.8` from publishing even after the Version Packages PR merged. The failed workflow is https://github.com/Shopify/ui-extensions/actions/runs/33114738095.

## Validation

- `yarn lint`
- Prettier check across `.github/workflows/**/*.{yml,yaml}`
- Verified the pinned v7.0.0 `action.yml` files declare `runs.using: node24`

### DIFF
--- a/.github/workflows/actions/prepare/action.yml
+++ b/.github/workflows/actions/prepare/action.yml
@@ -4,6 +4,8 @@ runs:
   steps:
     - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
       name: Setup node.js and yarn
+      env:
+        NODE_AUTH_TOKEN: ''
       with:
         registry-url: 'https://registry.npmjs.org' # Required for OIDC
         cache: yarn
--- a/.github/workflows/actions/prepare/action.yml
+++ b/.github/workflows/actions/prepare/action.yml
@@ -2,7 +2,7 @@ name: Prepare repo
 runs:
   using: 'composite'
   steps:
-    - uses: actions/setup-node@v5
+    - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
       name: Setup node.js and yarn
       with:
         registry-url: 'https://registry.npmjs.org' # Required for OIDC
--- a/.github/workflows/actions/prepare/action.yml
+++ b/.github/workflows/actions/prepare/action.yml
@@ -2,7 +2,7 @@ name: Prepare repo
 runs:
   using: 'composite'
   steps:
-    - uses: actions/setup-node@v4
+    - uses: actions/setup-node@v5
       name: Setup node.js and yarn
       with:
         registry-url: 'https://registry.npmjs.org' # Required for OIDC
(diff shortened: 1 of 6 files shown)


.github/workflows/changesets-reminder.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     timeout-minutes: 2
     if: ${{ !github.event.pull_request.draft && !startsWith(github.head_ref, 'changeset-release/') }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v5
```
andrewiggins: Any reason not to go to the latest version, v7?

.github/workflows/actions/prepare/action.yml
```diff
@@ -2,7 +2,7 @@ name: Prepare repo
 runs:
   using: 'composite'
   steps:
-    - uses: actions/setup-node@v4
+    - uses: actions/setup-node@v5
```
andrewiggins: Any reason not to go to the latest version, v7?

andrewiggins (commented):
While you are here, do you want to update these to use sha pinned versions, for example
```
uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```

.github/workflows/changesets-reminder.yml
Fionoble: Yea, no reason to stop at v5. Updated every checkout reference to v7.0.0 and pinned it to the immutable commit SHA.

.github/workflows/actions/prepare/action.yml
Fionoble: Updated this to setup-node v7.0.0 as well, pinned to its immutable commit SHA.

Fionoble:
Also addressed the review-level suggestion: both actions are now pinned to immutable v7.0.0 commit SHAs across every workflow.

Fionoble:
v7 follow-up is green now. `checkout@v7` worked directly; `setup-node@v7` needed an explicit empty `NODE_AUTH_TOKEN` because v7 removed its dummy export and Yarn 1 reads the generated registry config during cache lookup.

andrewiggins (approved):
Thanks!

PR #4672, opened 2026-09-02
Add cart.updateLineItemQuantity to the POS Cart API (2026-10)

js-goupil:
### Background

https://github.com/shop/issues-retail/issues/34342
Community feature request: https://community.shopify.dev/t/feature-request-updatelineitemquantity-in-the-pos-ui-extensions-cart-ap/36489

POS UI extensions have no line-scoped way to change the quantity of an existing cart line. Today an extension must either re-add the variant (which creates a separate line, or merges unpredictably across POS versions) or rewrite the whole cart with `bulkCartUpdate` (which risks rejecting on computed fields and has repriced existing lines to $0.00). Custom sale lines have no variant ID, so `bulkCartUpdate` is their only option.

### Solution

Adds `updateLineItemQuantity(uuid: string, quantity: number): Promise<void>` to the POS Cart API (`MutableCartApiContent`), available in API version 2026-10 and later.

- Sets the absolute quantity of one existing line item, identified by UUID. POS resolves the full line item internally, so the line's properties, discounts, and selling plans are preserved — the same behavior as a merchant tapping +/- on the native cart line.
- `quantity` must be an integer of 1 or greater; use `removeLineItem` to remove a line.
- The promise resolves after the cart state reflects the change.
- Throws if the line item is not found, the quantity is invalid, the cart is not editable (during a return or exchange), or the POS app version does not support the method.
- If a UUID matches multiple split-line allocations, the call throws (ambiguous target) rather than guessing a line.

Also included: a docs example, tester mock support (`@shopify/ui-extensions-tester`), and a changeset (minor for `@shopify/ui-extensions`, patch for the tester).

### :tophat:

- Host-side implementation: https://github.com/Shopify/extensibility/pull/1713 (cart plugin); POS runtime: shop/world Meteorite PR 2027388 (internal). Tophat end-to-end there. Type surface here validated with `yarn build`, `yarn type-check`, `yarn lint`, and the POS tester test suites.

### Checklist

### DIFF
--- a/packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/cart-api/update-line-item-quantity.jsx
+++ b/packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/cart-api/update-line-item-quantity.jsx
@@ -1,0 +1,23 @@
+import {render} from 'preact';
+
+export default async () => {
+  render(<Extension />, document.body);
+};
+
+const Extension = () => {
+  return (
+    <s-tile
+      heading="My App"
+      subheading="Call cart function"
+      onClick={() => {
+        const [lineItem] = shopify.cart.current.value.lineItems;
+        if (lineItem) {
+          shopify.cart.updateLineItemQuantity(
+            lineItem.uuid,
+            lineItem.quantity + 1,
+          );
+        }
+      }}
+    />
+  );
+};
--- a/.changeset/pos-cart-update-line-item-quantity.md
+++ b/.changeset/pos-cart-update-line-item-quantity.md
@@ -1,0 +1,6 @@
+---
+'@shopify/ui-extensions': minor
+'@shopify/ui-extensions-tester': patch
+---
+
+Add `cart.updateLineItemQuantity` to the POS Cart API for API version 2026-10. The method updates one line item's quantity while preserving its properties, discounts, and selling plans.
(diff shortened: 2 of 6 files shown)


packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/cart-api/update-line-item-quantity.jsx
```diff
@@ -0,0 +1,17 @@
...
+  render(<Extension />, document.body);
+};
+
+const Extension = () => {
+  return (
+    <s-tile
+      heading="My App"
+      subheading="Call cart function"
+      onClick={() => {
+        shopify.cart.updateLineItemQuantity('aa-1234567', 2);
```
fatbattk: nit: use valid uuid format?

fatbattk (approved):

packages/ui-extensions/docs/surfaces/point-of-sale/reference/examples/cart-api/update-line-item-quantity.jsx
js-goupil: Good call — updated the example to read the uuid (and current quantity) from `shopify.cart.current` instead of a placeholder, which also shows where line item UUIDs come from: cc572f2ff

PR #4680, opened 2026-09-05
Remove unused POS reference examples

henryStelle:
### Background

The POS files under `docs/surfaces/point-of-sale/reference/examples/` belonged to the legacy reference generator. The current POS documentation build reads only `src/surfaces/point-of-sale`, while published examples are maintained as World-owned MDX and example snapshots in shopify-dev.

The legacy directory is no longer referenced or validated, and some files advertise components that POS does not support.

### Solution

Remove the complete unused POS reference example tree: 174 files and 4,782 lines.

I regenerated the 2026-10 POS documentation before and after the deletion. `generated_docs_data_v2.json` and `targets.json` were identical, confirming that these files are not inputs to the current shopify-dev data pipeline.

No changeset is included because this removes unconsumed documentation source and does not change the published package or API.

### :tophat:

Not applicable; there is no package or runtime behavior change.

Validation:

- `yarn docs:point-of-sale 2026-10` before and after deletion — identical generated output
- `yarn build`
- `yarn test`
- `yarn type-check`
- `yarn lint`

### Checklist

- [x] Tophatting is not applicable to this cleanup
- [x] Current documentation generation was verified


fatbattk (approved):
lgtm

PR #4681, opened 2026-09-09
POS LineItem productCategory

chadcromwell:
### Background

Extensions that render receipts or read the cart often need a line item's product category, for example to apply category-specific treatment or for categorization and reporting. Today the `LineItem` payload does not carry it, so an extension has to resolve the category with a separate API call.

https://github.com/shop/issues-retail/issues/34611

### Solution

Adds an optional `productCategory` field to the point of sale `LineItem` interface, plus a new `ProductCategory` type exported from the point of sale surface:

- `id`: the stable, locale-independent Standard Product Taxonomy GID (for example `gid://shopify/TaxonomyCategory/aa-1-2-3`), safe as a mapping key.
- `name`: the category's own name (for example `Shirts & Tops`).
- `fullName`: the full ancestor path (for example `Apparel & Accessories > Clothing > Shirts & Tops`).

All three are sent so consumers never have to resolve the name or path from the id, which is the whole point of putting it on the payload rather than having each extension look it up.

The field is optional and populated per line item by the host. It is omitted for line items with no underlying product (custom or quick sale items, gift cards) and for products with no taxonomy category assigned, so the change is additive and existing extensions are unaffected. No other surface is changed.

Alternative considered: leave extensions to fetch the category themselves. Rejected because it adds a per-receipt lookup and latency for data the host already has.

### 🎩

Type-only change on this repo, so there is nothing to exercise at runtime here; the field is populated by the POS host in a separate change and will be tophatted end to end there.

- `yarn type-check` passes.
- `eslint` passes on the changed files.
- `productCategory` is present and optional on `LineItem`, and
  `ProductCategory` is importable from the point of sale surface, in the
  built types.

### Checklist

### DIFF
--- a/.changeset/pos-line-item-product-category.md
+++ b/.changeset/pos-line-item-product-category.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions': minor
+---
+
+Adds an optional `productCategory` field to the point of sale `LineItem` interface, exposing the product's Shopify Standard Product Taxonomy category (`id`, `name`, `fullName`) so receipt and cart extensions can read it without a separate lookup.
--- a/packages/ui-extensions/src/surfaces/point-of-sale/api.ts
+++ b/packages/ui-extensions/src/surfaces/point-of-sale/api.ts
@@ -109,6 +109,7 @@ export type {
   Customer,
   LineItem,
   LineItemComponent,
+  ProductCategory,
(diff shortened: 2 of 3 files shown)


fatbattk (approved):

chadcromwell:
This has been desecoped, so closing for now.

PR #4685, opened 2026-09-11
Remove deprecated POS session staffMemberId in 2026-10

IHu04:
## Summary

- Remove the deprecated `session.currentSession.staffMemberId` property from the POS `Session` type for 2026-10. Use the reactive `session.staffMember` signal instead.
- Update the POS tester factory for Session API render and data targets.
- Preserve the separate receipt event `session.staffMemberId` type and mock value. `BaseData` shared the `Session` type, so it now explicitly retains the optional snapshot field.
- Add regression coverage for tile, modal, and data targets (field absent), plus receipt header and footer targets (field preserved). The negative type assertions reject access to the removed Session API field.
- Correct stale Session API descriptions, update affected generated 2026-10 API data and tester READMEs, and add a minor changeset documenting the versioned breaking change.

## Testing

- `yarn build` — passed
- `yarn test -- packages/ui-extensions-tester/` — 12 suites, 110 tests passed
- `yarn type-check` — passed
- `npx tsc --project packages/ui-extensions-tester/tsconfig.typecheck.json` — passed, including test files and negative type assertions
- `npm run typecheck && npm test` in all four `examples/testing/*` projects — passed (15 tests)
- Documentation generation and assertions that the generated Session API omits the field while receipt data retains it — passed
- `yarn lint` — formatting passed; ESLint reports 8 errors only in unrelated build-generated `packages/ui-extensions/checkout/preact.js` and `packages/ui-extensions/customer-account/preact.js`. ESLint passed for changed TypeScript files and for the repository excluding those two generated entry points.
- Receipt regression tests were confirmed to fail (both runtime and TypeScript) before the preservation fix.

## Scope / follow-up

- Only targets `2026-10-rc`; 2026-07 and earlier published types are unchanged.
- The ticket also mentions `unstable`, but `documentation/versions-and-deploys.md` says that branch is deprecated, frozen at 2025-07, and must not receive new work.
- Host-side API-version gating and the `instanceReloaded` update still require a separate Shopify/extensibility change. This PR does not change the real POS runtime.
- Online shopify.dev publication, the developer changelog, and the external internal/E2E extension audit remain separate follow-ups.

## Related

- shop/issues-retail#34452 (partial implementation; do not close the entire ticket with this PR)
- Shopify/ui-extensions#4489


IHu04:
Replaced by https://github.com/Shopify/ui-extensions/pull/4686 with identical changes consolidated into a single commit.

PR #4661, opened 2026-09-01
[3/6] Add POS List web component types

ajanth-u:
### Background

The 2026-10 POS UI Extensions surface needs a virtualized list component for structured, incrementally loaded data. The public shape is under TAG review in [shop/world#2037349](https://meteorite.shopify.io/repos/shop/world/pulls/2037349) (migrated from the archived Shopify/ui-api-design#1576).

### Solution

Add the `s-pos-list` web component contract to the 2026-10 release branch:

- structured rows with labels, up to three subtitles, badges, images, trailing labels, chevrons, and toggle state
- per-row `onClick` callbacks that receive a plain `POSListRowClickEvent` (`{type: 'click'}`) — rows are data rather than elements, so the callback crosses the sandbox boundary as an RPC callable and can't receive a DOM `Event`; row context is held by each callback's closure
- a disabled toggle switch also blocks its row's activation, and `auto` image display is decided per row, both matching the legacy `List`
- `imageDisplayStrategy`, `loadingMore`, and `loadmore` support
- a `header` slot for composable content above the rows; no component-owned title
- availability on supported POS action extension targets
- default and incremental-loading examples
- a minor changeset

The web-component names intentionally replace legacy React-oriented names: `data` becomes `rows`, `isLoadingMore` becomes `loadingMore`, and `onEndReached` becomes `loadmore` / `onLoadMore`.

### Alternatives considered

1. Restore the legacy `List` contract unchanged — rejected because React-specific names and `listHeaderComponent` do not fit the web-component surface.
2. Add a `title` property — rejected because headings compose outside the list or through the `header` slot.
3. Add a component-level click event carrying row data — rejected because row activation is modeled by per-row `onClick` callbacks, with row context already available through the closure.

### Verification

- `yarn build`
- `yarn type-check`
- `yarn test` (no related Jest suites)
- Prettier on all changed files
- Focused ESLint: `StandardComponents.ts` passes; generated declarations and examples are intentionally ignored by this repository's ESLint configuration
- Full `yarn lint` reaches only pre-existing generated `checkout/preact.js` and `customer-account/preact.js` errors after `yarn build`

### Dependencies

This draft is blocked on approval of [shop/world#2037349](https://meteorite.shopify.io/repos/shop/world/pulls/2037349). End-to-end availability also requires the POS host implementation linked below.

After this PR merges, regenerate the 2026-10 POS reference data in World before the Shopify Dev documentation PR can merge.

### Related PRs

[...]

PR #4690, opened 2026-09-12
[5/7] Add s-pos-list web component types and the posListTemplate compiler

ajanth-u:
## What

Adds `<s-pos-list>` for POS API version 2026-10 and the `posListTemplate` tagged template that compiles its row markup. This is the **templated** candidate for GSD #51755; the legacy-parity alternative is #4661 (unmodified, held as contingency). Exactly one of the two should merge.

Companion PRs: World #2061117 (ui-api-design contract), #2061112 → #2061124 → #2061132 (pos-mobile host stack), Shopify/pos-e2e-ui-extension (sandbox screen, follows), World shopify-dev docs (follows).

## Design

`rows` is plain data (`{id, templateFor, ...fields}`) and `itemTemplates` is a compiled AST produced once in the extension bundle by `posListTemplate`:

```jsx
const itemTemplates = posListTemplate`
  <s-pos-list-item templateId="section-header" type="text">
    <s-heading>{{title}}</s-heading>
  </s-pos-list-item>
  <s-pos-list-item templateId="product">
    <s-stack direction="inline">
      <s-image src="{{image.src}}" alt="{{name}}" />
      <s-text>{{name}}</s-text>
      {{#if onSale}}<s-badge tone="success">Sale</s-badge>{{/if}}
      <s-switch bind:checked="active" />
    </s-stack>
  </s-pos-list-item>
`;
<s-pos-list rows={rows} itemTemplates={itemTemplates} onRowClick={(e) => e.detail.item.id} />
```

POS hydrates one template copy per FlashList-mounted row, so no extension code runs while scrolling and no functions cross the bridge (no Shopify/extensibility change needed).

Output is stamped with this package's API version (`version: '2026-10'`); POS renders templates only for an extension declaring the same API version and reports a mismatch by name (`itemTemplates were compiled for API version X but this extension runs API version Y`).

Grammar: `{{path}}` in text; `{{path}}` inside an attribute (always a string); `bind:prop="path"` (the only typed binding); `{{#if path}}…{{/if}}` (no nesting, no else). Compilation and validation happen when the helper executes, not during the extension build. Define templates at module scope to compile once per module execution. Rejected by the helper with `posListTemplate:`-prefixed errors: `on*` attributes (including `bind:onClick`), `${}` interpolation of any kind, `#each`/other expressions, nested/unbalanced `#if` (including nesting through component children), `#if` inside an attribute, text directly under `<s-pos-list-item>`, non-item roots, nested `s-pos-list-item`/`s-pos-list`, missing/duplicate `templateId`, unknown `type`, any other root attribute.

## Changes

- `src/surfaces/point-of-sale/pos-list-template.ts` — the compiler (`htm`-based; `htm` becomes a direct dependency).
- `src/surfaces/point-of-sale.ts` — exports `posListTemplate` plus `POSListTemplateTag`, `POSListTemplates`, `POSListItemTemplate`, `POSListTemplateNode`, `POSListTemplateProp`, `POSListTemplateSegment`, `POSListRow`, `POSListRowClickEvent`.
- `components/POSList.d.ts` (generated from pos-mobile `meta.ts` via `pnpm build-extension-components:publish 2026-10`, with the generator fixes from World #2061112) and the POSList blocks merged into `components.d.ts`. `POSListRowClickEvent` carries `{item, index}` in `detail`, the only event field POS's RemoteDOM bridge forwards to an extension; the ui-api-design contract (World #2061117) declares `item`/`index` on the event, and POS exposes them once the bridge forwards custom event properties. `POSListTemplates.version` is the compiling API version string (`` `${number}-${number}` | 'unstable' ``). Only POSList additions are included; the pre-existing drift between the generator output and this repo's `components.d.ts`/`components-shared.d.ts` is left alone.
- `components/targets/StandardComponents.ts` — `| 'POSList'`.
- `components/POSList/examples/{default,incremental-loading}.jsx`.
- `docs/.../2026-10/{generated_docs_data_v2.json,targets.json}` regenerated against both release baseline `2f21aa4b` and this branch; only POSList types, their dependencies, component unions, and target mappings are retained. Unrelated release-branch documentation drift is excluded.
- `yarn.lock` adds only `htm@^3.1.1`; unrelated publishing-tool lockfile changes are excluded.
- `.changeset/pos-list-template-component.md` (minor).

## Verification

- `yarn build`, `yarn type-check` green; `build/esm/surfaces/point-of-sale.mjs` exports `posListTemplate`.
- `loom test packages/ui-extensions/src/surfaces/point-of-sale/tests/pos-list-template` — 19 tests: exact AST output for text placeholders, `{{#if}}` blocks, `bind:`, attribute segments, valueless attributes, `type="text"`; and every rejection above with its literal message.
- eslint clean on the changed source files; generated JSON formatting passes.
- Clean dependency install with `yarn install --frozen-lockfile --ignore-scripts --non-interactive` passes without changing the narrowed lockfile.
- Built public-package smoke check passes for versioned, serializable templates, typed bindings, conditionals, and interpolation rejection.
- `package:publint` passes with the package-level CJS-type warning; `package:attw` passes with the repository-configured `false-cjs` exclusion.

Co-authored by AI.


tsdevshop (approved):
Approving based on pairing sessions and iterations. Ideally we also get sign-off from someone in the extensions team before this lands.

packages/ui-extensions/src/surfaces/point-of-sale/components/POSList/examples/incremental-loading.jsx
henryStelle: examples are no longer created in this repo and are instead created directly in the shopify-dev zone.

packages/ui-extensions/package.json
```diff
@@ -169,6 +169,7 @@
     }
   },
   "dependencies": {
+    "htm": "^3.1.1",
```
henryStelle: How large is this dependency and is it properly tree-shaked away if the extension doesn't use a virtualized list?

packages/ui-extensions/src/surfaces/point-of-sale/pos-list-template.ts
```diff
@@ -0,0 +1,358 @@
...
+  strings: TemplateStringsArray,
+  ...values: never[]
+) => POSListTemplates;
+
+/**
+ * The API version this package compiles templates for. POS renders templates only for an
+ * extension declaring the same API version, so a mismatch between the bundled package and
+ * the extension's `api_version` is reported instead of rendering unexpectedly.
+ */
+const API_VERSION = '2026-10';
```
henryStelle: This is a versioned package, so we shouldn't need to declare the API version within it.

PR #4666, opened 2026-09-01
Update Admin Section types and add Number and EmptyState types

kyledurand:
## What

Regenerates the Admin `Section` type definitions from `admin-ui-components` 2.21.2, picking up the header work that has landed in Polaris since these types were last generated (1.25.0), and adds the missing `Number` and `EmptyState` definitions from the same generator run.

### `Section`

New API surface for `s-section`:

| Addition | Kind | Notes |
| --- | --- | --- |
| `subheading` | property | Supporting text under the heading. Not a heading: no document-outline entry, doesn't shift child heading levels, not announced as part of the heading. |
| `primaryAction` | slot | Main call-to-action. |
| `secondaryActions` | slot | Alternative/supporting actions. A `ButtonGroup` holding an action plus an icon-only Menu activator renders as one segmented group. |
| `graphic` | slot | Decorative visual before the heading. Single icon-only `Badge` with `size="large"`; not rendered without a `heading`. |
| `accessory` | slot | Contextual info inline beside the heading text. Badge, Icon, Button, Menu, Text, Avatar, Thumbnail. |
| `supplemental` | slot | Status/metadata for the section as a whole, at the inline-end of the header. Badge, Avatar, Text, Icon, Thumbnail. |

```tsx
<Section heading="Payment" subheading="Captured 2 days ago">
  <Badge slot="graphic" icon="credit-card" size="large" />
  <Badge slot="accessory">Paid</Badge>
  <Text slot="supplemental">$248.00</Text>
  <Button slot="primaryAction">Refund</Button>
  Content
</Section>
```

### `Number` and `EmptyState` (net new)

Both are in the Polaris 3P allowlist (`UI_EXTENSIONS_COMPONENTS` in `admin-ui-components/lib/uiExtensionsBundlePlugin.js`) but had never been generated onto any definition surface, so an extension author writing `s-number` or `s-empty-state` got no types at all.

`s-number` — no slots:

| Property | Type | Notes |
| --- | --- | --- |
| `tone` | `'auto' \| 'neutral' \| 'info' \| 'success' \| 'warning' \| 'caution' \| 'critical'` | Default `'auto'`. |
| `color` | `'base' \| 'subdued'` | Default `'base'`. |
| `fontSize` | `'auto' \| 'small-200' \| 'small-100' \| 'small' \| 'base' \| 'large' \| 'large-100'` | Default `'auto'`; also applies matching line-height and letter-spacing. |
[...]

PR #4692, opened 2026-09-16
Expose EmptyState and Number to standard Admin targets

kyledurand:
### Background

#4666 added the Admin `EmptyState` and `Number` definitions but left both out of `StandardComponents`. This allowlist controls target-specific component typings and the generated target/component support matrix, so the components were not exposed to standard Admin UI extension targets.

### Solution

- Add `EmptyState` and `Number` to `StandardComponents`, preserving the shared allowlist rather than special-casing App Home.
- Regenerate `packages/ui-extensions/docs/surfaces/admin/generated/admin_extensions/2026-10-rc/targets.json`.
- Add a **minor** changeset for `@shopify/ui-extensions`.

`admin.app.home.render` inherits both components through `FormExtensionComponents`. The generated update covers all 50 public targets that inherit `StandardComponents`; the other 31 targets, existing reverse mappings, and all API mappings are unchanged.

### 🎩

Validated with Node `v20.20.0`:

- [x] Run `node packages/ui-extensions/docs/surfaces/admin/build-docs-targets-json.mjs` twice and compare the outputs: byte-for-byte identical.
- [x] Verify generated component lists and reverse mappings against the TypeScript-resolved target unions for all 50 affected public targets, including `admin.app.home.render`.
- [x] Verify built target-specific declarations import both component definitions for all 50 targets.
- [x] `yarn build`
- [x] `yarn type-check`
- [x] `yarn lint`
- [x] `yarn test -- packages/ui-extensions-tester/` — 12 suites, 110 tests passed.
- [x] `git diff --check`

### Follow-up after merge

Refresh the generated reference data in the existing Shopify.dev PR. The hand-authored Empty State and Number pages should not need further content changes.

### Checklist

- [ ] I have :tophat:'d these changes (no manual UI session; validated generated documentation and compiled target typings)


jas7457 (approved):
AI review looks good :thumbsup:

PR #4713, opened 2026-09-18
feat(point-of-sale): add resolution.onSave to validation resolution targets

henryStelle:
## What

Adds `shopify.resolution.onSave` to the POS validation resolution targets (`pos.cart.validations.resolution.render`, `pos.payment.validations.resolution.render`). An extension can register an async handler via `onSave(handler)`; registration returns a promise that resolves to an async cleanup function, and only the most recent handler is kept.

- New `ResolutionSaveHandler` type (`() => void | Promise<void>`) exported from `@shopify/ui-extensions/point-of-sale`, with `@private` JSDoc like the rest of the resolution surface.
- New `onSave(handler): Promise<() => Promise<void>>` member on `ResolutionApiContent<TEventData>`, also `@private`.
- `ui-extensions-tester` factory mocks for both resolution targets now include the asynchronous `onSave` and cleanup function, matching the updated type.

## Why

POS awaits the registered handler when staff advance the resolution flow (Save, Next, or Done), before it re-runs the intercepted validation. This lets extensions commit pending work — cart updates, storage writes — without shipping their own save button; they `await` those calls inside the handler. A rejected promise does not stop the flow: POS still re-runs validation and the interceptor remains the authority on the outcome. The API is `@private`, like the rest of the validation resolution surface.

## Test plan

- [x] `yarn type-check` (loom) — passes
- [x] `yarn lint` (loom: ESLint + Prettier) — passes
- [x] `yarn test pos-resolution-targets` — 8/8 pass, including two tests that assert `onSave` returns a promise resolving to an async cleanup function
- [x] `yarn test pos` — all POS suites pass (35 tests)
- [x] POS tester example `npm run typecheck && npm test` — 6/6 pass
- [x] POS tester example `shopify app build` — passes
- [x] `yarn build` — passes

Resolves: https://github.com/shop/issues-retail/issues/35219


aaronschubert0 (approved):

vctrchu (approved):

packages/ui-extensions/src/surfaces/point-of-sale/api/resolution-api/resolution-api.ts
```diff
@@ -21,4 +29,19 @@ export interface ResolutionApiContent<TEventData> {
...
+   * cart updates or storage writes, and `await` those calls inside the handler.
+   * A rejected promise does not stop the flow; POS still re-runs validation and
+   * the interceptor result decides what happens next.
+   *
+   * Only the most recent handler is kept. Returns a function that unregisters
+   * the handler.
+   *
+   * @private
+   */
+  onSave(handler: ResolutionSaveHandler): () => void;
```
vctrchu: onSave(
  handler: ResolutionSaveHandler,
): Promise<() => Promise<void>>;

henryStelle: Updated the contract and tester mocks to `onSave(handler): Promise<() => void>` in 89307125e.

vctrchu: any reason why we aren't making the clean up a promise as well?

henryStelle: oversight

henryStelle: Corrected the final contract to `onSave(handler): Promise<() => Promise<void>>` in 16fce9272. The registration and returned cleanup function are both async.

